PR #16, opened 2026-06-18
Add roadmap: one-parameter semigroups, completely monotone functions, and BCR Bochner

mrdouglasny:
A roadmap for the **dynamical / representation-theoretic functional-analysis layer** that Mathlib lacks: strongly continuous (C₀) semigroups, generators and resolvents, the **Hille–Yosida** generation theorem; the theory of **completely monotone** functions and **Bernstein's** representation; **continuous positive-definite functions** and **Bochner's** theorem; and the **BCR (Berg–Christensen–Ressel)** representation on involutive semigroups.

**Revised per @kim-em's review.** Rebased onto current `main` (merge conflict resolved). The roadmap now has a **library-building focus** rather than racing to named theorems — each part lists the *API to develop*, then the representation theorem as a milestone, then acceptance examples:

- **General C₀ semigroups first, contraction as a subclass**; Hille–Yosida bounds stated at general `(M, ω)`, contraction as the corollary.
- A full **resolvent API** tied to Mathlib's `resolvent`/`spectrum`: resolvent identity, analyticity, power/derivative formulas, iterated bounds; generator-uniqueness.
- `IsCompletelyMonotone` **structural theory**: closure under sums/products/limits/composition with Bernstein functions, `e^{−tx}` extreme rays, Stieltjes/Bernstein relations.
- **Positive-definite function API** (closure under sums/products/Schur products/limits; continuity-at-0 ⇒ uniform continuity), and **Bochner stated on a general finite-dimensional real inner-product space** (not `Fin d`), with LCA-via-Pontryagin as a labeled stretch goal.
- **Dropped all references to external/our own material** so AI reviewers judge the statements intrinsically and adversarially.

Adds `TauCetiRoadmap/OneParameterSemigroups/README.md` and links it from the top-level README. Markdown only — no effect on the build.

---

*AI attribution:* drafted by **Claude (Opus 4.8)**, with advice from **Gemini** and **Codex**; human-directed and reviewed by M. Douglas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ industry groups.
 7. [Multiquadratic fields and genus theory](TauCetiRoadmap/Multiquadratic/README.md)
 8. [Effective arithmetic bounds and geometry of numbers](TauCetiRoadmap/EffectiveBounds/README.md)
 9. [Geometric topology and the Kirby-list problems](TauCetiRoadmap/GeometricTopology/README.md)
+10. [One-parameter semigroups, completely monotone functions, and BCR Bochner](TauCetiRoadmap/OneParameterSemigroups/README.md)
 
 ## How changes are made
 
--- a/TauCetiRoadmap/OneParameterSemigroups/README.md
+++ b/TauCetiRoadmap/OneParameterSemigroups/README.md
@@ -1,106 +1,170 @@
 # Roadmap: one-parameter semigroups, completely monotone functions, and BCR Bochner
 
 Operator semigroups are the analytic backbone of evolution equations (heat, Fokker–Planck,
-Schrödinger) and of Markov-process theory. Mathlib has the static functional-analysis stack
-(Banach/Hilbert spaces, bounded operators, spectrum, the Bochner integral, Fourier theory)
-but **not** the dynamical layer: strongly continuous (C₀) semigroups, their generators and
-resolvents, the Hille–Yosida generation theorem, Bernstein's theorem on completely monotone
-functions, or the Berg–Christensen–Ressel (BCR) Bochner theorem for positive-definite
-functions on involutive semigroups.
+Schrödinger) and of Markov-process theory. Mathlib has the *static* functional-analysis
+stack — Banach/Hilbert spaces, bounded operators, spectrum, the holomorphic functional
+calculus, the Bochner integral, Fourier theory, unbounded operators via `LinearPMap` — but
+**not the dynamical layer**: strongly continuous (C₀) semigroups, their generators and
+resolvents, the **Hille–Yosida** generation theorem, **Bernstein's theorem** on completely
+monotone functions, or the **Berg–Christensen–Ressel (BCR)** Bochner theorem for
+positive-definite functions on involutive semigroups.
 
 The bar for "done": a researcher in evolution equations or Markov semigroups looks at this
 material and says *"the C₀-semigroup / generator / resolvent API is here, in reusable form,
 and the Hille–Yosida and Bochner-type representation theorems are available."*
 
-> **Provenance note.** A substantial AI-authored development of this material already exists
-> (`mrdouglasny/hille-yosida`, sorry-free, 0 project axioms), covering the Hille–Yosida
-> resolvent identities, Bernstein's theorem, and the BCR semigroup-Bochner theorem. It is a
-> natural source for "make existing material ready to Tau Ceti" PRs. Two adaptation tasks:
-> it targets Lean v4.29.0 (Tau Ceti is on v4.31.0 — a Mathlib bump is needed), and its BCR
-> *existence* half depends on a separate Bochner–Minlos package, so the **resolvent and
-> Bernstein items below are the cleanly Mathlib-only first targets**; the BCR existence item
-> needs a Mathlib-only route to the spatial Bochner theorem first.
+Suggested home: `TauCeti/Analysis/Semigroups/` (the C₀ API, resolvent, generation),
+`TauCeti/Analysis/CompletelyMonotone/` (Bernstein), `TauCeti/Analysis/BCR/` (the
+Bochner-type representations).
 
 ## The end goal (v1)
 
-The three pillars, each a standalone, reusable theorem about contraction C₀-semigroups on a
-Banach space and the representation of the functions that arise from them.
+The three pillars, each a standalone, reusable theorem (stated in `Targets.lean` with
+`sorry`/`def_wanted` as the supporting types land).
 
 ```lean
--- the shapes we are building toward (state in Targets.lean as the supporting types land):
 variable {X : Type*} [NormedAddCommGroup X] [NormedSpace ℝ X] [CompleteSpace X]
 
--- A strongly continuous semigroup and its (unbounded) generator.
--- structure C0Semigroup (X) : ...                 -- S 0 = id, S (s+t) = S s ∘ S t, strong continuity
--- def C0Semigroup.generator (S : C0Semigroup X) : X →ₗ.[ℝ] X        -- A u = lim (S t u - u)/t
+-- 1. Hille–Yosida resolvent: for a contraction C₀-semigroup, R λ = ∫₀^∞ e^{-λt} S t dt
+--    inverts (λ − A) on the domain, with the contraction-scale bound.
+-- theorem resolvent_right_inverse (S : ContractionC0Semigroup X) {l : ℝ} (hl : 0 < l) :
+--     (l • 1 − S.generator) ∘L S.resolvent l = 1
+-- theorem resolvent_norm_le (S : ContractionC0Semigroup X) {l : ℝ} (hl : 0 < l) :
+--     ‖S.resolvent l‖ ≤ 1 / l
 
--- 1. Hille–Yosida resolvent: for a contraction semigroup, R λ = ∫₀^∞ e^{-λt} S t dt
---    inverts (λ - A) and is a contraction-scale bound.
--- theorem resolvent_right_inverse (S : ContractionC0Semigroup X) {λ : ℝ} (hλ : 0 < λ) :
---     (λ • 1 - S.generator) ∘L S.resolvent λ = 1
--- theorem resolvent_norm_le (S : ContractionC0Semigroup X) {λ : ℝ} (hλ : 0 < λ) :
---     ‖S.resolvent λ‖ ≤ 1 / λ
-
--- 2. Hille–Yosida generation theorem (converse / Lumer–Phillips): a densely-defined
---    dissipative operator whose range condition holds generates a contraction semigroup.
+-- 2. Hille–Yosida generation (converse / Lumer–Phillips): densely-defined dissipative +
+--    a range condition ⇒ generates a contraction C₀-semigroup.
 -- theorem generates_of_dissipative (A : DenseLinearOperator X)
---     (hdiss : A.IsDissipative) (hrange : ∀ λ > 0, Surjective (λ • 1 - A)) :
+--     (hd : A.IsDissipative) (hr : ∀ l > 0, Surjective (l • 1 − A)) :
 --     ∃ S : ContractionC0Semigroup X, S.generator = A
 
--- 3a. Bernstein's theorem: a completely monotone function on [0,∞) is the Laplace transform
---     of a unique finite positive measure.
+-- 3a. Bernstein: a completely monotone function on [0,∞) is the Laplace transform of a
+--     unique finite positive measure.
 -- theorem bernstein (f : ℝ → ℝ) (hcm : IsCompletelyMonotone f) :
 --     ∃! μ : Measure ℝ, IsFiniteMeasure μ ∧ μ.support ⊆ Ici 0 ∧
 --       ∀ t ≥ 0, f t = ∫ x, Real.exp (-t * x) ∂μ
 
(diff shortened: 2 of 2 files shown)


kim-em:
Please rebase to fix the merge conflict.

kim-em:
Could you add an AI attribution in the PR description, which specifies which model you used here?

(It doesn't feel much like Claude or Codex to me, which I'm most familiar with, so I'm just trying to calibrate how much the things I don't like in the text are merely consequences of the model used!)

kim-em:
I would like to see this rewritten to have a more "library building" focus. Rather than racing to particular theorems, we should be asking to fill in the basic theory of the objects introduced.

I *hope* that the review rubrics would fight against PRs that directly followed this roadmap, but I'm not sure.

Examples (thanks :robot: Claude for assisting preparing these)
* we should have separate C_0 semigroups, and *then* the subclass of contraction semigroups (e.g. the Hille–Yosida bound you can state more generally)
* results like "the generator determines the semigroup uniquely" 
* it would be good to tie the resolvents here to the existing Mathlib resolvents
* and some basic theory about resolvents, e.g. the resolvent identity R(λ)−R(μ) = (μ−λ)R(λ)R(μ), analyticity of λ ↦ R(λ,A), the derivative formulas R(λ,A)ⁿ = ∫ tⁿ⁻¹/(n−1)! e^{−λt}T(t), and the iterated-resolvent bounds.
* `IsCompletelyMonotone` is introduced, but no mention of all the structural results about it: closed under sums, products, pointwise limits, composition with Bernstein functions; e^{−tx} as the extreme points; the Stieltjes/Bernstein-function relationships; derivative/integral closure
* as soon as we introduce positive definite functions, we probably should give their basic properties (closure under sums/products/Schur products/pointwise limits, continuity-at-0 ⇒ uniform continuity)
* Bochner's theorem is really a statement about a general finite-dimensional inner-product space (or even a locally compact abelian group, via Pontryagin duality). Fixing Fin d coordinates means a user with ℝ² as ℝ × ℝ (or a torus, or any LCA group) can't apply it.

kim-em:
Could try to rewrite it so it doesn't rely on the existing external material you have? At many places, I find it hard to guess what it is asking for, but there are references to particular files in your existing material, and I worry this will hinder the AI reviewers, who might be inclined to just take your existing material as "the standard", and then accept things without sufficient adversarial review.

mrdouglasny:
Thanks Kim, this was really helpful. I've pushed a revision (rebased and force-pushed) addressing all four points.

**Rebase.** Rebased onto current `main`; the README conflict is resolved (our entry is now item 10).

**AI attribution.** Drafted by Claude (Opus 4.8), with advice from Gemini and Codex. Lightly human-edited and directed by me. After the rewrite I also had it independently reviewed by Gemini and Codex and folded their fixes in (a few were real: Bernstein needs complete monotonicity on the closed `[0,∞)` for a *finite* measure, the resolvent has to be a pointwise `X`-valued Bochner integral rather than operator-valued, and `IsCompletelyMonotone` must bundle smoothness because `iteratedDeriv` is total).

**Library-building focus.** Rewrote the roadmap around developing the theory of each object, with the named theorems as milestones inside a fuller API rather than the endpoints. Following your examples:
- General C₀ semigroups first, contraction as a subclass; the Hille–Yosida bounds are stated at the general `(M, ω)` level, with contraction as a corollary.
- "The generator determines the semigroup uniquely" is now an explicit API item (with closedness, `S(t)D(A) ⊆ D(A)`, and bounded ↔ uniformly continuous).
- The resolvent gets its own unbounded notion with a bridge lemma to Mathlib's `resolvent`/`spectrum` (which are Banach-algebra-only), plus the resolvent identity, analyticity, the power/derivative formulas, and the iterated bounds.
- `IsCompletelyMonotone` now comes with its structural theory: closure under sums/products/pointwise limits/composition with Bernstein functions, `e^{−tx}` as the extreme rays, and the Stieltjes/Bernstein and Lévy–Khinchine relationships.
- Positive-definite functions are defined generically on a `StarAddMonoid` (so the product space gives the BCR involution `(t,a) ↦ (t,−a)` for free), with their basic API (closure under sums/products/Schur products, continuity at 0 ⇒ uniform continuity, the PD-kernel/GNS correspondence).
- Bochner is now stated on a general finite-dimensional real inner-product space `V` instead of `Fin d` coordinates (so `ℝ²` as `ℝ × ℝ`, a torus chart, etc. all work), with an LCA group via Pontryagin flagged as a stretch goal.

**External material.** I removed every pointer to my existing hille-yosida material and the per-file references. The roadmap now describes what's wanted intrinsically, so reviewers judge the statements on their own terms rather than treating my code as the standard.

Happy to keep iterating if the generality bar isn't where you'd want it, especially on the Bochner generality (finite-dim `V` for v1 vs going straight to LCA).

kim-em:
@mrdouglasny this looks good to me, it's reasonably general looking now!

Claude+Codex had a few more suggestions, but I don't mind how much of this you decide to incorporate. Just ping me when you've had a chance to look at this (compare against your external material for sanity? ask your AIs?), decided what if any changes to make, and I'll hit merge on this PR.

````quote
# Review notes — PR #16 (one-parameter semigroups roadmap, revised)

The revision lands the library-building shift well: the Part A/B/C "API → milestone →
acceptance examples" structure encodes the philosophy at the document level, the contraction
/ resolvent / completely-monotone / PD / Bochner-generality concerns are substantively
addressed, and the three correctness fixes from independent review (closed `[0,∞)` for a
finite Bernstein measure; the resolvent as a *pointwise* `X`-valued Bochner integral;
bundling smoothness into `IsCompletelyMonotone`) are real and well-judged. The remaining
points below are refinements, not blockers.

*(Cross-checked with a second opinion from Codex; where it corrected my first pass I've taken
the corrected version.)*

## Larger points

**1. Own the scalar-field decision — and fix the Schrödinger mismatch.**
The space is hard-coded to a real Banach space (`[NormedSpace ℝ X]`). This is fine *as a
choice* — Hille–Yosida / Lumer–Phillips are valid over real Banach spaces, and a complex
Hilbert space can be treated as a real Banach space for the semigroup action — but the
roadmap should make the choice explicitly rather than leave it implicit, because it already
says it must "use a complexification for the complex resolvent set and the analyticity of
`λ ↦ R(λ,A)`," and analyticity is intrinsically complex-analytic. Decide real-first vs
complex-first vs `[RCLike 𝕜]` and say so in the "Generality bar."

Relatedly, the motivation leads with **Schrödinger**, which is not just a complex semigroup —
it is a strongly continuous *unitary group* `e^{itH}`. So the contraction-semigroup API alone
does not cleanly cover it (it needs complex scalars *and* two-sided groups). Either add C₀
groups as a stretch goal / example, or drop the implication that the current Part A reaches
Schrödinger.

**2. Spell out the exact positive-definite condition.**
The generic `IsPositiveDefinite` on `[StarAddMonoid M]` is a good design, but the precise
matrix condition is left implicit. On an involutive (semi)group it is
`Σ_{i,j} cᵢ c̄ⱼ · F(aᵢ + aⱼ⋆) ≥ 0` — involving the involution `aⱼ⋆`, *not* the group-style
`F(aᵢ − aⱼ)`. Write the index formula out in the definition so the intended notion is
unambiguous (this is what makes the product-`StarAddMonoid` involution `(t,a)⋆ = (t,−a)` do
the right thing for BCR).

**3. Make the measure encoding consistent (`ℝ≥0`).**
The BCR milestone correctly uses `Measure (ℝ≥0 × V)`, so support on the time axis is automatic.
But the Bernstein milestone still uses `Measure ℝ` with a `μ.support ⊆ Set.Ici 0` side
condition — the same idea in a weaker encoding. Use `Measure ℝ≥0` for Bernstein too; it makes
the support automatic and avoids stating theorems where time is formally allowed to be
negative. Keep one convention across Part B and Part C.

## Smaller points

**4. `DenseLinearOperator` vs `LinearPMap`.**
The prose says model the generator as a `LinearPMap`, but the Lean stub names a bespoke
`DenseLinearOperator X`. Either define `DenseLinearOperator` explicitly as a thin wrapper over
`LinearPMap` (plus a dense-domain proof), or use `LinearPMap` directly with a density
hypothesis — otherwise the stub quietly invites a parallel unbounded-operator API the prose
just disowned.

**5. Add the abstract Cauchy problem as a milestone.**
Orbit differentiability is in the API, but the user-facing payoff — `u(t) = S(t)x` is *the*
solution of `u' = Au`, `u(0) = x` — is never stated as a milestone. Worth adding, phrased
with care around solution classes: classical solutions need `u(t) ∈ D(A)` with `u' = Au`;
the mild-solution statement holds for general `x`. Not a structural gap (it follows from
"generator determines the semigroup"), but it is the motivating result and should be visible.

**6. Two foundational semigroup lemmas.**
List the integral identity `∫₀ᵗ S(s)x ds ∈ D(A)` with `A ∫₀ᵗ S(s)x ds = S(t)x − x`, and a
*named* density-of-`D(A)` theorem for generated semigroups (currently only implicit in "dense
domain"). These are the workhorses behind the generation theorem's density sub-lemma. Density
of `D(Aⁿ)` / smooth vectors is genuinely useful but more advanced — list it as secondary, not
as equally foundational.

**7. Dissipativity: add the converse, and quarantine the Hilbert form.**
Lumer–Phillips currently appears only in the "dissipative + range condition ⇒ generates"
direction. Add the converse ("generator of a contraction semigroup is dissipative") and
maximal dissipativity. Note that the Hilbert-space characterization `Re⟪Ax, x⟫ ≤ 0` is *not*
the general Banach-space notion (which goes through duality maps / resolvent-range
inequalities), so present it explicitly as a Hilbert-only specialization rather than as the
definition.

**8. Keep Hille–Yosida and Lumer–Phillips distinct.**
The revision merges them under a single "Hille–Yosida generation theorem" milestone with
Lumer–Phillips as the contraction case. They are different theorems with different hypotheses
(a resolvent-family bound vs dissipativity + a range/maximality condition); keep their names
and statements separate.

**9. Fourier convention (minor).**
The Bochner / BCR stubs are now internally consistent on Mathlib's `2π` convention
(`e^{−2πi⟨·,·⟩}`), so this is the weakest point — but a stated conversion lemma to the
`e^{i⟨·,·⟩}` form would beat a `⚠` warning, since downstream users will mix the two.

````

mrdouglasny:
Thanks Kim — I've incorporated **all nine** of the round-2 points (and cross-checked against my external material + the AIs):

1. **Scalar field** stated explicitly in the Generality bar (real-Banach-first; complexification for the complex resolvent set / analyticity; `[RCLike 𝕜]` flagged as a later option). And **C₀ groups** (Stone, `e^{itH}`) added as a stretch goal, with the intro no longer implying the contraction API reaches Schrödinger.
2. **Exact PD condition** written out: `Σ_{i,j} cᵢ · conj(cⱼ) · F(aᵢ + aⱼ⋆) ≥ 0`, with the involution called out vs the group form.
3. **Bernstein measure on `ℝ≥0`** now, matching BCR — dropped the `support ⊆ Ici 0` side-condition.
4. **`DenseLinearOperator`** clarified as a thin `LinearPMap` wrapper, not a parallel API.
5. **Abstract Cauchy problem** added as a milestone (classical vs mild solutions).
6. **Foundational lemmas** added: `∫₀ᵗ S(s)x ds ∈ D(A)` with `A∫ = S(t)x − x`, and a named `D(A)`-density theorem.
7. **Dissipativity**: converse + maximal dissipativity added; `Re⟪Ax,x⟫ ≤ 0` flagged as Hilbert-only.
8. **Hille–Yosida and Lumer–Phillips** are now separate milestones with their own hypotheses.
9. **Fourier-convention conversion lemma** added (2π ↔ `e^{i⟨·,·⟩}`) instead of just a warning.

Should be ready to merge whenever you are — thanks for the careful read.

kim-em (approved):

mrdouglasny:
@kim-em — I see your recent PRs are building out Parts B/C of this roadmap directly: completely monotone functions (#281, #299), positive-definite functions on an involutive monoid (#276) + PD kernels (#289), and the char-function/easy-Bochner direction (#288). That's exactly the intent — thanks for picking it up.

To avoid us both doing the same milestones, a proposed division of labor: you've got the **object API** for Parts B/C well in hand; we'd like to contribute the **representation theorems** on top of your definitions, drawn from the (sorry-free, axiom-free) AI-authored `mrdouglasny/hille-yosida`:

- **Bernstein** — completely monotone (on closed `[0,∞)`) ⇔ Laplace transform of a unique finite measure — restated against your `TauCeti.IsCompletelyMonotone` (which matches our definition).
- **Bochner on a finite-dim inner-product space** — continuous positive-definite ⇔ Fourier transform of a finite measure — the build-here keystone, consuming your `charFun_posSemidef`.
- **BCR semigroup–Bochner** (Berg–Christensen–Ressel 4.1.13) — existence + uniqueness — over `ℝ≥0 × V`, using your `TauCeti.IsPositiveDefinite` instantiated rather than a bespoke PD predicate.

We'll reuse your objects, not re-introduce ours. Are you (or the worker campaign) already planning any of these three, or shall we take them? Flagging now so nobody duplicates Bernstein.

kim-em:
> To avoid us both doing the same milestones

Just noting I'm continuing this discussion at [#Tau Ceti > Contention @ 💬](https://leanprover.zulipchat.com/#narrow/channel/610393-Tau-Ceti/topic/Contention/near/605552355). I want to build a bit more infrastructure, first. (Mostly built successfully on Monday, will test tomorrow!)

PR #21, opened 2026-06-21
doc: add exchangeability and de Finetti roadmap

cameronfreer:
## Summary

Adds a new roadmap for Tau Ceti’s library regarding probabilistic symmetries and conditionally independent representations, with de Finetti’s theorem as the first summit theorem.

[Cameron Freer's](https://github.com/cameronfreer) formalization of [exchangeability and  de Finetti's theorem](https://github.com/cameronfreer/exchangeability) is one pinned migration source and source of API warnings, but not the mathematical specification per se.

The roadmap follows the existing TauCetiRoadmap format:

- narrative `README.md`
- `Targets.lean` with compiled Layer 0 `sorry`-signatures
- root `README.md` index entry
- root `TauCetiRoadmap.lean` import

`Targets.lean` compiles the Layer 0 signatures (the symmetry notions, the directing-measure object `ν : Ω → ProbabilityMeasure α`, finite-marginal uniqueness, and the easy bridges) as `sorry`-signatures against Mathlib. Later layers' signatures for product kernels, tails, reverse martingales, Koopman, and the de Finetti representation land as their types become expressible in `TauCeti/`.

## Roadmap shape

The roadmap organizes the project into nine layers:

1. sequence laws, finite marginals, and symmetry notions
2. product kernels, conditional independence, and mixtures
3. process tails and path-space dynamics
4. L² averaging library and the standard-Borel de Finetti route
5. reverse martingales and conditional-expectation limits
6. Koopman operators and invariant σ-algebras
7. directing measures and de Finetti representation
8. public API and examples
9. generalized exchangeability and representation theorems

It also records the reusable library spine, likely Mathlib extraction candidates, a pinned migration source, worked model examples, and later milestones beyond the v1 sequence theorem.

## Files changed

```text
TauCetiRoadmap/Exchangeability/README.md
TauCetiRoadmap/Exchangeability/Targets.lean
TauCetiRoadmap.lean
README.md
```

## AI attribution

Initial proposal seeded with advice from GPT Pro 5.5. PR contents and description drafted by Claude (Opus 4.8) with review by Codex (GPT 5.5). Moderately human-edited and directed by me.

### DIFF
--- a/TauCetiRoadmap/Exchangeability/README.md
+++ b/TauCetiRoadmap/Exchangeability/README.md
@@ -1,0 +1,598 @@
+# Roadmap: exchangeability and de Finetti
+
+This roadmap adds **de Finetti's theorem about exchangeable sequences** to Tau Ceti, along
+with the prerequisites it rests on: finite-dimensional exchangeability, contractability,
+conditionally i.i.d. sequences, tail σ-algebras for processes, reverse-martingale
+convergence along decreasing filtrations, the Koopman mean-ergodic bridge, and the
+de Finetti–Ryll-Nardzewski equivalence. It builds on Mathlib's substantial probability and
+analysis stack, which provides finite and product measures, kernels, `Measure.bind`,
+conditional expectation, filtrations and martingales, `Lp` spaces, Hilbert-space
+projections, measure-preserving maps, and the mean ergodic theorem, but none of the
+exchangeability theory on top.
+
+The source project
+[`cameronfreer/exchangeability`](https://github.com/cameronfreer/exchangeability)
+formalizes Kallenberg's Theorem 1.1 (in *Probabilistic Symmetries and Invariance
+Principles*) for infinite sequences on standard Borel spaces,
+
+```text
+contractable ↔ exchangeable ↔ conditionally i.i.d.
+```
+
+and carries three complete proof routes: reverse martingales, L² contractability bounds,
+and Koopman / mean ergodic theory. We migrate the mathematics in layers, separating
+general-purpose probability infrastructure from the final exchangeability-specific
+theorem.
+
+Suggested homes:
+
+```text
+TauCeti/Probability/Exchangeability/
+TauCeti/Probability/DeFinetti/
+TauCeti/Probability/Process/
+TauCeti/Probability/Martingale/
+TauCeti/Probability/Ergodic/
+TauCeti/MeasureTheory/Measure/
+```
+
+Migration sources, pinned at
+[`e0532e59ceff23edab44dda9ab0655debbc9cc22`](https://github.com/cameronfreer/exchangeability/tree/e0532e59ceff23edab44dda9ab0655debbc9cc22):
+
+The map below tracks the primary Lean sources. Source README, work-plan, and note files are
+background only.
+
+* Layer 0: `Exchangeability/Core.lean`, `Exchangeability/Contractability.lean`,
+  `Exchangeability/ConditionallyIID.lean`, `Exchangeability/Util/StrictMono.lean`, and
+  `Exchangeability/Util/ProductBounds.lean`.
+* Layer 1: `Exchangeability/Probability/MeasureKernels.lean`,
+  `Exchangeability/Probability/InfiniteProduct.lean`,
+  `Exchangeability/Probability/CondIndep.lean`, the
+  `Exchangeability/Probability/CondIndep/` subtree, and
+  `Exchangeability/DeFinetti/CommonEnding.lean`.
+* Layer 2: `Exchangeability/Tail/*.lean`, `Exchangeability/PathSpace/*.lean`, and
+  `Exchangeability/Probability/SigmaAlgebraHelpers.lean`.
+* Layer 3: `Exchangeability/DeFinetti/ViaL2.lean`,
+  `Exchangeability/DeFinetti/ViaL2/*.lean`, `Exchangeability/DeFinetti/L2Helpers.lean`,
+  `Exchangeability/DeFinetti/TheoremViaL2.lean`,
+  `Exchangeability/Bridge/CesaroToCondExp.lean`,
+  `Exchangeability/Probability/CenteredVariables.lean`,
+  `Exchangeability/Probability/IntegrationHelpers.lean`, and
+  `Exchangeability/Probability/LpNormHelpers.lean`.
+* Layer 4: `Exchangeability/Probability/Martingale.lean`,
+  `Exchangeability/Probability/Martingale/Reverse.lean`,
+  `Exchangeability/Probability/Martingale/Convergence.lean`,
+  `Exchangeability/Probability/Martingale/Crossings/*.lean`, and
+  `Exchangeability/Probability/TimeReversalCrossing.lean`. These source crossing and
+  convergence files are the reverse-direction delta; consume Mathlib for the forward
+  upcrossing and convergence API per the build plan below.
+* Layer 5: `Exchangeability/Ergodic/*.lean` and
+  `Exchangeability/DeFinetti/ViaKoopman.lean`,
+  `Exchangeability/DeFinetti/ViaKoopman/*.lean`, and
+  `Exchangeability/DeFinetti/TheoremViaKoopman.lean`.
+* Layer 6: `Exchangeability/DeFinetti/ViaMartingale.lean`,
+  `Exchangeability/DeFinetti/ViaMartingale/*.lean`,
+  `Exchangeability/DeFinetti/MartingaleHelpers.lean`, and
+  `Exchangeability/DeFinetti/TheoremViaMartingale.lean`.
+* Layer 7: `Exchangeability/DeFinetti/BridgeProperty.lean`,
+  `Exchangeability/DeFinetti/Theorem.lean`, `Exchangeability/DeFinetti.lean`, and top-level
+  `Exchangeability.lean`.
+* Cross-layer helpers: `Exchangeability/Probability/CondExp.lean`,
+  `Exchangeability/Probability/TripleLawDropInfo.lean`, and the
+  `Exchangeability/Probability/TripleLawDropInfo/` subtree. Pull these into whichever Tau
+  Ceti layer first needs the corresponding general-purpose facts.
+
+Credit `cameronfreer/exchangeability` in each ported or adapted file, and record when a
+Tau Ceti file intentionally diverges from this source API.
+
+## The end goal (v1)
+
+For a probability space `(Ω, μ)` and a measurable sequence
+
+```lean
+X : ℕ → Ω → α
+```
+
+with `α` a standard Borel space, prove the de Finetti–Ryll-Nardzewski equivalence:
+
+```lean
+-- the shape we are building toward, once the definitions land in TauCeti:
+-- theorem deFinetti_RyllNardzewski_equivalence
+--     {Ω : Type*} [MeasurableSpace Ω]
+--     {α : Type*} [MeasurableSpace α] [StandardBorelSpace α] [Nonempty α]
+--     {μ : Measure Ω} [IsProbabilityMeasure μ]
+--     (X : ℕ → Ω → α)
+--     (hX_meas : ∀ i, Measurable (X i)) :
+--     Contractable μ X ↔ Exchangeable μ X ∧ ConditionallyIID μ X
+--
+-- theorem deFinetti
+--     {Ω : Type*} [MeasurableSpace Ω]
+--     {α : Type*} [MeasurableSpace α] [StandardBorelSpace α] [Nonempty α]
+--     {μ : Measure Ω} [IsProbabilityMeasure μ]
+--     (X : ℕ → Ω → α)
+--     (hX_meas : ∀ i, Measurable (X i))
+--     (hX_exch : Exchangeable μ X) :
+--     ConditionallyIID μ X
+```
+
+The standard-Borel hypothesis is on the value space `α`, where the directing measure and
+the conditional distributions live; the sample space `Ω` needs only a measurable structure
+here. If the martingale proof routes the directing measure through `condExpKernel` (or
+through `condDistrib` specialized to an identity-valued kernel with value space `Ω`), it
+may temporarily also require `[StandardBorelSpace Ω]`, which is a constraint of that route
+rather than of the statement.
+
+The default public theorem should eventually use the reverse-martingale proof. The L²
+route remains as a lighter real-valued theorem, and the Koopman route remains as the
+ergodic-theory route with reusable operator-theoretic infrastructure.
+
+## Standing hypotheses
+
+Spell hypotheses out; do not bundle them.
+
+The basic definitions should be as hypothesis-light as possible:
+
+* `Exchangeable μ X`: invariance of finite-dimensional laws under permutations of `Fin n`.
+* `FullyExchangeable μ X`: invariance of the path law under all permutations of `ℕ`.
+* `Contractable μ X`: invariance under strictly increasing finite subsequences.
+* `ConditionallyIID μ X`: existence of a measurable random probability measure whose
+  finite product kernels give the finite-dimensional laws. Pin the directing-measure API
+  before stating it: either `ν : Ω → ProbabilityMeasure α` (with Mathlib's
+  `ProbabilityMeasure.pi`, and `ProbabilityMeasure.toMeasure_pi` / `ProbabilityMeasure.pi_pi`
+  for the bridge to `Measure.pi` and rectangle evaluation), or a raw `ν : Ω → Measure α`
+  together with an explicit `∀ ω, IsProbabilityMeasure (ν ω)` hypothesis. A
+  `ConditionallyIIDWith μ X ν` relation plus an existential wrapper keeps the directing
+  measure nameable in proofs.
+
+The standard-Borel hypotheses belong in the directing-measure construction and final
+de Finetti theorem, not in every elementary definition. Similarly, L² assumptions belong
+only in the L² route, and shift-preservation assumptions belong only in the Koopman /
+ergodic-theory lane.
+
+Index the v1 theorem by `ℕ`. General exchangeability over other countable index types,
+exchangeable arrays, Aldous–Hoover, Markov exchangeability, and finite de Finetti bounds
+are long-horizon generalizations, not prerequisites for v1.
+
+## What Mathlib already has (consume)
+
+* **Finite products and path laws:** `Measure.map`, finite product measurable spaces,
+  `Measure.pi`, and product-measure API.
+* **π-system uniqueness:** measure extension / uniqueness from a generating π-system,
+  used to prove that finite-dimensional marginals determine laws on sequence space.
+* **Kernels and bind:** `Kernel`, `Measure.bind`, and the Giry-style measure API needed
+  for mixtures of finite product measures.
+* **Conditional expectation:** `μ[f | m]`, tower properties, `condExpL2`, and
+  conditional-expectation estimates.
+* **Filtrations and martingales:** `Filtration`, martingales, submartingales, and the
+  available upcrossing machinery.
+* **`Lp` and Hilbert-space machinery:** `MemLp`, `Lp`, orthogonal projections, symmetric
+  idempotents, and conditional expectation as an `L²` projection.
+* **Ergodic and operator theory:** measure-preserving maps, composition with a
+  measure-preserving map on `Lp`, and the mean ergodic theorem for suitable continuous
+  linear maps.
+
+Consume these directly rather than re-proving Mathlib's product-measure,
+conditional-expectation, Hilbert-space, or mean-ergodic infrastructure.
+
+## What is missing (build here)
+
+The missing pieces are:
+
+* finite-dimensional exchangeability and full exchangeability for sequence laws;
+* contractability and the proof `Exchangeable → Contractable`;
+* finite-dimensional marginal uniqueness on `ℕ → α`;
+* product-kernel measurability for random finite product measures;
+* the common de Finetti ending turning a directing-measure bridge into `ConditionallyIID`;
+* process-relative tail σ-algebras and their antitone filtration structure;
+* reverse martingale convergence for conditional expectations along decreasing filtrations;
+* Koopman operators and the identification of the mean-ergodic projection with conditional
+  expectation onto the invariant σ-algebra;
+* the L², Koopman, and reverse-martingale proof routes for de Finetti.
+
+State the general infrastructure at full generality: reverse martingales for arbitrary
+decreasing filtrations, and Koopman machinery for arbitrary measure-preserving
+transformations, before specializing to path-space shifts.
+
+---
+
+## The build, in layers
+
+The ordering below is the dependency order. As each layer makes the next layer's *types*
+expressible in `TauCeti/`, state its milestones in `Targets.lean` with `sorry`.
+
+### Layer 0: core exchangeability and finite marginals
+
+Suggested home:
+
+```text
+TauCeti/Probability/Exchangeability/Basic.lean
+TauCeti/Probability/Exchangeability/Contractability.lean
+TauCeti/Probability/Exchangeability/FullyExchangeable.lean
+```
+
+Build:
+
+* `Exchangeable μ X`;
+* `FullyExchangeable μ X`;
+* `ExchangeableAt μ X n`;
+* `Contractable μ X`;
+* `pathLaw μ X`;
+* prefix projections and prefix cylinders;
+* finite-dimensional marginal uniqueness for laws on `ℕ → α`;
+* finite approximation of infinite permutations;
+* extension of strictly monotone finite selections to finite permutations.
+
+Key milestones:
+
+```lean
+measure_eq_of_fin_marginals_eq
+measure_eq_of_fin_marginals_eq_prob
+exchangeable_iff_fullyExchangeable
+exists_perm_extending_strictMono
+contractable_of_exchangeable
+Contractable.map_single
+Contractable.map_pair
+Contractable.comp
+```
+
+⚠ **API warning.** Do not define exchangeability as a property of a measure on path space
+only, or as a property of a process only, without bridges. Both viewpoints are useful:
+the process-level statements are what users want, while the path-law statements make
+π-system and shift arguments cleaner.
+
+### Layer 1: product kernels and the common ending
+
+Suggested home:
+
+```text
+TauCeti/MeasureTheory/Measure/ProductKernel.lean
+TauCeti/Probability/DeFinetti/CommonEnding.lean
+```
+
+Build general product-kernel infrastructure:
+
+* measurable rectangles form a π-system;
+* measurable rectangles generate the finite product σ-algebra;
+* measurability of `ω ↦ Measure.pi fun _ : Fin m => ν ω`;
+* the AE-measurable version needed for `Measure.bind_apply`;
+* rectangle evaluation for finite product measures;
+* equality of finite measures from agreement on rectangles.
+
+Then build the common de Finetti ending:
+
+```lean
+conditional_iid_from_directing_measure
+```
+
+The intended bridge hypothesis is an indicator-product factorization: for every finite
+injective selection `k : Fin m → ℕ` and measurable rectangle `B`, the law of
+`(X (k i))ᵢ` equals the mixture of product measures induced by `ν`.
+
+This layer is shared by the L² and Koopman routes, and also useful for the martingale
+route's final finite-product step.
+
+### Layer 2: tail σ-algebras and path-space shift
+
+Suggested home:
+
+```text
+TauCeti/Probability/Process/Tail.lean
+TauCeti/Probability/PathSpace/Shift.lean
+TauCeti/Probability/Ergodic/ShiftInvariantSigma.lean
+```
+
+Build process-relative tails:
+
+```lean
+tailFamily X n
+tailProcess X
+tailFamily_antitone
+tailProcess_le_tailFamily
+tailProcess_le_ambient
+tailProcess_eq_iInf_revFiltration
+```
+
+Build path-space shift:
+
+```lean
+shift : (ℕ → α) → (ℕ → α)
+shift_measurable
+shift_iterate_measurable
+```
+
+Build shift-invariant σ-algebras:
+
+```lean
+isShiftInvariant
+shiftInvariantSigma
+shiftInvariantSigma_le
+mem_shiftInvariantSigma_iff
+shiftInvariantSigma_measurable_shift_eq
+shiftInvariant_implies_shiftInvariantMeasurable
+```
+
+⚠ **Tail vs invariant σ-algebra.** Do not silently identify the tail σ-algebra with the
+shift-invariant σ-algebra. For one-sided sequences, the relationship runs through
+invariance, almost invariance, and completions. State exactly the theorem each proof
+route needs.
+
+### Layer 3: the L² contractability proof
+
+Suggested home:
+
+```text
+TauCeti/Probability/Exchangeability/L2/Covariance.lean
+TauCeti/Probability/Exchangeability/L2/BlockAverages.lean
+TauCeti/Probability/DeFinetti/ViaL2.lean
+```
+
+This is the first proof route to port after the shared layers. It is real-valued and
+requires L² assumptions.
+
+Build:
+
+* equality of means and integrals from equal one-dimensional laws;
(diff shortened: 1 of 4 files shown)


TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+```text
+TauCeti/Probability/Exchangeability/
+TauCeti/Probability/DeFinetti/
+TauCeti/Probability/Process/
+TauCeti/Probability/Martingale/
+TauCeti/Probability/Ergodic/
+TauCeti/MeasureTheory/Measure/
+```
+
+Migration sources, pinned at
```
kim-em: The per-layer source map helps whoever implements, but I'd like it out of the spec and into a clearly secondary "provenance" section. Same worry as on the semigroups roadmap: our reviewers will treat your `exchangeability` files as ground truth and wave statements through, doubly so when you wrote both. Say what each milestone wants from the mathematics, so the source map is "where to look", not "what's correct".

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+--     (hX_meas : ∀ i, Measurable (X i))
+--     (hX_exch : Exchangeable μ X) :
+--     ConditionallyIID μ X
+```
+
+The standard-Borel hypothesis is on the value space `α`, where the directing measure and
+the conditional distributions live; the sample space `Ω` needs only a measurable structure
+here. If the martingale proof routes the directing measure through `condExpKernel` (or
+through `condDistrib` specialized to an identity-valued kernel with value space `Ω`), it
+may temporarily also require `[StandardBorelSpace Ω]`, which is a constraint of that route
```
kim-em: Let's settle this now rather than defer it. The statement doesn't need `[StandardBorelSpace Ω]`: the directing measure lives on `α`, already standard Borel. Build it as a regular conditional distribution over `α^ℕ` (standard Borel because `α` is) given the tail, via `condDistrib`, rather than disintegrating over `Ω` with `condExpKernel`, and the hypothesis never appears. Let's commit to the `Ω`-free statement as the v1 target and build the proof for it. If there's a real obstruction, name it; otherwise drop "temporarily".

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+l2_bound_two_windows_uniform
+l2_bound_long_vs_tail
+weighted_sums_converge_L1
+directing_measure_satisfies_requirements
+conditionallyIID_of_contractable_viaL2
+deFinetti_viaL2
+deFinetti_RyllNardzewski_equivalence_viaL2
+```
+
+The final route should make clear that it proves a real-valued L² theorem, not the full
```
kim-em: I don't want the L² route to stop at real-valued. The argument is naturally real and second-moment, but the standard-Borel result is reachable, so "a lighter real-valued theorem" is a scoping choice, not a limit of the method. The upgrade isn't "apply it to each `g(Xᵢ)` separately": encode finite tuples of bounded test functions into one bounded real variable, apply the real-valued result to those (controlling the joint finite-dimensional laws), then assemble one consistent measurable directing kernel with a countable-generator argument. That assembly is basically the Layer 1 common ending. I'd rather the L² route, and the Koopman route below (also real-valued in your source), target the same standard-Borel `deFinetti`, with real-valued as an internal lemma. If the assembly is genuinely hard, write down why rather than relabel the conclusion.

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+
+Spell hypotheses out; do not bundle them.
+
+The basic definitions should be as hypothesis-light as possible:
+
+* `Exchangeable μ X`: invariance of finite-dimensional laws under permutations of `Fin n`.
+* `FullyExchangeable μ X`: invariance of the path law under all permutations of `ℕ`.
+* `Contractable μ X`: invariance under strictly increasing finite subsequences.
+* `ConditionallyIID μ X`: existence of a measurable random probability measure whose
+  finite product kernels give the finite-dimensional laws. Pin the directing-measure API
```
kim-em, line 112: A few things on `ConditionallyIID`, since every later layer leans on it.

Pick the representation now rather than leaving it open: `ProbabilityMeasure.pi` gives the product and rectangle lemmas for free, so I'd default to `ν : Ω → ProbabilityMeasure α`, with the raw-measure form as an escape hatch. And write the defining condition's exact finite-index formula into the definition, not just prose ("indicator-product factorization", at Layer 1).

Then give `ν` its API as an object: a.e. uniqueness; measurability w.r.t. the exchangeable / tail-invariant σ-algebra (`ν` can be taken as a version of the regular conditional law of `X₀` given that σ-algebra, with `σ(ν)` contained in it up to completion); and the conditional factorization `E[∏_{j<k} f_j(X_{i_j}) | σ(ν)] = ∏_{j<k} ∫ f_j dν` (finite products, integrability stated). Non-vacuity check: an iid sequence has deterministic `ν`, the common marginal law. (The companion fact, triviality of the symmetric σ-algebra for iid sequences, is Hewitt–Savage, worth stating in Layer 2.)

kim-em (commented):
A couple of places settle for a weaker statement or an extra hypothesis than the result needs: the L² route stops at real-valued, and `Ω` picks up a standard-Borel hypothesis "temporarily". I'd rather get the clean version right the first time. "Relax it later" usually bakes in the wrong proof architecture.

Second, the roadmap tracks the file structure of your `exchangeability` repo closely. I'd like the asks stated intrinsically, so reviewers judge each statement on its own terms instead of treating the existing formalization as the standard.

Third, and most important long-term: the roadmap lists the lemmas on the path to the theorem and leaves several objects as bare definitions plus the one fact the proof needs. I'd like each object developed for its own sake. The directing measure is the clearest case (stated in only one of de Finetti's three standard forms), but the symmetry classes, reverse martingales, and Koopman operator all want their basic API too.

TauCetiRoadmap/Exchangeability/Targets.lean
```diff
@@ -0,0 +1,48 @@
...
+* Layer 7: the public API surface (`Exchangeable`, `FullyExchangeable`, `Contractable`,
+  `ConditionallyIID`, and the route theorems) together with the worked examples.
+
+The first implementation PR should add compiled signatures here only for Layer 0
+milestones whose statements elaborate cleanly against the then-current Tau Ceti API.
+-/
+
+namespace TauCetiRoadmap.Exchangeability
+
+-- (no compiled targets yet; see README.md)
```
kim-em: Can the Layer 0 signatures land here now? `Exchangeable`, `FullyExchangeable`, `Contractable`, and the finite-marginal uniqueness statements need only Mathlib, not any not-yet-chosen Tau Ceti API, so they can be compiled `sorry`-signatures today. That matches the other roadmaps and makes the core objects concrete instead of prose. "Wait for the destination API" only bites for `ConditionallyIID` and later.

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+-- the shape we are building toward, once the definitions land in TauCeti:
+-- theorem deFinetti_RyllNardzewski_equivalence
+--     {Ω : Type*} [MeasurableSpace Ω]
+--     {α : Type*} [MeasurableSpace α] [StandardBorelSpace α] [Nonempty α]
+--     {μ : Measure Ω} [IsProbabilityMeasure μ]
+--     (X : ℕ → Ω → α)
+--     (hX_meas : ∀ i, Measurable (X i)) :
+--     Contractable μ X ↔ Exchangeable μ X ∧ ConditionallyIID μ X
+--
+-- theorem deFinetti
```
kim-em, line 59: de Finetti has three standard statements and the roadmap gives only one. Alongside "a directing measure exists" (`ConditionallyIID`), I'd like both other forms stated as milestones:

- empirical-measure form: `(1/n) Σ_{i<n} δ_{Xᵢ}(ω) ⇒ ν(ω)` a.e., weak convergence in `P(α)` (tested against bounded continuous functions on `α`, not pointwise measure equality). This is the law-of-large-numbers form of de Finetti, a milestone rather than a throwaway lemma.
- mixture form: `pathLaw X = ∫ p^{⊗ℕ} dπ(p)` with `π` a unique probability measure on `P(α)` (the mixing measure, `= law ν`). Uniqueness rests on product measures `p^{⊗ℕ}` separating mixing laws, where standard Borel `α` is the right hypothesis.

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+* extension of strictly monotone finite selections to finite permutations.
+
+Key milestones:
+
+```lean
+measure_eq_of_fin_marginals_eq
+measure_eq_of_fin_marginals_eq_prob
+exchangeable_iff_fullyExchangeable
+exists_perm_extending_strictMono
+contractable_of_exchangeable
```
kim-em, line 273: `contractable_of_exchangeable` and `exchangeable_of_conditionallyIID` appear as proof steps; I'd like the symmetry classes developed as classes. The basic theory: closure under coordinatewise pushforward `X ↦ (f ∘ Xᵢ)`; mixtures of iid are exchangeable; exchangeable implies stationary; and the full implication lattice among `ExchangeableAt n`, `Exchangeable`, `FullyExchangeable`, `Contractable`, `ConditionallyIID`, with every easy arrow named and the hard one (`Contractable → ConditionallyIID`) isolated. (`Exchangeable` is `∀ n, ExchangeableAt n`, not a single-`n` arrow.) Two reframings worth stating: `Contractable` is invariance under the monoid of strictly increasing maps `ℕ → ℕ`, and `Exchangeable` (the finitary one) reduces to adjacent transpositions. (Adjacent transpositions generate the finite permutations, so that's `Exchangeable`, not `FullyExchangeable`.)

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+   ```lean
+   condExp_exists_ae_limit_antitone
+   ae_limit_is_condexp_iInf
+   condExp_tendsto_iInf
+   ```
+
+Target theorem:
+
+```lean
+theorem condExp_tendsto_iInf
```
kim-em, line 499: This is the proof's workhorse and general infrastructure, so I'd state more than the a.e. limit. For `f ∈ L¹`, `μ[f | 𝔽ₙ]` also converges in `L¹` to that same `⨅ₙ 𝔽ₙ` limit (conditional expectations of a fixed `f ∈ L¹` are uniformly integrable), and in `Lᵖ` for `f ∈ Lᵖ`; the L¹ form is what most uses want. A backwards submartingale theorem with this as a corollary, and a named reverse-martingale object, are worth it only if Mathlib's martingale API can't already express antitone filtrations cleanly. Don't add a structure to restate what lemmas can say.

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+TauCeti/Probability/Ergodic/Koopman.lean
+TauCeti/Probability/Ergodic/InvariantSigma.lean
+TauCeti/Probability/DeFinetti/ViaKoopman.lean
+```
+
+Build the generic operator-theoretic lane first:
+
+```lean
+koopman
+koopman_isometry
```
kim-em, line 532: `koopman_isometry` on L² understates the operator. For a measure-preserving map it's an isometry on `Lᵖ` for `1 ≤ p < ∞` (norm-preserving on `L^∞`), and positive, unital, and multiplicative on `L^∞` (a Markov operator, and multiplicative, which the "Markov operator" name hides). It's an isometric embedding, not surjective unless the map is invertible mod null sets, so for the one-sided shift it isn't unitary. Stating it only on L² hides that it's the same operator everywhere, and the path-space arguments lean on the multiplicativity.

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,598 @@
...
+* finite de Finetti bounds;
+* de Finetti for other countable index types;
+* exchangeable arrays and Aldous–Hoover;
+* Markov exchangeability;
+* ergodic decomposition of exchangeable laws;
+* Mathlib extraction of general infrastructure.
+
+These are not prerequisites for the v1 theorem.
+
+## Worked examples
```
kim-em, line 684: These are nearly all easy-direction checks. The two examples that delineate the theory are missing. One: a canonical exchangeable-not-iid sequence with an explicit directing measure, stated as the mixture construction (`P ~ Beta(a,b)`, then iid `Bernoulli(P)`), which gives an explicit `ν` and exercises the empirical-measure form; leave the Pólya urn-process equivalence as a later add-on, since the urn dynamics are more work. Two: a stationary-not-exchangeable sequence (two-state stationary Markov chain, `P(0→0) = P(1→1) = q`, `q ≠ 1/2`) to mark the boundary. And once `π`'s uniqueness exists, the extreme-point corollary is cheap: an exchangeable law is extreme iff `π` is a Dirac iff it's iid (no Choquet machinery needed).

cameronfreer:
Thanks @kim-em -- I've tried to revise it into a probabilistic symmetry roadmap rather than a port of my existing repo; please let me know if I should go further in some respects.

- The opening is now about the core math itself (exchangeability, spreadability/contractability, directing measures, tails, reverse martingales as the missing theory), with de Finetti as the first large goal.
- A new "library spine" section names the reusable deliverables rather than emphasizing the final goal.
- The [cameronfreer/exchangeability](https://github.com/cameronfreer/exchangeability) source map is now in a "Migration source" section after "What is missing", framed as where to look rather than as the specification.
- The layer headings are now renamed to library-oriented names.
- The worked examples are now model examples rather than checks on theorem shape.
- Public statements no longer include `[StandardBorelSpace Ω]` (and the "temporarily require `[StandardBorelSpace Ω]`" language is gone); they still quantify over `Ω`.
- Targets.lean now compiles the Layer 0 signatures (the symmetry notions, path-space data, the directing-measure object ν : Ω → ProbabilityMeasure α, finite-marginal uniqueness, and the easy bridges) as sorry-signatures.
- The long-horizon/optional list holds the deeper API items, and I'm deferring the larger results from later layers for now.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Good idea. I think they should both be included eventually. For now I put the empirical-measure and mixture forms in the optional-extensions list rather than core v1, and was thinking they could turn into to milestones after the representation work starts. Is that appropriate, or is it better for a Roadmap to just include them as milestones now?

(I think Mathlib already has the relevant `ProbabilityMeasure` weak-topology infrastructure; the empirical measures themselves still need to be defined.)

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: OK, thanks. I've kept the `ν : Ω → ProbabilityMeasure α` pin, which gets `ProbabilityMeasure.pi`/`toMeasure_pi`/`pi_pi` for free.

re the fuller `ν` API you describe (a.e. uniqueness, measurability against the tail/exchangeable σ-algebra, and the factorization `E[∏_j f_j(X_{i_j}) | σ(ν)] = ∏_j ∫ f_j dν`), I've left this for the implementation, noted as an extension rather than dropped -- is that appropriate for a Roadmap?

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: This is partly done now: the spine treats the symmetry notions and the process/path-law bridges as first-class, and `Targets.lean` now states the core Layer 0 objects (`ExchangeableAt`, `Exchangeable`, `FullyExchangeable`, `Contractable`) plus the easy bridges `Exchangeable ↔ FullyExchangeable`, `Exchangeable → Contractable`, `ConditionallyIID → Exchangeable`, and full-exchangeability ⇒ shift-preservation.

I’m leaving the full implication lattice and alternate characterizations (adjacent transpositions, the strictly-increasing-map monoid) for a follow-up, where they can be developed as named API.

(I also added a naming note so this doesn't read as two parallel APIs: `Contractable` is the one formal target, `Spreadable` just the standard synonym.)

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: That makes sense.

While for some proofs it's straightforward to start with the more general case, here I think it might work well to encode tuples of bounded test functions into a single real variable, apply the real-valued result, and assemble the directing kernel through the common ending.

Assuming it's appropriate for the way Roadmaps are supposed to work, I think it might be best to do that in the implementation rather than relabel the conclusion now (and if the assembly turns out to be the hard part I'll write down why instead of leaving it implicit).

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Upon checking against Mathlib, it looks like `Filtration` is parameterized by a general `Preorder`, so an antitone filtration is just `Filtration ℕᵒᵈ`, with no new structure needed.

Also `tendsto_eLpNorm_condExp` and `uniformIntegrable_condExp_filtration` are there, but the downward theorem isn't.

I renamed the layer to "reverse martingales and conditional-expectation limits", and the L¹/Lᵖ convergence (not just a.e.) will be stated with the Layer 4 signatures.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Good point, thanks. It looks like Mathlib's `Lp.compMeasurePreserving` already gives the Lᵖ isometry (keyed by `[Fact (1 ≤ p)]`; it's the continuity facts that exclude `∞`); what's missing is the Markov side (positive, unital, multiplicative on `L^∞`), which the path-space arguments actually lean on. I've listed that as an extension to build.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Thanks, I reworked these.

The examples are now actual models rather than checks of theorem shape:
 - the i.i.d. law as `ConditionallyIID`/`Exchangeable`/`Contractable`,
 - a two-point-kernel mixture whose directing measure is `ω ↦ κ(θ ω)`,
 - finite-marginal uniqueness,
 - shift-preservation,
 - antitone tails,
 - the downward-theorem smoke test, and
 - `MemLp 2` for bounded observables.
 
The stationary-not-exchangeable chain (with `0 < q < 1`, `q ≠ ½`) and the extreme-point corollary are in the long-horizon list for later.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Sounds good, thanks. The per-layer source map now lives in its own `Migration source` section, framed as
where to find proof scripts, declarations, API warnings, and attribution rather than as the spec; it now says directly that it isn't the mathematical specification.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Agreed; now "temporarily" is gone.

The plan now builds the tail-conditional path law with `condDistrib` on the path-valued variable `Ω → (ℕ → α)`, which needs `ℕ → α` standard Borel (and that follows from `α` being standard Borel), then reads the directing measure off as the one-coordinate conditional law. The standard-Borel requirement is on the codomain of the conditioned variable, not on the conditioning variable.

I think now `[StandardBorelSpace Ω]` would show up only if you went through `condExpKernel` (or `condDistrib` with an identity-valued Ω-kernel), and that stays inside the proof, never in the public statement.

TauCetiRoadmap/Exchangeability/Targets.lean
cameronfreer: Yes, now this is part of this PR, along with a bit of `ConditionallyIID`.

`Targets.lean` now compiles the Layer 0 signatures:
- the symmetry notions (`ExchangeableAt`/`Exchangeable`/`FullyExchangeable`/`Contractable`),
- the directing-measure object (`ConditionallyIID` with `ν : Ω → ProbabilityMeasure α` and the explicit finite-index factorization),
- path-space data (`pathLaw`/`prefixProj`/`shift`),
- finite-marginal uniqueness, and
- the easy bridges (`Exchangeable ↔ FullyExchangeable`, `Exchangeable → Contractable`, `ConditionallyIID → Exchangeable`, and full-exchangeability ⇒ shift-preservation).

(All elaborate against Mathlib as `sorry`-signatures.)

TauCetiRoadmap/Exchangeability/README.md
kim-em: Great, that's exactly the framing I wanted. Thanks.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Perfect, thanks for sorting out the codomain story.

TauCetiRoadmap/Exchangeability/Targets.lean
kim-em: Lovely, thanks for getting these to compile.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Good, thanks for checking the Mathlib side. `Filtration ℕᵒᵈ` is exactly right.

kim-em:
Thanks Cameron, this is in great shape. I've resolved the threads that are settled.

On the question you raised across most of the rest (whether it's fine to put things in an extensions list rather than v1 milestones): here's how I'd like roadmaps to work. Everything on a roadmap is essential, so let's drop the word "optional" entirely. In my experience it mostly teaches our AI contributors that they're allowed to skip the work. It's good to have milestones along the way and to sequence them, but every item should sit inside *some* milestone.

So rather than the "Optional extensions" list, give each of those items a specific home:

- empirical-measure and mixture forms of de Finetti: Layer 6, alongside the directing-measure theorem (that's where `ν` is built), surfaced in the Layer 7 public API.
- the detailed directing-measure API (a.e. uniqueness, the factorization identity): Layer 6; the tail/exchangeable-σ-algebra measurability of `ν` belongs with the σ-algebras in Layer 2.
- Hewitt–Savage (the symmetric σ-algebra is trivial for i.i.d.): Layer 2; the extreme-point corollary follows once `π`'s uniqueness lands, so Layer 6 (or Layer 7).
- the full implication lattice and the alternate characterizations (adjacent transpositions, the strictly-increasing-map monoid): Layer 0, as named API alongside the symmetry classes.
- the full Koopman Markov API (positive, unital, multiplicative on `L^∞`): Layer 5.
- the stationary-not-exchangeable Markov chain: Layer 7, with the worked examples.

For the L² route, the one I most want fixed: the text still says it "proves a real-valued L² theorem, not the full standard-Borel theorem". Let's not relabel the conclusion. The Layer 3 milestone should reach the standard-Borel result through the common ending, with the real-valued statement as the intermediate step.

And the "Long horizon" items (finite de Finetti bounds, other index types, Aldous–Hoover, Markov exchangeability, ergodic decomposition): give them a named later milestone of their own rather than a list that reads as droppable.

Most of this is placing existing items into milestones plus a little rewording. Ping me when it's in and I'll approve.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Layer 3 no longer relabels the conclusion: the real-valued L² convergence is the intermediate analytic step, and through the common ending and a determining class of bounded measurable real observables on the standard Borel state space the route reaches the standard-Borel de Finetti statement.

The heading and the end-goal paragraph now match.

(I flagged that this is a stronger roadmap target than what the source currently carries.)

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: The detailed ν API, including a.e. uniqueness (equality of probability measures a.e., rather than pointwise) and the factorization identity, is now a Layer 6 milestone.

The measurability of ν w.r.t. the tail/exchangeable σ-algebra is in Layer 2 (ν is built in Layer 6, but its σ-algebra target lives with the σ-algebras).

(The `ν : Ω → ProbabilityMeasure α` pin stays.)

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: The empirical-measure form (weak convergence in `P(α)` against bounded continuous functions) and the mixture form are now Layer 6 milestones, surfaced in the Layer 7 public API.

I noted that the empirical form brings in the weak topology on `ProbabilityMeasure α` and isn't a prerequisite for the base directing-measure theorem.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Layer 0 now states the full implication lattice, with every easy implication named and Contractable → ConditionallyIID isolated.

It also includes the alternate characterizations:
- exchangeability via finite permutations and via adjacent transpositions, stated finite-dimensionally over Fin n;
- contractability via the strictly-increasing-map monoid;
- closure under coordinatewise pushforward;
- mixtures-of-i.i.d.-are-exchangeable; and
- exchangeable ⇒ stationary.
 
Each equivalence carries its hypotheses, including that finite ↔ full needs a probability measure plus finite-marginal uniqueness.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Layer 5 now states the operator as the same operator on `Lᵖ` (`1 ≤ p < ∞`), an isometric embedding (not unitary for the one-sided shift), plus the Markov API: positive, unital, and multiplicative for the deterministic Koopman operator. I phrased multiplicativity as special to composition operators, since a general Markov operator is positive and unital but not multiplicative.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Added a stationary non-reversible finite-state Markov chain (the deterministic 3-cycle with uniform stationary law) as the boundary example, which is shift-invariant but not exchangeable, since the law of `(X₀, X₁)` differs from that of `(X₁, X₀)`.

(The two-state `q ≠ ½` version was a trap: a reversible chain has symmetric two-point laws.)

Also added Hewitt–Savage as a worked example; the extreme-point corollary is now a Layer 6 milestone.

cameronfreer:
Thanks! I added a pair of commits that avoids "optional" items and puts everything under some milestone.

The old extensions list is now a named Layer 8 (generalized exchangeability and representation theorems).

Layer 3 now targets the standard-Borel de Finetti statement through the common ending, with the real-valued L² convergence as the intermediate step.

The σ-algebra/Hewitt–Savage, directing-measure API + empirical/mixture forms, Koopman–Markov API, implication lattice, and a Markov-chain example that is stationary but not exchangeable are now in the layers you suggested.

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,727 @@
...
+
+Build:
+
+* `Exchangeable μ X`;
+* `FullyExchangeable μ X`;
+* `ExchangeableAt μ X n`;
+* `Contractable μ X`;
+* `pathLaw μ X`;
+* prefix projections and prefix cylinders;
+* finite-dimensional marginal uniqueness for laws on `ℕ → α`;
```
RemyDegenne: Mathlib already has the fact that finite distributions determine the full distribution (for a general index, not just Nat). See the file Mathlib.Probability.Process.FiniteDimensionalLaws.

TauCetiRoadmap/Exchangeability/README.md
```diff
@@ -0,0 +1,727 @@
...
+```
+
+Build general product-kernel infrastructure:
+
+* measurable rectangles form a π-system;
+* measurable rectangles generate the finite product σ-algebra;
+* measurability of `ω ↦ Measure.pi fun _ : Fin m => ν ω`;
+* the AE-measurable version needed for `Measure.bind_apply`;
+* rectangle evaluation for finite product measures;
+* equality of finite measures from agreement on rectangles.
```
RemyDegenne: Mathlib has most of those things about pi-systems and rectangles.
See [generateFrom_pi](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/MeasurableSpace/Pi.html#generateFrom_pi) or [MeasureTheory.generateFrom_squareCylinders](https://leanprover-community.github.io/mathlib4_docs/Mathlib/MeasureTheory/Constructions/Cylinders.html#MeasureTheory.generateFrom_squareCylinders) depending on your application, and related lemmas.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Good, the heading and end-goal now match, and thanks for flagging that it's a stronger target than the source carries.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Great, the Layer 6 milestone plus the Layer 2 measurability is exactly what I wanted.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Good, both forms as Layer 6 milestones surfaced in Layer 7 works well.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Lovely, the full lattice and the alternate characterizations are just right, and good call keeping the adjacent-transposition form over `Fin n`.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Good, and thanks for the multiplicativity nuance: special to composition operators, not general Markov.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Nice catch on the reversible two-state trap; the non-reversible 3-cycle is the better example.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Thanks. I reworded Layer 0 and the "what Mathlib has" inventory so the roadmap consumes `Mathlib.Probability.Process.FiniteDimensionalLaws` (`map_eq_iff_forall_finset_map_restrict_eq` and `identDistrib_iff_forall_finset_identDistrib`, general index, finite measures) for finite-dimensional-law uniqueness, and keeps only the Tau Ceti-facing `pathLaw` / `Fin n`-prefix wrappers as new work in the roadmap.

TauCetiRoadmap/Exchangeability/README.md
cameronfreer: Thanks. I changed Layer 1 from building product-kernel infrastructure to consuming Mathlib's product/cylinder API and building only the de Finetti-facing adapters, citing `generateFrom_pi`, `isPiSystem_pi`, `squareCylinders`, `isPiSystem_squareCylinders`, `generateFrom_squareCylinders`, and `Measure.pi_eq` / `Measure.pi_pi`. Only the random-product-kernel measurability and the common ending remain as Tau Ceti work.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Thanks Rémy. Cameron's reworked Layer 0 to consume `FiniteDimensionalLaws`, so resolving.

TauCetiRoadmap/Exchangeability/README.md
kim-em: Thanks Rémy. Layer 1 now consumes the cylinder / π-system API, so resolving.

kim-em (approved):
This is in good shape now. Thanks Cameron for the careful revisions across both rounds, and Rémy for the grounding catches. Approving.

cameronfreer:
Thanks @kim-em and @RemyDegenne for the careful reviews!

@kim-em: I tried to claim some of this via the `Intention` template but don't see `Exchangeability` listed in the pulldown. Should I add this as an option somehow, or is there automation that should have done this?

cameronfreer:
> I tried to claim some of this via the `Intention` template but don't see `Exchangeability` listed in the pulldown. Should I add this as an option somehow, or is there automation that should have done this?

@kim-em not sure if this is the preferred method, but please see #39. Thanks!

PR #25, opened 2026-06-22
Roadmap: Weighted orthogonal L² bases (OrthogonalL2Bases)

mrdouglasny:
A new roadmap area, **`OrthogonalL2Bases`**, for the **L²-Hilbert-basis layer** of orthogonal polynomial systems.

## What it builds
Mathlib has orthogonal-polynomial families as *algebraic* objects (`Polynomial.hermite`, `Chebyshev`, `shiftedLegendre`) and an abstract `HilbertBasis` API, but **nothing bridges them**: no result turns a polynomial orthogonality relation `∫ pₘ pₙ w = cₙδ` into a complete orthonormal **basis** of an L² space, and there is no `L²`-product-basis API. This area builds that layer.

- **Part A — the Hermite basis of `L²(ℝ)`** (v1 anchor): the Hermite functions `ψₙ = cₙ Hₙ(x√2) e^{-x²/2}`, their orthonormality, completeness, and `hermiteHilbertBasis : HilbertBasis ℕ 𝕜 (Lp 𝕜 2 volume)` with Parseval — plus the supporting polynomial/function API (generating function, ladder relations, oscillator eigen-equation).
- **Part B — the family-agnostic spine** (factored out of A's proofs): a moment-determinacy completeness toolkit; the √w-normalization bridge from an orthogonality relation to a `HilbertBasis` (measure- and scalar-generic over `[RCLike 𝕜]`); and the `L²(μ ⊗ ν)` / `L²(Measure.pi)` **product-basis** lemma (the completed-tensor analogue of Mathlib's finite-dim `OrthonormalBasis.tensorProduct`).
- **Part C — further instances** (future): multidimensional Hermite, Laguerre, and Jacobi bases, each consuming its family's orthogonality relation.

## Scope boundary
Per-family **orthogonality relations and polynomial identities are inputs** (consumed from Mathlib's polynomial API, not re-derived); the **Hilbert-space structure** — completeness, `HilbertBasis`, Parseval, product bases — is the contribution. The general orthogonal-polynomials-of-a-measure construction (Gram–Schmidt, Favard) is deliberately out of scope.

## Notes
- **README-only** for now, matching `OneParameterSemigroups`. A formal `Targets.lean` of the milestone signatures is prepared (it elaborates against Mathlib) and will follow as a second PR once checked against this repo's pinned toolchain.
- Dependency ordering is "build A concretely, then factor B out of A's proofs," so the spine is validated by reproducing the anchor rather than built speculatively.

*Authoring:* drafted and cross-checked with Claude (Opus 4.8); independent adversarial review passes by Gemini and Codex (GPT-5.x), whose findings (the completeness decay hypothesis, the √w weight-role, scalar/measure genericity) are folded in.


TauCetiRoadmap/OrthogonalL2Bases/README.md
```diff
@@ -0,0 +1,182 @@
...
+## Part C — Further instances (future; consume B)
+
+*Not v1; recorded so Parts A/B are built generically enough to instantiate.*
+
+- **Multidimensional Hermite basis** of `L²(ℝᵈ)` — `Ψ_α(x) = ∏ᵢ ψ_{αᵢ}(xᵢ)` — as the immediate
+  B3 ∘ A instantiation.
+- **Laguerre basis** of `L²([0,∞), x^α e^{-x})` and **Jacobi basis** of `L²([-1,1], (1-x)^α(1+x)^β)`:
+  each is B2 applied to that family's orthogonality relation plus its completeness (B1, or
+  Weierstrass density on the compact Jacobi interval). **Only the completeness + basis** is in scope
+  here; the Laguerre/Jacobi polynomial identities are inputs.
```
kim-em: What does "are inputs" mean? If they don't exist, then they need to be done, and so need to be part of *some* roadmap, quite possibly this one.

This is the "has to be connected to the ground" requirement for roadmaps. If there are "inputs", then the AIs will behave unpredictably / poorly when deciding what to do with them.

TauCetiRoadmap/OrthogonalL2Bases/README.md
```diff
@@ -0,0 +1,182 @@
...
+Jacobi (`μ = volume.restrict (Icc (-1) 1)`) instantiate on their own supports, not just `volume`.
+The √w rescaling + `mkOfOrthogonalEqBot` plumbing, once, for any family. **Instantiation carries a
+change of variables when the family is defined on a rescaled argument:** the Hermite case uses
+`w = e^{-x²}`, `pₙ = Hₙ(·√2)`, `cₙ = n!√π` (so `pₙ·√w/√cₙ = ψₙ`), which is A1's probabilists'
+relation transported by `u = x√2` — *not* `w = e^{-x²/2}` read off A1 directly.
+
+### B3 — L²-product basis
+A Hilbert basis of `L²(μ)` and one of `L²(ν)` give a Hilbert basis of `L²(μ ⊗ ν)` indexed by the
+product; iterated, a basis of `L²(Measure.pi μ)` indexed by `Π i, κ i`. (Mathlib has only the
+finite-dim algebraic `OrthonormalBasis.tensorProduct`.) Not special-function-specific.
```
kim-em: Here you talk about indexing by products, but above you only allow indexing by `Nat`. This needs to be resolved (presumably by generalizing the rest of Part B away from `Nat`).

TauCetiRoadmap/OrthogonalL2Bases/README.md
mrdouglasny: Good point — fixed in 01e419f. You're right that "inputs" left things ungrounded. The rewrite makes every dependency grounded (in Mathlib, or a target here):

- The Hermite orthogonality relation is **not** in Mathlib (only `Polynomial.hermite` and `integral_gaussian` are), so it's now an explicit **target of this roadmap** — A1, `integral_hermite_mul_hermite_mul_gaussian` — not an input.
- The general bridge **B2** takes the orthogonality relation as a *lemma hypothesis*, so it's grounded by construction — a parametrized theorem, not a dangling dependency.
- **Part C is trimmed to the fully-grounded multidimensional Hermite basis** (`B3 ∘ A`, every ingredient a target above). **Laguerre/Jacobi are moved out** to a dedicated future roadmap: their weighted-orthogonality relations aren't in Mathlib, so they must be grounded there as their own targets rather than dangle as inputs here.

TauCetiRoadmap/OrthogonalL2Bases/README.md
mrdouglasny: Clarified in 01e419f. There's no actual `ℕ`-vs-product conflict — it was unclear prose. B3 (the product-basis lemma) is **index-generic**: it takes Hilbert bases indexed by arbitrary `ι₁, ι₂` and produces an `ι₁ × ι₂` index. The family bridges (B2) are `ℕ`-indexed (degree), and B3 **consumes** them as the `ι = ℕ` case, so the multidimensional Hermite basis comes out indexed by `Π i, ℕ` (multi-indices). So B2 stays `ℕ`-indexed (natural for a polynomial system) and the lift to products lives entirely in B3 — the README now states this explicitly. Happy to generalize B2 to a general index instead if you'd prefer that framing.

TauCetiRoadmap/OrthogonalL2Bases/README.md
kim-em: Hmm, I'd prefer if they were just added here, without this we don't really get to exercise the abstract framework proposed.

mrdouglasny: Done — added a second family, **Chebyshev** (Part C), in `0ea057b`. You're right that Hermite alone doesn't exercise the spine.

I went with Chebyshev rather than Laguerre/Jacobi because **Mathlib has neither of those polynomial families** (only `jacobiSym`/`jacobiTheta`), so grounding them would mean defining the families first — whereas Chebyshev is fully grounded *today*:

- Its **orthogonality relation is already proved in Mathlib** — `integral_eval_T_real_mul_eval_T_real_measureT_of_ne` plus the diagonal lemmas (`…_measureT_zero = π`, `integral_T_real_mul_self_measureT_of_ne_zero = π/2`), w.r.t. the Chebyshev measure `measureT`. So here the orthogonality is a genuine existing input, not a target.
- Completeness uses a **different route** than Hermite — Weierstrass density (`polynomialFunctions_closure_eq_top`) on the compact interval, vs Hermite's moment-determinacy/Fourier route. So the completeness toolkit (B1), B2, and B3 are demonstrated across *both* mechanisms — a stronger justification of the abstraction than a second Gaussian-type family would be. B1 now states both routes.
- The open targets the Chebyshev basis still entails (`IsFiniteMeasure measureT`, the `Ioc`↔compact-subtype bridge, the `C ↪ L²` density push, the Chebyshev-span density, the real→`[RCLike 𝕜]` extension) are **named explicitly** in Part C, so nothing dangles.

Multi-d Hermite moved to Part D; Laguerre/Jacobi noted as a separate future roadmap (they'd start by defining the polynomials). Happy to use Legendre instead, or add a third, if you'd prefer.

TauCetiRoadmap/OrthogonalL2Bases/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+### B2 — Orthogonality relation → `HilbertBasis`
+Given polynomials `p : ℕ → ℝ[X]`, a weight `w` with `0 ≤ w` **and `0 < w` a.e.** (else `g·√w = 0`
+fails to force `g = 0`), the relation `∫ pₘ pₙ w ∂μ = cₙ δ` with `cₙ > 0`, and completeness (B1),
+the √w-normalized functions `x ↦ pₙ(x)·√(w x)/√(cₙ)` are orthonormal in `L²(μ)` and, with
+completeness, form a `HilbertBasis ℕ 𝕜 (Lp 𝕜 2 μ)` (cast via `algebraMap ℝ 𝕜`), with Parseval —
+**for any σ-finite reference measure `μ` on `ℝ`**, so the **Chebyshev** instance (Part C) lands on
+`[-1,1]` — either as `μ = μ_T` (the Chebyshev measure) with `w = 1`, or as `μ = volume.restrict
+[-1,1]` with `w = (√(1-x²))⁻¹` — not just `volume` on `ℝ`. The √w rescaling +
+`mkOfOrthogonalEqBot` plumbing, once, for any family. **Instantiation carries a
```
kim-em, line 205: :robot: Claude suggests here:

Adding a second family *here* would actually exercise the B1/B2 spine, which Hermite alone never does. Legendre is the cheap option: in the current Mathlib toolchain `Polynomial.shiftedLegendre` already exists (with `factorial_mul_shiftedLegendre_eq`, `degree_shiftedLegendre`), whereas Laguerre and Jacobi have no polynomial files at all. So Legendre costs only the orthogonality integral `∫₀¹ lₘ lₙ = cₙδ` (`cₙ > 0`) and one B2 instantiation, with no new polynomial family, and it's qualitatively different from Hermite: compact support, weight ≡ 1, `μ = volume.restrict (Icc 0 1)`, so it tests B2's σ-finite-`μ` genericity instead of re-running the Gaussian case. It needs no new completeness route either: compact support already gives `L¹` with all exponential moments, so B1's entire-Fourier-transform lemma applies unchanged. I'd suggest Legendre as a second worked anchor, keeping Laguerre/Jacobi for later since they'd need their polynomial families built first.

TauCetiRoadmap/OrthogonalL2Bases/README.md
```diff
@@ -0,0 +1,258 @@
...
+### A1 — Hermite polynomial API
+- `derivative_hermite_succ : derivative (hermite (n+1)) = (n+1) • hermite n` (+ `@[simp]`); Mathlib's
+  `degree_hermite` is reused (it makes the polynomials span the moment space in A3).
+- Integrability of a polynomial against the Gaussian weight:
+  `Integrable (fun x => aeval x p * Real.exp (-(x²/2)))`.
+- The one-step weighted-pairing recursion `∫ p·H_{n+1}·w = ∫ p'·Hₙ·w` (one IBP via Rodrigues
+  `(Hₙ·w)' = -(H_{n+1}·w)`); and the **generating function** `∑' n, Hₙ(x)·tⁿ/n! = e^{x·t - t²/2}`.
+- **Milestone** `integral_hermite_mul_hermite_mul_gaussian`:
+  `∫ x, Hₘ(x)·Hₙ(x)·e^{-x²/2} = if m = n then n!·√(2π) else 0`.
+- *Acceptance:* `H₀=1, H₁=X, H₂=X²-1`; `⟨H₀,H₀⟩=⟨H₁,H₁⟩=√(2π)`; `⟨H₀,H₂⟩=0`; gen. fn. at `t=0` is `1`.
```
kim-em, line 119: :robot: Claude suggests here:

Worth adding the Fourier-transform behaviour of the basis alongside the oscillator equation: the Hermite functions diagonalize the Fourier transform, `𝓕 ψₙ = (-i)ⁿ ψₙ`. It's the key structural fact about this basis after orthonormality/completeness, it isn't in Mathlib, and A3 already pulls in `Fourier.fourierIntegral` for completeness so the machinery is on hand.

One convention caveat to state explicitly: `(-i)ⁿ` is the eigenvalue for the *unitary angular-frequency* transform `(2π)^{-1/2} ∫ e^{-ixξ} f`. Under Mathlib's `Fourier.fourierIntegral` (the `e^{-2πixξ}` convention, which A3's completeness proof already uses), `ψₙ` as defined with envelope `e^{-x²/2}` is **not** self-dual (the self-dual Gaussian scale is `e^{-πx²}`), so the clean eigen-relation should be stated for the `2π`-scaled functions, or carry the dilation factor. Pin which scaling the milestone uses.

TauCetiRoadmap/OrthogonalL2Bases/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+**Route 1 — moment determinacy (unbounded support, e.g. Hermite).**
+An `L¹` function `g` *all of whose exponential moments are finite* (`∫ e^{a|x|}·|g| < ∞` for every
+`a ≥ 0` — and a single positive `a`, giving strip-analyticity, already suffices for uniqueness) and
+all of whose polynomial moments vanish is a.e. `0` — because that integrability makes `𝓕 g`
+**entire**, with the moments as its Taylor coefficients at `0`, so `𝓕 g ≡ 0`. **The decay
+hypothesis is this weighted-`L¹` integrability, *not* a pointwise bound `|g| ≤ C·e^{-x²/2}`** — the
+distinction matters because the completeness instance `g = (basis test fn)·√w` with an `L²` factor
+is *not* pointwise-bounded, but *does* have finite exponential moments (Cauchy–Schwarz against the
+Gaussian `√w`). *This is the load-bearing analytic content.*
```
kim-em, line 183: :robot: Claude suggests here:

The milestone here should name the load-bearing part. Orthonormality of the products `ψᵢ(x)φⱼ(y)` is the easy Fubini half; the actual theorem is *completeness*, i.e. density of the algebraic tensor `L²(μ)⊗L²(ν)` in `L²(μ⊗ν)`, which needs `μ, ν` σ-finite. Mathlib's `OrthonormalBasis.tensorProduct` (current toolchain) is the abstract tensor with finite-index hypotheses `[Fintype ι₁] [Fintype ι₂]`, so it doesn't give the concrete `L²(μ.prod ν)` statement at all. State B3 as that density/completeness result with the σ-finiteness hypotheses explicit, so it doesn't read as cheaper than it is.

TauCetiRoadmap/OrthogonalL2Bases/README.md
```diff
@@ -0,0 +1,258 @@
...
+### A2 — The Hermite functions `ψₙ : ℝ → ℝ`
+- `hermiteFunction n x = cₙ · aeval (x·√2) (hermite n) · e^{-x²/2}`, with regularity:
+  `Continuous`, `ContDiff ℝ ⊤`, `MemLp _ 2 volume`, and a `SchwartzMap ℝ ℝ` with this underlying
+  function. Companion API: parity `ψₙ(-x)=(-1)ⁿψₙ(x)`, `ψ₀`/`ψ₁` `@[simp]` forms, `‖ψₙ‖₂=1`.
+- **Ladder relations** (pointwise/Schwartz-level): `x·ψₙ = √((n+1)/2)ψ_{n+1} + √(n/2)ψ_{n-1}` and
+  `ψₙ' = √(n/2)ψ_{n-1} - √((n+1)/2)ψ_{n+1}`, whence `aψₙ=√n ψ_{n-1}`, `a†ψₙ=√(n+1)ψ_{n+1}`.
+  Packaging `a, a†` as operators on `L²` with `[a,a†]=1` is a deliberate downstream target — state
+  these ladder identities at the pointwise/`SchwartzMap` level so they elevate to continuous linear
+  operators on `𝒮(ℝ)` later (the GFF / tempered-distribution use) without re-proving.
+- **Milestones:** pointwise orthonormality `integral_hermiteFunction_mul_hermiteFunction`
```
kim-em, line 130: :robot: Claude suggests here:

Minor: since completeness is proved over `ℝ` but the basis is claimed for every `[RCLike 𝕜]`, worth stating the `ℝ → ℂ` step. Because each `ψₙ` is real, `f ∈ L²(ℝ;ℂ)` orthogonal to all `ψₙ` forces `Re f` and `Im f` each orthogonal to all `ψₙ`, so both vanish by the real completeness. Trivial, but it's the line that makes "`ℝ` and `ℂ` are one theorem" actually true.

TauCetiRoadmap/OrthogonalL2Bases/README.md
mrdouglasny: Adopted the key insight, kept Chebyshev. You're right that compact support needs no new completeness route — a compactly-supported `g` automatically has all exponential moments finite, so B1 (moment determinacy) applies unchanged. I've dropped the Weierstrass route entirely; Chebyshev now reuses B1 and exercises the spine through B2's σ-finite-`μ` genericity (a compact weighted measure, not the Gaussian). I kept **Chebyshev** rather than switching to Legendre because its orthogonality relation is *already proved* in Mathlib (`integral_eval_T_real_mul_eval_T_real_measureT_of_ne` + the diagonal lemmas), whereas Legendre's `∫₀¹ lₘlₙ` would be a new target — so Chebyshev is strictly more grounded for the same demonstration. Happy to switch to Legendre if you'd rather. (a13edc8)

TauCetiRoadmap/OrthogonalL2Bases/README.md
mrdouglasny: Added (a13edc8) as an A3 milestone — `𝓕 ψₙ = (-i)ⁿ ψₙ`, flagged as a target (not in Mathlib). And pinned the convention exactly as you noted: `(-i)ⁿ` is the unitary angular-frequency eigenvalue, but under Mathlib's `Fourier.fourierIntegral` (the `e^{-2πixξ}` convention completeness already uses) `e^{-x²/2}` isn't self-dual (self-dual scale `e^{-πx²}`), so the milestone states it for the `2π`-scaled functions / carries the dilation.

TauCetiRoadmap/OrthogonalL2Bases/README.md
mrdouglasny: Restated (a13edc8). B3's milestone now names the load-bearing part explicitly: the real theorem is **density of the algebraic tensor `L²(μ)⊗L²(ν)` in `L²(μ⊗ν)`** (the products being a *complete* ON system), with `[SigmaFinite μ] [SigmaFinite ν]` stated; orthonormality is flagged as the easy Fubini half. Noted that `OrthonormalBasis.tensorProduct`'s `[Fintype]` hypotheses don't give the concrete `L²(μ.prod ν)` statement, so this density result is the gap — not something cheaper.

TauCetiRoadmap/OrthogonalL2Bases/README.md
mrdouglasny: Added (a13edc8). A3's completeness now states the `ℝ→ℂ` step: each `ψₙ` is real, so `f ∈ L²(ℝ;ℂ)` orthogonal to every `ψₙ` forces `Re f` and `Im f` each orthogonal, hence both vanish by the real completeness — the line that makes "ℝ and ℂ are one theorem" literally true.

mrdouglasny:
### API request from a downstream consumer: export the element-level `coe` lemma as a named deliverable

Context: RandomFields **Topic 2** (Gaussian Hilbert / Wiener chaos) declares an interim contract (`OrthogonalPolynomials/Hermite.lean`) that this roadmap will discharge on merge. While vetting that contract, the faithfulness check flagged that a Hilbert basis given as a *bundled* `HilbertBasis` term, without an element-level characterization, is **near-vacuous** — bare existence of an `ℕ`- (resp. `ι×κ`-) indexed basis is just a separability/cardinality fact and carries none of the Hermite / outer-product content.

The roadmap already commits to the non-vacuous form — A3 builds via `mkOfOrthogonalEqBot` from the explicit family, and B3 is emphatic that "the milestone must read as [tensor-density], not as cheaper." So this is **not a scope change**, just an **API-surface** request:

- **Make the element-level characterization a named public deliverable** in A3, B2, and B3 — e.g. `coe_hermiteHilbertBasis : ⇑(hermiteHilbertBasis 𝕜) = hermiteFunctionLp 𝕜`, the B2 analogue `⇑(basis …) = fun n => normalize (pₙ·√w/√cₙ)`, and `coe_productHilbertBasis : ⇑(productBasis b c) = fun (i,j) => b i ⊗ c j`. `HilbertBasis.coe_mkOfOrthogonalEqBot` makes each provable for free; the ask is only that it be **stated and exported**, since downstream `coe_*` discharge obligations consume it and a bundle-only export is green-but-undischargeable.

Two notes for the Gaussian consumer specifically (no action needed, just confirming the path):
- Topic 2 needs the **`L²(γ)` / Hermite-polynomial** basis (`∫ Hₘ Hₙ dγ = n!·δ`), which routes through **B2 at `μ = stdGaussian`, `w = 1`** — relying on B2's stated σ-finite-`μ` genericity — rather than Part A's `L²(ℝ, volume)` / Hermite-function instance.
- It also consumes **B3** index-generic over `ι × κ` for the multi-index chaos basis `Hₐ(x) = ∏ᵢ Hₐᵢ(xᵢ)` — exactly the `Π i, ℕ` shape B3/Part D already target.

kim-em:
> * **Make the element-level characterization a named public deliverable**

Seems reasonable to me. I'll wait for you to make this change, @mrdouglasny. 

If you can remove the `awaiting-author` label when you're ready to take a look, that would be helpful. (Or if you can't remove it, let me know and I'll sort out permissions!)

TauCetiRoadmap/OrthogonalL2Bases/README.md
kim-em: Chebyshev is a good call, and arguably better than Legendre since its orthogonality relation is already in Mathlib. Two families exercises the spine. Resolving.

TauCetiRoadmap/OrthogonalL2Bases/README.md
kim-em: Good, Chebyshev over Legendre makes sense, and dropping the Weierstrass route is right since compact support reuses B1. Resolving.

TauCetiRoadmap/OrthogonalL2Bases/README.md
kim-em: Great, and thanks for pinning the convention exactly. Resolving.

TauCetiRoadmap/OrthogonalL2Bases/README.md
kim-em: Good, the density statement with `SigmaFinite` names the real content. Resolving.

TauCetiRoadmap/OrthogonalL2Bases/README.md
kim-em: Thanks, that's the line. Resolving.

mrdouglasny:
Done — pushed `db294f6`. The element-level characterization is now a named public deliverable in A3, B2, and B3:

- **A3:** `coe_hermiteHilbertBasis : ⇑(hermiteHilbertBasis 𝕜) = hermiteFunctionLp 𝕜`
- **B2:** `coe_*` for the bridge output (`⇑(ofOrthogonality …) = fun n => pₙ·√w/√cₙ` as `Lp`), so Hermite/Chebyshev inherit by specialization
- **B3:** `coe_productHilbertBasis : ⇑(productHilbertBasis b c) = fun ij => b ij.1 ⊗ c ij.2`

Each is immediate from `HilbertBasis.coe_mkOfOrthogonalEqBot`. I also added a generality-bar bullet stating the rule (export the element-level `coe_*`, not just the bundled term) so it doesn't drift across families. No scope change. Removed `awaiting-author`; back to you.

kim-em (approved):

PR #35, opened 2026-06-24
Add roadmap: Contour Integration & the Hungerbühler–Wasem generalized residue theorem

CBirkbeck:
Adds a new roadmap under `TauCetiRoadmap/ContourIntegration/` (README + `Targets.lean`).

**What it is.** A reusable complex-analysis roadmap (home `TauCeti/Analysis/Contour/`) building to the **Hungerbühler–Wasem generalized residue theorem** (*Non-integer valued winding numbers and a generalized Residue Theorem*, arXiv:1808.00997, Thm 3.3): a residue theorem valid when singularities lie **on** the contour, weighted by a **non-integer winding number**. Mathlib has the Cauchy integral formula and circle integrals but no winding-number or residue theory.

**Layers.** (0) curves/cycles + the generalized winding number → (1) its geometry (HW Def 2.1, Prop 2.2, Prop 2.3; the model sector `α/2π`) → (2) residues + the classical residue theorem → (3) the global (homological) Cauchy theorem, by *Dixon's argument* (named as the proof, not the theorem) → (4) **HW Thm 3.3**.

**Why it lives here.** It is the engine the **valence formula** (the forthcoming Modular Forms roadmap) runs on: the fundamental-domain contour passes through the elliptic points `i`, `ρ`, and their `½`, `⅓` valence weights are exactly HW winding numbers of points on the contour.

**Provenance.** Migrated and cleaned from the AINTLIB `LeanModularForms` project (https://github.com/CBirkbeck/AINTLIB), where HW Thm 3.3 is formalized `sorry`-free.

**Status.** Draft. `Targets.lean` elaborates against the pinned Mathlib (`sorry`-stubbed milestones, allowed in this human-owned roadmap library); the `build` check should pass.


TauCetiRoadmap/ContourIntegration/README.md
```diff
@@ -0,0 +1,212 @@
...
+  (`Complex.circleIntegral_sub_center_inv`, the Cauchy integral formula, the disc/annulus
+  theory).
+- **Circle integrals:** `Mathlib/MeasureTheory/Integral/CircleIntegral.lean` (`circleIntegral`,
+  `circleMap`, `∮ z in C(c, R), f z`).
+- **Meromorphic functions:** `Mathlib/Analysis/Meromorphic/*` (`MeromorphicAt`,
+  `MeromorphicOn`, `MeromorphicAt.order`, principal parts) — the residue is the order-`(−1)`
+  Laurent coefficient against this API.
+- **Interval integrals and the FTC:** `Mathlib/MeasureTheory/Integral/IntervalIntegral.lean`,
+  `…/FundThmCalculus.lean` — the substrate for the arc FTC and the real winding integral.
+- **Homotopy of paths and signed curvature:** `Mathlib/Topology/Homotopy/Path.lean`, and the
```
kim-em: The "consume from Mathlib" list includes "the differential-geometry curvature API" as an input to Prop 2.3, but Mathlib has no signed curvature of plane curves. A grep of the pinned toolchain turns up nothing (one unrelated hit in `Measure/Doubling.lean`). So the `½·k_Λ·|Λ̇|` value at a crossing is resting on something that isn't in Mathlib and isn't a target here.

The value itself is `(ẋÿ − ẏẍ)/(2(ẋ² + ẏ²))`, elementary to write down. But I don't want to just bless an inline ad-hoc definition called "signed curvature": properly, this should be a specialisation of a general theory of local signed curvature, and writing our own one-line version under that name is the kind of private-dialect drift we're trying to avoid. So two acceptable outs, your call:

1. **Admit the shortcut.** Say in the roadmap that Prop 2.3 uses an ad-hoc local expression in place of a general signed-curvature theory, and require implementations to flag that in the module docstring, so the shortcut is visible and not mistaken for the real notion.
2. **Develop it properly.** Roadmap a general local-signed-curvature theory for plane curves, either as a layer here or as a subsidiary roadmap this one cites, and get Prop 2.3's value as its specialisation.

Either is fine; what I'd like to avoid is a quietly-invented `signed curvature` sitting in a consume list as if Mathlib had it.

(Related, same milestone: the curvature limit needs `C²` near the crossing, not just `C^{1,1}`. `C^{1,1}` gives the bounded integrand but not the curvature value, so it's worth pinning which regularity each part of Prop 2.3 uses.)

:robot: prepared with assistance from Claude+Codex

TauCetiRoadmap/ContourIntegration/README.md
```diff
@@ -0,0 +1,212 @@
...
+  ⚠ The complex form needs the PV; the real form does not. Keep both and relate them.
+
+### Layer 2: residues, the Cauchy primitive, and the classical residue theorem
+- **FTC along an arc:** if `F' = f` on a neighbourhood of `image γ`, then
+  `∮_γ f = F(γ b) − F(γ a)`, hence `0` for a closed curve; Cauchy's theorem for a contractible
+  contour as a corollary (reconciled with Mathlib's disc statements).
+- **Residue at an isolated singularity** `Res_{z₀} f` (the order-`(−1)` Laurent coefficient;
+  for a simple pole the limit `lim_{z→z₀}(z − z₀)f(z)`); `ℂ`-linearity; the **simple-pole
+  decomposition** `f = (holomorphic) + Σ_s (Res_s f)/(z − s)`.
+- **The classical residue theorem** `∮_C f = 2πi · Σ_s n_s(C)·Res_s f` for a closed
```
kim-em: Two things on this Layer 2 statement, one a hypothesis gap and one an ordering point.

The hypothesis "a cycle `C` avoiding the finite pole set `S`" isn't sufficient on an arbitrary holomorphy domain `Ω`. You also need `C` null-homologous in `Ω` (`n_w(C) = 0` for every `w ∉ Ω`).

And the ordering: the general arbitrary-cycle version is logically at least as strong as the homological Cauchy theorem, since its `S = ∅` case is exactly `∮_C f = 0` for null-homologous `C` (your Layer 3). So stating it in Layer 2 is a forward dependency on Layer 3. That's fine if you scope Layer 2 to a case provable directly from Mathlib's disc Cauchy theorem (a contour bounding a region with the poles inside) and defer the general null-homologous version to Layer 3 or later; otherwise reorder. Mainly I'd like the intended hypothesis made explicit.

:robot: prepared with assistance from Claude+Codex

mrdouglasny (commented):
**Approve in principle — strong and mathematically correct, but the summit isn't pinned.**

*Verified against the roadmap's pinned Mathlib (`v4.31.0-rc1`, `9caeba10`); math cross-checked with Gemini and Codex second opinions.*

Every signature the roadmap commits to checks out (the model-sector winding values, the argument principle), the "consume from Mathlib" inventory is accurate, and the scoping is sound. The one real gap: the headline **HW Thm 3.3** and its load-bearing **definitions** — generalized winding number, residue, and a Cauchy principal-value contour integral — are prose-only, so the objects the roadmap is *about* aren't machine-checked.

## Suggestions (priority order)

1. **Pin the missing definitions as `sorry`-targets** — the generalized `windingNumber` (HW Def 2.1, for arbitrary points incl. on the cycle), `residue`, and especially a **Cauchy principal-value contour-integral `def`**: Mathlib has *none* of these, and Layer 4 cannot even be *stated* without the PV def.
2. **Pin the summit statement** (HW Thm 3.3, + the classical residue theorem and the homological Cauchy theorem). Verify conditions **(A′)/(B)** and "null-homologous" against arXiv:1808.00997 — note **(B) couples `f` with `C`'s parametrization** (entry/exit tangents at the on-cycle pole), so confirm it's expressible in Lean; forcing the statement is the only check that it is. (This dovetails with the null-homologous hypothesis @kim-em already flagged.)
3. **Name the targets** — replace the anonymous `example`s with named `theorem`/`def`s so milestones are claimable, trackable, and referenceable by the valence formula.
4. **Seed the on-cycle acceptance gates** — `n_c(circle) = 1` and the half-residue `πi·Res_s f` for a simple pole on a smooth arc (the bridge to the valence formula's `i`, `ρ`).
5. **Drop the redundant `hbdry` hypothesis** in the argument-principle targets (derivable from `honly`/`hsupp` + `R > 0`); no `f ≢ 0` hypothesis is needed — it's already implicit (`f ≡ 0` ⇒ order `⊤`, which `honly` + `R > 0` exclude).
6. **README inventory drifts (one-line fixes)** — replace *all* README occurrences of `MeromorphicAt.order` with the free function `meromorphicOrderAt`; `circleMap` now lives in `Analysis/SpecialFunctions/Complex/CircleMap.lean`; `MeromorphicOn` is in `Analysis/Meromorphic/Basic.lean` (with `meromorphicOrderAt` in `Order.lean`); and double-check the exact current name of the Cauchy-kernel circle-integral lemma you cite (`circleIntegral_sub_center_inv`).

**What's already good (keep):** the math is correct, the consumed-Mathlib inventory is accurate, off-contour cases are rightly not seeded, and the "name the theorem / attribute the proof" style (homology Cauchy theorem via *Dixon's argument*) is the right Tau-Ceti convention. No mathematical errors found — items 1–3 are what keep it from being a fully reviewable spec; 4–6 are polish.

:robot: prepared with assistance from Claude+Codex+Gemini

TauCetiRoadmap/ContourIntegration/README.md
CBirkbeck: Addressed in 0c26cad. Signed curvature is removed from the "consume from Mathlib" list — you're right that Mathlib has none. HW Prop 2.3's crossing value is now stated by its explicit closed form `(ẋÿ − ẏẍ)/(2(ẋ² + ẏ²))` (= ½·k·|Λ̇|, the textbook signed curvature written out), exactly as HW and the existing AINTLIB proof do, rather than routed through a general signed-curvature theory; the module docstring is to record that the value is given by this explicit formula since there is no Mathlib signed-curvature API to cite.

I also pinned the regularity per part, as you flagged: the bounded integrand needs only `C^{1,1}`, while the crossing value `½·k·|Λ̇|` needs `C²` near the crossing.

TauCetiRoadmap/ContourIntegration/README.md
CBirkbeck: Addressed in 0c26cad. The null-homologous hypothesis is now explicit, and the classical residue theorem is split into two scopes:

- **Disc/circle case (kept in Layer 2):** poles strictly inside the disc, provable from Mathlib's disc Cauchy theory — seeded as `classicalResidueTheorem_circle` in `Targets.lean`.
- **General arbitrary-cycle case (deferred to Layer 3+):** requires `C` null-homologous in `Ω` (`n_w(C) = 0` for every `w ∉ Ω`); as you note, avoiding `S` alone is not sufficient on an arbitrary `Ω`. It is deferred because its `S = ∅` case is exactly the homology Cauchy theorem (Layer 3), so it is at least as strong.

`IsNullHomologous` is pinned as a `def` in `Targets.lean`, and the homology Cauchy theorem (`homologyCauchyTheorem`) and HW Thm 3.3 (`hungerbuhlerWasem_residueTheorem`) both carry it as a hypothesis.

CBirkbeck:
Thanks @mrdouglasny — all six suggestions are in (commit 0c26cad), build green:

1. **Defs pinned** as `sorry`-targets: `windingNumber` (HW Def 2.1), `residue`, and the Cauchy principal value `HasCauchyPV` (plus `IsNullHomologous`, `ConditionAprime`, `ConditionB`).
2. **Summit pinned**: `classicalResidueTheorem_circle`, `homologyCauchyTheorem`, and `hungerbuhlerWasem_residueTheorem` (HW Thm 3.3), with the null-homologous hypothesis and conditions (A′)/(B) as explicit named hypotheses.
3. **Targets named** — the six anonymous `example`s are now `theorem`/`def`s, so they're claimable and referenceable.
4. **On-cycle gates** seeded: `windingNumber_circle` (`n_c(circle) = 1`) and `hasCauchyPV_half_residue` (the half-residue `πi·Res_s f` on a smooth arc).
5. **Dropped `hbdry`** from both argument-principle targets (derivable from `honly`/`hsupp` + `R > 0`); no `f ≢ 0` hypothesis is needed.
6. **Name drift fixed** in the README: `meromorphicOrderAt`, the `circleMap`/`MeromorphicOn` paths, and `circleIntegral.integral_sub_center_inv`.

Conditions (A′)/(B) are stated as named hypothesis-predicates (`ConditionAprime`/`ConditionB`) so the summit statement elaborates against pinned Mathlib — happy to refine their exact Lean form against arXiv:1808.00997.

TauCetiRoadmap/ContourIntegration/README.md
```diff
@@ -0,0 +1,242 @@
...
+The ultimate target of this roadmap is the **Hungerbühler–Wasem generalized residue
+theorem** (N. Hungerbühler, M. Wasem, *Non-integer valued winding numbers and a generalized
+Residue Theorem*, arXiv:1808.00997, 2018): a residue theorem that remains valid when the
+singularities lie **on** the contour, with the residues weighted by a **non-integer-valued
+winding number** `n_{z₀}(C) = PV (2πi)⁻¹ ∮_C dz/(z − z₀)` that has a genuine geometric
+meaning (an angle, divided by `2π`) for points on the curve. Along the way we build its
+prerequisites — the generalized winding number and its geometry, the classical residue
+theorem, and the global (homological) Cauchy theorem — none of which Mathlib has.
+
+This is ordinary complex analysis and belongs in Mathlib's analysis tree. Its first client is
```
kim-em:
```suggestion
This is ordinary complex analysis and eventually belongs in Mathlib's analysis tree. We'll land a version in Tau Ceti, but if Mathlib later develops this material more fully, then Tau Ceti will refactor as needed and defer to the Mathlib version, as usual. The first client is
```

TauCetiRoadmap/ContourIntegration/README.md
```diff
@@ -0,0 +1,242 @@
...
+`ρ`. Their fractional contributions to the valence formula (`½` at `i`, `⅓` at `ρ`) are
+exactly Hungerbühler–Wasem winding numbers of points on the contour. So the non-integer
+theory is not decoration; it is the engine the modular application runs on.
+
+Suggested home: `TauCeti/Analysis/Contour/`.
+
+## Standing conventions
+
+Pin these once; implementors drift badly otherwise.
+
```
kim-em:
```suggestion
```

TauCetiRoadmap/ContourIntegration/Targets.lean
```diff
@@ -0,0 +1,233 @@
...
+/-! ## The objects (Layers 0, 2, 4): definitions pinned as `sorry`-targets
+
+Mathlib has none of these. The contour integral itself is `∫ t in a..b, deriv γ t • f (γ t)`
+(`intervalIntegral`, agreeing with `circleIntegral` on a circle); these `def`s name the winding
+number, residue, principal value, and the HW hypotheses built on it. -/
+
+/-- **Generalized winding number** (HW Def 2.1): `n_{z₀}(γ) = PV (2πi)⁻¹ ∮_γ dz/(z − z₀)`, for a
+closed piecewise-`C¹` curve `γ : ℝ → ℂ` on `[a, b]` and *any* `z₀ ∈ ℂ` — including `z₀` on the
+curve, where it is in general **non-integer** (the geometric angle `α/2π`). For `z₀ ∉ image γ` it is
+the classical integer index. -/
```
kim-em: :robot: Claude suggests:

The docstring says "for a closed piecewise-`C¹` curve `γ`", which reads as an enforced precondition — but the `def` quite deliberately doesn't require `γ a = γ b`, and shouldn't. The underlying object is just the contour integral `(2πi)⁻¹ ∫ t in a..b, γ'/(γ − z₀)`, which is a well-defined value for *any* curve. Closedness isn't what makes it well-defined; it's the hypothesis under which the value is a genuine *winding* number (off-curve integrality, the `n_c(circle) = 1` gate, the homology/residue theorems — all of which already carry `hclosed : γ a = γ b` explicitly). Keeping it off the `def` is also what lets `windingNumber_modelSector` reuse this on a single arc, which isn't closed at all. Suggest making the docstring match the design:

```suggestion
/-- **Generalized winding number** (HW Def 2.1): `n_{z₀}(γ) = PV (2πi)⁻¹ ∮_γ dz/(z − z₀)`, the
value of the contour integral for *any* piecewise-`C¹` curve `γ : ℝ → ℂ` on `[a, b]` and *any*
`z₀ ∈ ℂ`. Closedness (`γ a = γ b`) is **not** part of the definition; it is the hypothesis (carried
on the theorems below) under which this value is a genuine winding number — the classical integer
index for `z₀ ∉ image γ`, and in general **non-integer** for `z₀` on the curve (the geometric angle
`α/2π`). -/
```

TauCetiRoadmap/ContourIntegration/README.md
CBirkbeck: Applied in bd9cd08 — used your wording.

TauCetiRoadmap/ContourIntegration/README.md
CBirkbeck: Applied in bd9cd08 — dropped the "pin these once; implementors drift badly otherwise" aside and its blank line.

TauCetiRoadmap/ContourIntegration/Targets.lean
CBirkbeck: Applied in bd9cd08 — the docstring now states that closedness (γ a = γ b) is not part of the def, only the hypothesis on the theorems below under which the value is a genuine winding number. Good catch on the windingNumber_modelSector reuse on a single arc.

CBirkbeck:
@kim-em @mrdouglasny — all review feedback is addressed and `build` is green:

- Latest suggestions (the Mathlib-deferral wording, dropping the "pin these once" aside, and the `windingNumber` docstring so closedness is the theorems' hypothesis rather than part of the `def`) → `bd9cd08`.
- Earlier round (the defs and the summit pinned as `sorry`-targets, the anonymous `example`s named, the on-cycle gates seeded, the README name-drift fixed, and the Layer 2 null-homologous / disc-case scoping) → `0c26cad`.

Ready for another look when you have a moment.

kim-em (approved):

PR #41, opened 2026-06-27
doc: add grounded roadmap for infinitary syntax, back-and-forth, and Scott analysis

cameronfreer:
## Summary

Adds `TauCetiRoadmap/InfinitaryLogic/`, a grounded roadmap for the first Tau Ceti infinitary-model-theory spine: (1) the fixed-carrier infinitary syntax `BoundedFormulaInf L ι α n` over `FirstOrder.Language`, with Lω₁ω as the definitional `ι := ℕ` specialization, plus the carrier-coding/transport algebra (`IndexCoding`, coded connectives, `reindex`) and countable-fragment recovery (`indexBound`/`IsCountable`/`ofCountable`); (2) the carrier-generic finitary embedding `toInf` + realization compatibility; (3) back-and-forth / potential isomorphism — an explicit back-and-forth system of Mathlib `FGEquiv`s with its basic API (`potentialIso_of_equiv`, `PotentialIso.symm`, `PotentialIso.trans`) and a one-way bridge from Mathlib's `IsExtensionPair` — and Karp's theorem at a common carrier (`karp_theorem_at`); (4) the countable coded-formula proxy for the Scott-refinement formulas and refinement counting; (5) Scott's isomorphism theorem for countable relational languages (the v1 summit).

This roadmap does not cover admissible sets / Barwise compactness, invariant descriptive set theory (Silver / G₀ / Glimm–Effros, López–Escobar, Morley counting), EM / Erdős–Rado / Morley–Hanf, many-sorted model theory, or Lκλ; those are out of scope here and belong in separate roadmap PRs with their own grounded dependency paths. A "Coordination" section links the in-flight Zulip threads this roadmap touches.

Layer 0 uses the fixed-carrier formulation suggested by Aaron Liu and validated against the main downstream consumers in [infinitary-logic #43](https://github.com/cameronfreer/infinitary-logic/pull/43). Mathlib has not reviewed the proposed API; per repository policy the roadmap builds it here and adapts to Mathlib's names if Mathlib later supplies the corresponding API.

## Contents

- `TauCetiRoadmap/InfinitaryLogic/README.md` (narrowed roadmap), `Suggested.lean` (Layer 0/1 elaborating signatures, opened with the standard "README.md is definitive / not exhaustive" note), root index entry, the `Suggested` import.

## Verification

- Full `lake build` green (8741 jobs); `Suggested.lean` elaborates with only the ten intended `sorry` milestones (`realize_iInfAlong`, `realize_reindex`, `realize_toInf`, `card_le_aleph0_iff_countable_relations`, `karp_theorem_at`, `potentialIso_of_equiv`, `PotentialIso.symm`, `PotentialIso.trans`, `potentialIso_of_isExtensionPair`, `countable_potentialIso_iff_iso`) — all definitions carry real bodies (no `def _ : Prop := sorry`).
- Mathlib paths verified at the pin (mathlib `de5ce8a9a` / Lean v4.34.0-rc1).

## Migration source

- **Production checkpoint**: [infinitary-logic@06bd522](https://github.com/cameronfreer/infinitary-logic/tree/06bd522586f97ccf3503b7f47f250943b33a7f99), used for proof scripts and the declaration map. Its syntax predates the fixed-carrier design, so its Layer 0 types should not be copied.
- **Design evidence**: [infinitary-logic #43](https://github.com/cameronfreer/infinitary-logic/pull/43), used to validate the proposed fixed-carrier Layer 0 API.

Neither source is the specification.

## Attribution

Cameron Freer wrote this with assistance from GPT Pro, Claude Code, Codex, lean4-skills, and lean-lsp-mcp. The fixed-carrier formulation was suggested by Aaron Liu on Zulip.

Thanks to @kim-em and @mrdouglasny for helpful reviews.


kim-em:
I don't think this is quite at our "connected all the way to the ground" bar yet, and there are a couple of gaps I'd like closed before we merge:

- **Layers 8–9 (the descriptive set theory).** At the pinned Mathlib, there's no Borel hierarchy / pointclasses, no Borel codes, no E₀ / hyperfinite, and no Gandy–Harrington. The prose names some of this, but the target list goes from `IsBorelEquivalence` / `BorelReducible` straight to Silver / G₀ / Glimm–Effros without making the intervening infrastructure explicit. (Your migration source files `GandyHarrington` and `SilverBurgess` under `Conditional/`, which makes me wary of treating them as ready targets.) I think it's best if we drop these layers from the roadmap for now, and deal with them in a follow-up.
- **Layer 4 (admissible sets / KP).** You correctly say Mathlib has none and it "must be built here", but the targets jump to `AdmissibleFragment` / `barwise_compactness`. The admissible-set theory itself (KP, Σ-definability, A-finiteness, set-coded syntax) needs to be a target too, by the same rule.

Separately, there's a fair amount of related work already in motion that I think we'll need to coordinate with rather than route around:

- Your own [#mathlib4 > ModelTheory: API for infinitary formulas of L_{∞,ω}](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/ModelTheory.3A.20API.20for.20infinitary.20formulas.20of.20L_.7B.E2.88.9E.2C.CF.89.7D) thread — the single-core-vs-parallel-types question you raised there never got an answer, and this roadmap commits to parallel types. Worth citing it and saying why.
- [#mathlib4 > Cantor-Bendixson analysis](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Cantor-Bendixson.20analysis) (Violeta is about to PR perfect-kernel / the ordinal-stabilization lemma your Scott rank needs) and [#mathlib4 > Naming Suslin Spaces](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Naming.20Suslin.20Spaces) (the Effros theorem + `AnalyticSpace`, [mathlib4#32742](https://github.com/leanprover-community/mathlib4/pull/32742)) overlap Layers 8–9.
- [#mathlib4 > Many-sorted model theory](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Many-sorted.20model.20theory) (Mathias Stout's group, with Aaron Anderson) overlaps Layer 11.

Given how much of this is live, I suspect merging this roadmap may need to wait until Tau Ceti is public — coordinating with these efforts will be much easier once the Zulip channel is public.

This is sufficiently far outside my own expertise that I'm not the right person to give it a final sign-off. We'll need to find another reviewer with real model-theory / DST background (waiting for release should help there too — several of the people in the threads above would be ideal). :-)

mrdouglasny:
# Structural review — TauCetiRoadmap #41

**Roadmap:** *Infinitary logic, Scott analysis, and the descriptive set theory of countable models*
**Author:** cameronfreer · **PR:** [FormalFrontier/TauCetiRoadmap#41](https://github.com/FormalFrontier/TauCetiRoadmap/pull/41) · **Head:** `221cde7`
**Reviewers (purely AI, independent):** Claude Opus 4.8 · Gemini 3.1 Pro · Codex / GPT‑5.4
**Date:** 2026‑06‑27 · **Type:** structural review (not a proof review)

---

## Verdict (unanimous)

**Structurally excellent and house‑pattern‑perfect — but scoped as a multi‑year *program*, not a single roadmap.** Cut v1 to **Layers 0–2 (+ relationalization)** and split the remainder into dependent sibling roadmaps. The roadmap would pass a TauCeti review on *quality*; it should be **restructured for scope and hidden dependencies** before it is handed to the pod, or the agents will hit the unscaffolded admissible‑set and effective‑DST walls mid‑run.

All three reviewers reached this independently — the agreement is itself a confidence signal.

---

## What all three agree on

### Strengths
- **Exemplary house‑pattern + generality discipline** — every required section; unbundled instances (`[L.IsRelational]`, `[Countable M]`, `[Countable (Σ n, L.Relations n)]` carried separately); `ω₁ = Ordinal.omega 1` with no bespoke subtype; intrinsic spec with the existing Lean formalization pinned only as a **secondary** migration source.
- **Correct core architecture** — parallel infinitary inductive (not an extension of finitary `BoundedFormula`); the `FormulaCode` countable proxy that turns "countably many Scott formulas" into a theorem; universe‑`w` Karp.
- **Honest about the source's real gaps** — flags that the migration source's hard cases live in its `Conditional/` directory (Hanf transfer, Gandy–Harrington).
- **Consume/build split is factually correct** — Codex verified against live Mathlib docs that `BoundedFormula`'s `iSup`/`iInf` really are `[Finite]`‑restricted and that `IsRelational` / `Symbols` / `FGEquiv` / `IsExtensionPair` / standard‑Borel / analytic‑sets genuinely exist to consume. No bogus "Mathlib already has it" claims.

### The core finding: scope → split
It bundles **four separate disciplines** (infinitary model theory · admissible‑set theory/KP · partition calculus · descriptive set theory) into one linear roadmap.

- **Consensus minimal v1 = Layers 0–2** (syntax → Karp → Scott), with **relationalization (Layer 11) pulled up** (it depends only on 0–2, and users want Scott for function/algebraic languages early).
- Everything else → dependent sibling roadmaps. Gemini groups into ~3–4 (Core / Partition / DST / *eject* admissible sets); Codex splits finely (A–M). Same shape, different granularity.
- **Spin off Layer 6 (Erdős–Rado / partition calculus)** as standalone, Mathlib‑upstreamable combinatorics.
- **Eject Layer 4 (Barwise / admissible sets)** to a dedicated Kripke–Platek roadmap — the KP / Δ₀‑set‑theory API does not exist in Mathlib and the roadmap allocates **zero** layers to it.

### Under‑scoped layers (difficulty hidden as bullets)
**L4** (admissible sets — no Mathlib KP), **L9** (Silver / G₀ / Glimm–Effros — three deep dichotomies needing effective DST / Gandy–Harrington scaffolded nowhere), **L8**, **L13**. Codex adds **L0** (under‑scoped for its blast radius — binding/substitution/universes/recursion principles every later theorem inherits) and **L3**.

### Critical‑path single points of failure
1. **`refinement_countable` / `BFEquiv_iff_agree_codes`** — the entire Scott summit rides on this one bridge. Vet the `FormulaCode` encoding by hand before any agent starts (avoid dependent‑type‑transport hell).
2. **Silver → Morley counting** — cannot proceed without a real Borel‑equivalence‑relation library that does not yet exist.

---

## Distinct contributions by reviewer

### Gemini 3.1 Pro
- **The "effective‑DST iceberg":** L9/L13 silently need lightface / Gandy–Harrington machinery absent from both Mathlib and Layers 0–8.
- **Universe‑clash warning** for `BoundedFormulaInf` (universe‑indexed) vs model universes — Mathlib's `FirstOrder.Language` already struggles with universe bumping; infinitary conjunctions make it perilous.
- Eject admissible sets to a KP roadmap; move relationalization up.

### Codex / GPT‑5.4
- **"No forward references is mostly aspirational"** — L8–10 silently depend on more DST than listed; L4 on a whole KP development; **L12 (Lκλ) may force a different syntax architecture back onto Layer 0** (a real hidden forward‑dependency the ordering claim misses).
- **Layer 0 under‑scoped** for its blast radius.
- Replace the *"coded proxy **or** direct game‑counting"* ambiguity with **one chosen route**; demote the other to a provenance note.
- Add a **per‑milestone dependency table** (prior milestones + Mathlib imports + new infra) and **per‑layer compile‑before‑summit examples**.
- **Name‑audit against Mathlib conventions** before implementation (flagged `LInfEquivW`).
- Add **bridge lemmas**: `Countable (Σ n, L.Relations n) ↔ L.card ≤ ℵ₀`, and `Countable M ↔ Structure.CG` (to mesh with `equiv_between_cg`).

### Claude
- **The L10(a) vs L10(b) seam:** the bounded‑Scott‑height dichotomy (`counting_dichotomy_of_bounded_scottHeight`) needs only Borel‑iso + Silver‑for‑Borel and is reachable **even if the effective track slips**; full Morley (`morley_counting`) and G₀ need the harder machinery. State this as the milestone boundary so the DST track has a reachable sub‑summit.
- House‑pattern compliance audit; per‑layer ⚠ API‑warning quality is unusually high.

---

## Recommended actions (consensus, ordered)

1. **Retitle/cut v1 to "Core Lω₁ω & Scott Analysis" = Layers 0–2 + 11**, with an explicit *"not in scope for v1: Barwise, Hanf, Morley counting, Lκλ, effective."*
2. **Promote the coded‑formula bridge to its own named milestone group** with acceptance examples — it is a subproject, not a bullet.
3. **Spin off Erdős–Rado** (standalone combinatorics) and **eject admissible sets** (dedicated KP roadmap).
4. **Pick one counting route** (proxy *or* game‑counting); the other becomes a provenance note.
5. **Add a dependency table** per milestone and **examples that compile using only prior layers.**
6. **Name‑audit** against Mathlib conventions before the pod starts.
7. **Vet the Layer‑0 syntax + `FormulaCode` encoding by hand** before any agent touches it — every later theorem inherits those choices, and L12 may otherwise force a redesign.

---

## Methodology note

Purely AI review, three independent models given the same artifact (the full roadmap README at head `221cde7`) and the same structural rubric (library‑building, generality bar, unbundled hypotheses, correct Mathlib reuse, dependency ordering, naming/placement). Codex additionally ran a live Mathlib‑docs verification pass on the consume/build claims. No proofs were attempted; this assesses roadmap *structure* only. Convergence across all three on the scope/split diagnosis is the headline.

cameronfreer:
Thanks, @kim-em and @mrdouglasny -- these reviews are helpful. I've narrowed this PR rather than trying to patch the whole long-range program in place.

The active roadmap now contains just the spine of results along the way to Scott isomorphism:
- Lω₁ω / L∞ω syntax and semantics;
- the finitary embedding and realization lemmas;
- back-and-forth / potential isomorphism / Karp;
- the coded-formula proxy and refinement-counting bridge;
- Scott rank, Scott sentences, and Scott's isomorphism theorem.

Everything beyond this is removed from the roadmap entirely (admissible sets / Barwise, invariant DST, Silver / G₀ / Glimm–Effros, Morley counting, EM / Erdős–Rado / Morley–Hanf, many-sorted, Lκλ, effective Scott analysis, and relationalization of functions/constants). These can be grouped into their roadmap PRs later with appropriate grounding details and checks for WIP.

I also folded in some review points that still apply to what remains:
 - Layer 0's recursion/substitution/relabeling/universe API is now explicit
 - `FormulaCode` is the chosen counting route
 - there's a per-layer dependency/acceptance table
 - the bridge-lemma names are target shapes pending a Mathlib name audit.

cameronfreer:
@kim-em wrote:
> Given how much of this is live, I suspect merging this roadmap may need to wait until Tau Ceti is public — coordinating with these efforts will be much easier once the Zulip channel is public.
> 
> This is sufficiently far outside my own expertise that I'm not the right person to give it a final sign-off. We'll need to find another reviewer with real model-theory / DST background (waiting for release should help there too — several of the people in the threads above would be ideal). :-)

Sounds good; happy to wait! =]

cameronfreer:
Important correction since the previous review: `PotentialIso` was defined using Mathlib's global `IsExtensionPair` predicate, which quantifies over **all** finitely generated partial equivalences and is strictly stronger than classical potential isomorphism. For example, `(ℕ, <)` is isomorphic to itself, but the partial equivalence `1 ↦ 0` cannot be extended to include `0` in its domain — so with the old definition, `karp_theorem` and `countable_potentialIso_iff_iso` were both false as stated (only the one direction previously written in Lean, `PotentialIso → iso`, was true, which is why the file compiled).

`PotentialIso` is now defined by the existence of a nonempty back-and-forth system `S`, with `potentialIso_of_isExtensionPair` retaining the one-way `S = univ` bridge to Mathlib's vocabulary. The countable milestone is now the correct iff `countable_potentialIso_iff_iso`, and the basic API is pinned alongside (`potentialIso_of_equiv`, `PotentialIso.symm`, `PotentialIso.trans`, with both halves of the transitivity back-and-forth recorded in the docstring). The counterexample is documented in the definition's docstring and a Layer-1 API warning in the README. Full `lake build` is green.

---

🤖 written by Claude, reviewed by Codex, directed by Cameron

cameronfreer:
I've been revising the foundations of this material in response to the zulip discussion [#mathlib4 > ModelTheory: API for infinitary formulas of L_{∞,ω} @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/ModelTheory.3A.20API.20for.20infinitary.20formulas.20of.20L_.7B.E2.88.9E.2C.CF.89.7D/near/613390601), and I'll revise this accordingly once I finish trying out the ideas in the downstream [infinitary-logic](https://github.com/cameronfreer/infinitary-logic).

cameronfreer:
> I've been revising the foundations of this material in response to the zulip discussion [#mathlib4 > ModelTheory: API for infinitary formulas of L_{∞,ω} @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/ModelTheory.3A.20API.20for.20infinitary.20formulas.20of.20L_.7B.E2.88.9E.2C.CF.89.7D/near/613390601), and I'll revise this accordingly once I finish trying out the ideas in the downstream [infinitary-logic](https://github.com/cameronfreer/infinitary-logic).

I've now revised this in light of suggestions in that zulip thread by @plp127 for defining `BoundedFormulaInf L ι α n`, with `Lω₁ω` its definitional specialization (with `ι := ℕ`). I've tested this in  [cameronfreer/infinitary-logic#43](https://github.com/cameronfreer/infinitary-logic/pull/43); it seems to work well there, and I've backported the results to this roadmap.

CBirkbeck (changes requested):
🤖 Review prepared with ChatGPT (GPT-6 Astra Pro), posted at the maintainer's request.

# Review of TauCetiRoadmap PR #41

**Assessment: request changes.**

**Head reviewed:** `a12a6a03973a41d2755df5b7c8444d9a44f18ee9`.

I reviewed the current README and Suggested.lean, the earlier discussion, the relevant pinned Mathlib partial-equivalence interface, and the exact migration checkpoint `06bd522586f97ccf3503b7f47f250943b33a7f99`. The narrowed scope and the fixed-carrier syntax are sensible. The explicit back-and-forth-system definition of potential isomorphism, rather than identifying it with Mathlib's stronger IsExtensionPair, is an important correction worth retaining.

However, Layer 2 still mandates a false countable-code bridge as the critical path to Scott's theorem. There is also a false finite-structure acceptance statement, and the Scott-analysis interfaces need to be made definite. These require changes before approval; they do not require reinstating the discarded Barwise/DST/partition-calculus layers.

## 1. The finite-list code fragment cannot capture the claimed back-and-forth equivalence

**Location:** README, Standing hypotheses / Countability via the coded proxy; Layer 2; Library spine; Ordering.

Layer 2 asks for a countable `FormulaCode L n` using explicit list branching, interpreted into Lω₁ω, and for the implication `agree_codes_implies_BFEquiv` (with its converse) to capture infinitary back-and-forth equivalence. The relevant distinction is not simply that raw infinitary syntax is uncountable: it is that replacing its branching functions by inductive finite lists removes the infinitary expressive power needed by this implication.

For the advertised constructors, every code is equivalent to a finitary first-order formula. A `nil`/`cons` inductive list is finite, even when called a code for a countable conjunction. Structural induction translates its conjunctions and disjunctions to finite Boolean combinations. Neutral padding with infinitely many copies of true or false does not change this.

### Explicit countable counterexample, already at the first back-and-forth step

Let the language have unary relation symbols `P_i`, one for each `i ∈ ℕ`, and no other nonlogical symbols. Let

$$
M=\operatorname{Finset}(\mathbb N),\qquad
P_i^M(s)\Longleftrightarrow i\in s.
$$

Let `N` be obtained by adjoining one new element `*`, with every `P_i(*)` true; retain the same interpretations on the old elements. Both structures are countable, and the language satisfies the roadmap's countable relational hypotheses.

For every finite set `F` of predicate symbols, the reducts of `M` and `N` to `F` are isomorphic. In `M`, each Boolean membership pattern on `F` is realized by countably infinitely many finite subsets, obtained by varying the finite subset outside `F`. In `N`, precisely one of those already-countably-infinite cells receives one additional element. The cell cardinalities do not change, so a cellwise bijection is an isomorphism of the reducts.

Every first-order sentence mentions only finitely many predicates. Thus `M` and `N` agree on every first-order sentence, and hence on every sentence code built from the proposed finite-list constructors.

But

$$
N\models\exists x\,\bigwedge_{i\in\mathbb N}P_i(x),
\qquad
M\not\models\exists x\,\bigwedge_{i\in\mathbb N}P_i(x).
$$

The back-and-forth condition already fails at stage 1 on the empty tuples: the element `*` has no counterpart in `M` with the same complete atomic unary type. This also refutes the proposed bridge when agreement is restricted to codes of quantifier rank at most 1, since the structures agree on all the codes anyway.

The missing step is exactly the invalid interchange of witnesses: satisfying every finite approximation does not supply a single element satisfying the entire countable conjunction.

### This is not repaired by an unspecified richer but still uniform countable fragment

The signature `FormulaCode L n` and its interpretation depend only on the language and arity, not on a chosen structure. Even allowing a countable collection of genuine infinitary formulas does not yield a single uniform fragment characterizing all countable structures up to back-and-forth equivalence/isomorphism.

Indeed, enumerate the sentence interpretations as `φ_i`. Their infinitary complexities have a common countable bound. For each structure, the countable conjunction of the `φ_i` it satisfies and the negations of the others has a uniform countable complexity bound as well. If agreement on this family characterized isomorphism among countable structures, these truth-profile conjunctions would be Scott sentences with a uniform countable complexity bound for every structure of the language. This conflicts with unbounded Scott complexity even in the fixed language of linear orders; see Gonzalez–Harrison-Trainor–Ho, *Scott analysis, linear orders and almost periodic functions*, [arXiv:2406.01836](https://arxiv.org/abs/2406.01836).

A countable fragment chosen **for a particular structure**, or a properly scoped bounded fragment, is a different and potentially valid interface. The universally sufficient language-only countable fragment currently advertised is not.

### The exact cited migration source has already moved away from this route

At the roadmap's own checkpoint:

- [Scott/Code.lean](https://github.com/cameronfreer/infinitary-logic/blob/06bd522586f97ccf3503b7f47f250943b33a7f99/InfinitaryLogic/Scott/Code.lean) labels the module legacy/off-path. Its actual bridge declaration is the forward `BFEquiv_implies_agree_codes`; the requested converse is absent. Its constructors are exactly the finite `FormulaCodeList.nil`/`cons` construction described above.
- [Scott/RefinementCount.lean](https://github.com/cameronfreer/infinitary-logic/blob/06bd522586f97ccf3503b7f47f250943b33a7f99/InfinitaryLogic/Scott/RefinementCount.lean) instead starts from `exists_complete_self_stabilization` and uses the internal-stabilization / constant-witness-chain argument. It does not obtain refinement counting from the claimed code converse.

This is a mismatch against the specified checkpoint, not a request to chase an unspecified future version of the source. These observations are about the inspected interfaces and proof route; I have not audited every proof in the source repository.

**Required change:** remove the false code-to-BF implication and remove it from the critical dependency path. Finite-list codes may remain as a finitary fragment if useful, with the honest forward preservation theorem. Replace the Scott counting route by a correctly quantified structure-relative stabilization argument.

## 2. A concrete replacement for Layer 2, and the interfaces Layer 3 needs

The standard internal counting argument gives a direct replacement without expanding the roadmap's scope.

For a fixed countable structure `M`, form the countable set

$$
X_M=\coprod_{n<\omega}(M^n\times M^n).
$$

Let `R_α` be the relation/set of tuple pairs satisfying `BFEquiv α`, with **both tuples in M**. Prove the successor recursion, monotonicity, and the limit-intersection law. If the relation changed strictly at every countable successor stage, choosing a pair lost at each such stage would give distinct elements of the countable set `X_M` for uncountably many stages. Therefore there is an `α < ω₁` at which all tuple sizes stabilize simultaneously:

$$
\forall n\;\forall\bar a,\bar b\in M^n,
\quad
\mathrm{BFEquiv}_\alpha(\bar a,\bar b)
\Longleftrightarrow
\mathrm{BFEquiv}_{\alpha+1}(\bar a,\bar b).
$$

The common fixed-point condition then propagates by transfinite induction.

The restriction to internal pairs is essential. Tuples from an arbitrary varying external structure `N` are not elements of one countable set merely because each individual `N` is countable. If external refinement counting is retained as a target, give the additional internal-to-external argument explicitly, as the cited `RefinementCount.lean` attempts via stabilization and `α + ω`; do not assert its countability directly from the cardinality of `M`.

### Canonical formulas and the Scott sentence

After the internal stabilization target, construct the genuine countably branching formulas `θ^M_{α,ā}` with the characteristic theorem

$$
N\models\theta^M_{\alpha,\bar a}(\bar b)
\quad\Longleftrightarrow\quad
\mathrm{BFEquiv}_\alpha(\bar a,\bar b).
$$

The zero stage is the complete atomic and negated-atomic diagram, countable because the relational language is countable. The successor stage retains the previous formula and has the two extension clauses

$$
\bigwedge_{c\in M}\exists y\,\theta^M_{\beta,\bar a c}(\bar x,y),
\qquad
\forall y\,\bigvee_{c\in M}\theta^M_{\beta,\bar a c}(\bar x,y).
$$

At a countable limit, conjoin the earlier stages. All of these index families fit the fixed carrier `ℕ` by the proposed coding API.

One explicit Scott-sentence route at an internally stabilizing `α` is

$$
\theta^M_{\alpha,\varnothing}\ \wedge\
\bigwedge_{n<\omega}\bigwedge_{\bar a\in M^n}
\forall\bar x\,
\left(\theta^M_{\alpha,\bar a}(\bar x)
\rightarrow\theta^M_{\alpha+1,\bar a}(\bar x)\right).
$$

Internal stabilization makes this true in `M`. In any countable model `N` of it, the tuples realizing these formulas determine a nonempty back-and-forth system; the existing `S`-relative countable back-and-forth milestone then gives `M ≃[L] N`. This exhibits the missing connection to the summit without a universal countable formula-code family.

The roadmap need not adopt this exact presentation if it spells out another valid route. It does need to choose a true route and expose its interfaces.

### Pin the conventions and add representative signatures

`Suggested.lean` currently contains only Layer 0/1 interfaces; the entire disputed Layer 2/3 path is outside the checked signatures. It is not necessary to prototype every lemma, but the replacement should include at least the actual BF recursion, internal stabilization, canonical-formula realization equivalence, and Scott-sentence characterization. In particular, `scottFormula_realize_self` alone is not a substitute for the equivalence with BF types in arbitrary comparison structures.

Also define precisely what `scottRank` and `scottHeight` mean. The README asks contributors to use one convention but does not select its defining condition: least internal stabilizing ordinal, an external stabilization notion, or a supremum of tuple ranks, and the location of any `+1`. The inspected source's `Rank.lean` and `Sentence.lean` already illustrate distinct element-rank, internal, and external interfaces. A citation to that source does not settle the intrinsic roadmap convention. Give the chosen definition, its comparison theorem, and at least one numerical rank acceptance check.

Specify the BF successor formula precisely, including retention of the preceding relation. This matters particularly because empty structures and nullary relations are not excluded: merely writing the two extension clauses could make the successor condition vacuous on empty structures while their nullary atomic diagrams differ.

**Required change:** supply the replacement Layer 2/3 signatures and the chosen rank convention as part of the repair, rather than leaving their meaning to the eventual implementer.

## 3. Finite structures need not have first-order Scott sentences in an infinite language

**Location:** Worked examples, first item.

The roadmap says that a finite structure's Scott sentence is equivalent to a first-order sentence. This is false under the standing assumption of a merely countable relational language.

Use the language `{P_i : i ∈ ℕ}` of unary predicates again. Let `A` be a one-element structure in which every `P_i` is false. Any first-order sentence `σ` true in `A` mentions only finitely many of the predicates. Choose `P_j` not mentioned by `σ`, and change its interpretation to true on the same singleton. The resulting one-element structure `B` still satisfies `σ`, but it is not isomorphic to `A`. Thus no first-order sentence can be a Scott sentence for `A`, even when only countable comparison structures are considered.

There is an infinitary Scott sentence for `A`: assert that there is exactly one element and that every `P_i` is empty. Its finite quantifier rank does not make its infinite conjunction first-order.

**Required change:** separate the two acceptance statements. The finite-rank assertion can remain at the intended countable-language generality; the first-order Scott-sentence assertion needs a **finite relational language**, or another explicit sufficient hypothesis. Add the singleton/countably-many-unary-predicates example as a regression test for the distinction.

## Smaller scope and metadata corrections

The Standing hypotheses paragraph says `[Countable M]` and language countability belong on every Karp statement. But the actual `karp_theorem_at` is intentionally more general, and the worked examples include uncountable potentially isomorphic structures. Keep the useful general Karp signature and rewrite the prose to distinguish its scope from the countability assumptions needed for Scott's theorem.

The PR description still lists a root index entry and a Suggested import, although the current diff contains only the two InfinitaryLogic files. Update the description; do not reintroduce the generated-file edits removed by the latest commit.

## What should be retained

The narrowed syntax → back-and-forth/Karp → Scott scope is appropriate. The single fixed-carrier inductive, definitional `ι := ℕ` specialization, carrier-independent finitary embedding, explicit codings with neutral padding, and semantic transport are coherent choices. The existential back-and-forth-system definition of `PotentialIso` and its one-way bridge from Mathlib's stronger `IsExtensionPair` correctly avoid the earlier ultrahomogeneity problem. The common-carrier formulation of Karp is also appropriate. I found no reason in this pass to discard that Layer 0/1 design.

## Validation and recommendation

GitHub CI for the reviewed head reports success: [run 32431022221](https://github.com/TauCetiProject/TauCetiRoadmap/actions/runs/32431022221), completed on 21 August 2026. The manifest pins Mathlib `de5ce8a9a66a4aa68a9bdbb35b63a06d34d9ca11`, Tau Ceti `86cc55d9192fc3f094f293ead6f2e7a153c79d17`, and Lean v4.34.0-rc1. I have not run a fresh local Lean build or formalized the counterexamples. The green build checks the existing prototype file, not the truth of the missing Layer 2/3 specifications.

**Request changes:** replace the false countable-code route, expose the correct Scott-analysis/rank interfaces, and restrict the finite-structure first-order example appropriately. The fixed-carrier syntax and Karp design can remain. This is a substantive repair of the Scott proof route and should receive a further review after revision, rather than approval after only cosmetic edits.

PR #45, opened 2026-06-27
OrthogonalL2Bases: add the weight↔measure isometry (both normalizations) + Targets.lean

mrdouglasny:
A small **additive** enhancement to the `OrthogonalL2Bases` roadmap (v1 kept intact, no API dropped), plus the `Targets.lean` promised at the #25 merge. Intention/claim: **#46**.

## What & why

One orthogonal family gives two bases depending on where the weight sits: the bare polynomials `Hₙ/√(n!)` as an ONB of the **weighted / probability** `L²` (e.g. `N(0,1)` — expansions, moment problems, chaos), or the `Hₙ·√w` functions as an ONB of **Lebesgue** `L²(dx)` (Hermite functions / QHO states, PDE, signal processing). The two are isometric — related by `weightL2Isometry : L²(w·μ) ≃ₗᵢ L²(μ)` (multiplication by `√w`) — but are different `HilbertBasis` objects on different spaces. This adds that isometry (plus the `HilbertBasis.mapₗᵢ` transport Mathlib lacks) so both normalizations come from one construction; neither is privileged.

*(Note: for a family on a rescaled argument like Hermite's `Hₙ(x√2)`, the function-side basis is the isometry image **plus a `u=x√2` dilation** — the `N(0,1)` basis maps to `2^{-1/4}·ψₙ(·/√2)`, the dilated Hermite functions, not `hermiteHilbertBasis` directly. The README states this explicitly.)*

## Additions

- **Part 0:** `weightL2Isometry` (+ `weightL2Isometry_apply`/`_symm_apply`), `HilbertBasis.mapₗᵢ` (+ `mapₗᵢ_apply`).
- **A1:** Gaussian-measure orthogonality `integral_hermite_mul_hermite_gaussianReal` alongside the Lebesgue form.
- **B1:** measure-level determinacy `ae_eq_zero_of_forall_moment_eq_zero_of_finite_expMoments`.
- **B2:** `barePolyLp`, `orthonormal_barePolyLp`, `barePolyLp_ortho_eq_bot` (grounded on the `hexp` moment-determinacy hypothesis), `hilbertBasisOfWeightedMeasure` + derived `hilbertBasisOfOrthogonalSystem`, with `coe_*` pins.
- **A3′:** `gaussianHermiteHilbertBasis` (+ `coe_*`, `memLp_hermite_gaussianReal`).
- **B3 / Part D:** product/`pi` bases + `gaussianHermitePiBasis`.

v1's Hermite-function API (A2), function-side `hermiteHilbertBasis`, the B1–B3 spine, and Chebyshev (Part C) are unchanged.

## Scope vs. #46

`Targets.lean` seeds **representative target signatures** for the core layers, which *includes* the B1 determinacy and B3 product/`pi` signatures. The **implementation** claim in intention #46 is narrower (Part 0 / A1 / B2 / A3′); the B1/B3 *signatures* here are stated so the file is self-contained, not claimed as implementation work.

## Status

Rebased on `main`; root `lake build TauCetiRoadmap` green on the pin (v4.31.0-rc1), `sorry`-warnings only (the new `Targets.lean` is wired into the root aggregator). A1 polynomial API + Lebesgue orthogonality is implemented in FormalFrontier/TauCeti#499 (held as draft). Review feedback (external 3-pass + our own Codex pass) addressed across the history.

### DIFF
--- a/TauCetiRoadmap.lean
+++ b/TauCetiRoadmap.lean
@@ -13,3 +13,4 @@ import TauCetiRoadmap.HeegaardFloer.Targets
 import TauCetiRoadmap.GeometricTopology.Targets
 import TauCetiRoadmap.Exchangeability.Targets
 import TauCetiRoadmap.ConformalMapping.Targets
+import TauCetiRoadmap.OrthogonalL2Bases.Targets
--- a/TauCetiRoadmap/OrthogonalL2Bases/README.md
+++ b/TauCetiRoadmap/OrthogonalL2Bases/README.md
@@ -29,7 +29,7 @@ The **Hermite basis of `L²(ℝ)`** is the worked anchor (**Part A**, the v1 del
 family-agnostic spine (**Part B**) is abstracted. The spine is then **exercised by a second family —
 the Chebyshev basis of `L²([-1,1])`** (**Part C**), whose orthogonality relation is already in
 Mathlib and which tests the bridge on a compact weighted measure (not the Gaussian); and it
-yields the **multidimensional Hermite basis** (**Part D**, future). Laguerre and Jacobi are a
+yields the **multidimensional Hermite basis** (**Part D**, a later milestone here). Laguerre and Jacobi are a
 *separate* future roadmap — Mathlib has neither family, so grounding them means defining the
 polynomials first.
 
@@ -58,6 +58,12 @@ L²") is this area's contribution.
 - **The basis layer is family-agnostic and scalar-generic.** Bases are stated through the
   measure-generic `HilbertBasis ι 𝕜 (Lp 𝕜 2 μ)` API over `[RCLike 𝕜]` (the real pointwise
   functions cast via `algebraMap ℝ 𝕜`); do **not** duplicate the API for `ℝ` and `ℂ` separately.
+- **Pin the domain.** The polynomial bridge (B2) and the Hermite/Chebyshev instances fix the
+  reference measure to **`μ : Measure ℝ`** — the bridge evaluates `Polynomial.eval` (needs `ℝ`), and
+  `ℝ` is where the compact-support and interval families (and the Laguerre/Jacobi-on-subsets follow-on)
+  live. The `weightL2Isometry` primitive itself is **not** specialized — it is measure-theoretic and
+  stated over an arbitrary measurable space; only the polynomial-facing layers are `ℝ`. Product/`pi`
+  bases (B3) are over generic measurable factors.
 - **Every basis milestone exports its element-level `coe_*`, not just the bundled term.** A
   `HilbertBasis` shipped only as a bundled term — without `⇑basis = explicit family` — is
   near-vacuous: bare existence of an `ℕ`- (resp. `ι×κ`-) indexed Hilbert basis is just a
@@ -104,7 +110,7 @@ L²") is this area's contribution.
 
 ## What is missing (build here)
 
-The completeness toolkit (both routes), the relation→`HilbertBasis` bridge, and the L²-product-basis
+The completeness toolkit (the moment-determinacy mechanism), the relation→`HilbertBasis` bridge, and the L²-product-basis
 lemma; plus their instances: the Hermite basis (Part A) and the Chebyshev basis (Part C), and the
 multidimensional Hermite basis (Part D). (Laguerre/Jacobi are a separate future roadmap.)
 
@@ -182,8 +188,8 @@ analytic facts (A2–A3) live under `Analysis/SpecialFunctions/…`. Names descr
   `‖f‖² = ∑' n, ‖⟪ψₙ,f⟫‖²`; both `ℝ` and `ℂ` instantiate.
 
 ### A3′ — The Gaussian Hermite basis (measure side; the enhancement's named instance)
-The `weightL2Isometry`-partner of `hermiteHilbertBasis`, on the **probability measure** rather than
-`L²(dx)` — the ONB any `L²(N(0,1))` expansion is taken against.
+The bare-polynomial ONB on the **probability measure** rather than `L²(dx)` — the ONB any `L²(N(0,1))`
+expansion is taken against.
 - **`gaussianHermiteHilbertBasis 𝕜 : HilbertBasis ℕ 𝕜 (Lp 𝕜 2 (gaussianReal 0 1))`** — the bare
   normalized `Hₙ/√(n!)`; the immediate instance of `hilbertBasisOfWeightedMeasure` (B2) with
   `μ = volume`, `w = gaussianPDFReal 0 1`, `cₙ = n!` (since `gaussianReal 0 1 = volume.withDensity
@@ -192,8 +198,19 @@ The `weightL2Isometry`-partner of `hermiteHilbertBasis`, on the **probability me
   `⇑(gaussianHermiteHilbertBasis n) =ᵐ[N(0,1)] fun x => algebraMap ℝ 𝕜 (aeval x (hermite n)/√(n!))`.
 - **`memLp_hermite_gaussianReal (n) (v : ℝ≥0)`** — variance-general `L²` membership of `Hₙ/√(n!)`
   under `gaussianReal 0 v` (the `Hₙ` of a centered Gaussian of any variance).
-- *Acceptance:* `⟨H₀,H₀⟩=⟨H₁,H₁⟩=1`, `⟨H₀,H₂⟩=0` under `N(0,1)`; it is the `mapₗᵢ`-image of
-  `hermiteHilbertBasis` under `weightL2Isometry`.
+- **Relation to the Lebesgue Hermite-function basis — note the dilation.** Applying
+  `weightL2Isometry` (multiplication by `√(gaussianPDFReal 0 1) = (2π)^{-1/4} e^{-x²/4}`) to
+  `gaussianHermiteHilbertBasis` does **not** give the A3 `hermiteHilbertBasis` *as is*: the image is
+  `x ↦ (2π)^{-1/4} Hₙ(x) e^{-x²/4} / √(n!) = 2^{-1/4} · ψₙ(x/√2)`, the **dilated** Hermite-function
+  basis, because v1's `ψₙ(x) = (n!√π)^{-1/2} Hₙ(x√2) e^{-x²/2}` is built on the rescaled argument
+  `x√2`. So the two named bases are related by `weightL2Isometry` **plus the `u = x√2` dilation**, not
+  by the isometry alone. Pin this with a target
+  `weightL2Isometry_gaussianHermiteHilbertBasis_apply` stating the image is `x ↦ 2^{-1/4} ψₙ(x/√2)`
+  (or, equivalently, a `dilation`-equivalence target relating the two). (One could instead base the
+  weighted side on `w = e^{-x²}`, `pₙ = Hₙ(·√2)`, i.e. `N(0,½)`, to make the image exactly `ψₙ`; we
+  keep the standard `N(0,1)` normalization that consumers expect and carry the dilation explicitly.)
+- *Acceptance:* `⟨H₀,H₀⟩=⟨H₁,H₁⟩=1`, `⟨H₀,H₂⟩=0` under `N(0,1)`; and the dilation relation above
+  holds (image `= 2^{-1/4} ψₙ(·/√2)`).
 
 ## Part B — The family-agnostic spine (the reusable layer)
 
@@ -247,18 +264,25 @@ but undischargeable downstream.
 
 **Both normalizations via the weight↔measure isometry (the enhancement).** Two new primitives — both
 genuine Mathlib gaps — make B2's basis available on either side without re-proof:
-- **`weightL2Isometry (μ : Measure ℝ) (w) (hwpos) (hwm) : L²(w·μ) ≃ₗᵢ[𝕜] L²(μ)`**, multiplication by
+- **`weightL2Isometry (μ : Measure α) (w) (hwpos) (hwm) : L²(w·μ) ≃ₗᵢ[𝕜] L²(μ)`**, multiplication by
   `√w` (`w·μ := μ.withDensity (ENNReal.ofReal ∘ w)`); an *equivalence* exactly because `0 < w` a.e.
-  (`‖√w·f‖²_{L²(μ)} = ∫ w|f|² = ‖f‖²_{L²(w·μ)}`).
+  (`‖√w·f‖²_{L²(μ)} = ∫ w|f|² = ‖f‖²_{L²(w·μ)}`). The isometry is purely measure-theoretic, so stated
+  over an **arbitrary** measurable `α` (only the polynomial bridge below needs `Measure ℝ`); it ships
+  the element-level `weightL2Isometry_apply` (a.e. `= √w · f`) as its anti-vacuity pin. Both this and
+  `mapₗᵢ` are general-purpose and are flagged as **upstream-Mathlib candidates**.
 - **`HilbertBasis.mapₗᵢ (b) (e : E ≃ₗᵢ F) : HilbertBasis ι 𝕜 F`** with `@[simp] mapₗᵢ_apply`
   (`ofRepr (e.symm.trans b.repr)`; Mathlib has `ofRepr` but no `≃ₗᵢ`-transport).
 
 So alongside the `√w`-envelope basis of `L²(μ)` above, the bridge also yields the **bare-polynomial
 basis of the weighted measure** `hilbertBasisOfWeightedMeasure : HilbertBasis ℕ 𝕜 (Lp 𝕜 2 (w·μ))` (the
 `pₙ/√cₙ`, with its own `coe_*` — orthogonal polynomials are *natively* an ONB of their weighted `L²`,
-so this needs no rescaling). Either basis is the `mapₗᵢ`-image of the other under `weightL2Isometry`;
-a consumer picks the normalization their space wants. (Chebyshev, Part C, lands its basis directly on
-the weighted `measureT` this way; the Gaussian instance is A3′.)
+so this needs no rescaling). The two are `mapₗᵢ`-images of each other under `weightL2Isometry` — **up
+to the rescaling-argument change of variables noted above**: for a family defined on a rescaled
+argument (Hermite's `Hₙ(x√2)`) the function-side basis carries a dilation, so the bare isometry image
+of the weighted-measure basis is the *dilated* function basis, not the un-dilated one (see A3′). A
+consumer picks the normalization their space wants. (Chebyshev, Part C, lands its basis directly on
+the weighted `measureT` this way — no dilation, since `Tₙ` is not on a rescaled argument; the Gaussian
+instance is A3′.)
 
 ### B3 — L²-product basis
 **The load-bearing milestone is completeness, not orthonormality.** Orthonormality of the products
@@ -314,9 +338,11 @@ rather than the Gaussian — so it tests the σ-finite-`μ` genericity of the br
   `integral_measureT_eq_integral_cos` + a unitary-transfer statement, not from `T_real_cos` alone)
   `chebyshevHilbertBasis` corresponds to the cosine basis under `x = cos θ`.
 
-## Part D — The multidimensional Hermite basis (future; consume B)
+## Part D — The multidimensional Hermite basis (a later milestone of this roadmap; consumes B3)
 
-*Not v1; recorded so Parts A/B are built generically enough to instantiate. Fully grounded — `B3 ∘ A`.*
+*A grounded, sequenced milestone — not "future"/optional: it consumes only B3 (which lands in this
+enhancement) and `A`/`A3′`, and its targets `piHilbertBasis` / `gaussianHermitePiBasis` are stated in
+`Targets.lean`. Sequenced after B3, but in scope here.*
 
 - **Multidimensional Hermite basis** of `L²(ℝᵈ)` — `Ψ_α(x) = ∏ᵢ ψ_{αᵢ}(xᵢ)`, `α : ι → ℕ` — the
   immediate `B3 ∘ A` instantiation: B3 over `ι` copies of the 1-D Hermite basis (A3), a
--- a/TauCetiRoadmap/OrthogonalL2Bases/README.md
+++ b/TauCetiRoadmap/OrthogonalL2Bases/README.md
@@ -14,7 +14,16 @@ This area builds that **L² Hilbert-basis layer** for orthogonal systems:
 - a **completeness** toolkit (moment determinacy / Fourier uniqueness) — the step that upgrades
   "orthogonal" to "complete orthonormal basis";
 - the **orthogonality-relation → `HilbertBasis`** bridge (the √w normalization + Parseval);
+- the **weight↔measure isometry** `L²(w·μ) ≃ₗᵢ L²(μ)` (multiplication by `√w`), so each family's
+  basis is available in **both normalizations** — the bare polynomials as an ONB of the weighted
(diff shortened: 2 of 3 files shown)


mrdouglasny:
@kim-em @CBirkbeck — could one of you give this an approve when you have a moment? I'm the author so I can't self-approve, and review is already requested from `roadmap-reviewers`. Tracking intention: #46.

**What this is.** A small *additive* enhancement to the `OrthogonalL2Bases` roadmap (v1 kept intact, no API dropped) plus the `Targets.lean` promised at the #25 merge. It re-keys everything around one primitive — the **weight↔measure isometry** `weightL2Isometry : L²(w·μ) ≃ₗᵢ L²(μ)` (multiplication by `√w`) — so that the **weighted-measure** basis (bare `Hₙ/√(n!)` as an ONB of `L²(N(0,1))`, the object downstream Wiener-chaos consumers actually need) is the named target, and the original **L²(Lebesgue) Hermite-function** basis is its one-line `.map` image. Both normalizations fall out of one construction; Laguerre/Jacobi/multi-d then instantiate without re-proof.

**Already reviewed — two independent passes, folded in:**
- A roadmap-design review endorsed the re-keying (confirmed `gaussianReal 0 1 = volume.withDensity (gaussianPDF 0 1)` makes the Gaussian instance immediate; the `√(2π)` reconciliation; `weightL2Isometry` soundness under `hwpos`).
- **Codex/GPT-5.4**, which **elaboration-checked it against Mathlib v4.30**. Its fixes are folded in: `μ : Measure ℝ` (generic `α` broke `Polynomial.eval`; `ℝ` also matches the Laguerre/Jacobi-on-subsets design), `ℤ[X]`→`ℝ[X]` via `hermiteℝ`, `HilbertBasis.mapₗᵢ := ofRepr (e.symm.trans b.repr)` (confirmed a genuine Mathlib gap, not droppable), a split assembler (`barePolyLp` / `orthonormal_barePolyLp` / `barePolyLp_ortho_eq_bot`, with `hdeg` moved out into the completeness target), and `coe_*` / `*_apply` pins against vacuity. It now elaborates with **0 errors / 16 `sorry`-targets** under v4.30.

**CI.** `build` passes; `intentions/lifecycle` passes. The `enable` check fails, but that's the fork-PR app-token config (`actions/create-github-app-token` missing `app-id` on fork-triggered runs) — infrastructure, not content; the required check `build` is green.

Happy to adjust anything. Thanks!

cameronfreer:
I'm not familiar with the mathematical content, but here are some comments from GPT 5.5 in Codex (+ lean4-skills + lean-lsp-mcp) that might be helpful.

----

 Findings

  1. Blocking: barePolyLp_ortho_eq_bot is not grounded as stated.
     In [Targets.lean#104-L111](https://github.com/mrdouglasny/TauCetiRoadmap/blob/df111ed98bad7999de68f02e96ee9808d6920f18/TauCetiRoadmap/OrthogonalL2Bases/Targets.lean#L104-L111), the theorem claims completeness for arbitrary μ, w, and polynomial family p from hwpos, hc, hdeg, and hmem. That is too weak. The README says this uses moment determinacy / B1, but the Lean target does not carry the hypotheses that make B1 applicable. This is exactly the repo’s “connected to the ground” issue. I’d either add the real determinacy/completeness hypotheses or make this theorem specialized to the Gaussian/Chebyshev cases.

  2. Important: weightL2Isometry needs an element-level characterization, not just an opaque equivalence. PR #25’s discussion established that bundled basis objects need coe_* anti-vacuity lemmas. The same applies here: [Targets.lean:41](https://github.com/mrdouglasny/TauCetiRoadmap/blob/df111ed98bad7999de68f02e96ee9808d6920f18/TauCetiRoadmap/OrthogonalL2Bases/Targets.lean#L41) defines the isometry by sorry, and [README.md:248-260](https://github.com/mrdouglasny/TauCetiRoadmap/blob/df111ed98bad7999de68f02e96ee9808d6920f18/TauCetiRoadmap/OrthogonalL2Bases/README.md#L248-L260) relies on it being multiplication by √w. Add targets like weightL2Isometry_apply and probably weightL2Isometry_symm_apply, stating the a.e. formulas.

  3. Medium: the new target file is not imported by TauCetiRoadmap.lean.
     The PR adds TauCetiRoadmap/OrthogonalL2Bases/Targets.lean, but the root file at the PR head still imports only the older target modules: [TauCetiRoadmap.lean:5-14](https://github.com/mrdouglasny/TauCetiRoadmap/blob/df111ed98bad7999de68f02e96ee9808d6920f18/TauCetiRoadmap.lean#L5-L14). The Lake glob may still build the file, but repo convention says the root is the human-owned target-signature root. I’d add import TauCetiRoadmap.OrthogonalL2Bases.Targets.

  4. Low: stale “both routes” wording remains.
     [README.md:74-77](https://github.com/mrdouglasny/TauCetiRoadmap/blob/df111ed98bad7999de68f02e96ee9808d6920f18/TauCetiRoadmap/OrthogonalL2Bases/README.md#L74-L77) says one completeness mechanism; [README.md:107](https://github.com/mrdouglasny/TauCetiRoadmap/blob/df111ed98bad7999de68f02e96ee9808d6920f18/TauCetiRoadmap/OrthogonalL2Bases/README.md#L107) still says “both routes.” Make that “the moment-determinacy completeness toolkit” or similar.

  5. Low/process: the PR body still says the intention/claim is “to be linked,” while the comment says it is #46. Update the PR body before merge.

  Overall, I think the PR is good in spirit and mostly follows the merged-review pattern from #25: it grounds the former “inputs,” adds coe_* deliverables, and keeps the change additive. I would not approve it as-is because of finding 1. Fixing that plus the isometry apply lemmas would make the roadmap much more robust.

cameronfreer:
some comments from Opus 4.8 in Claude Code (+ lean4-skills + lean-lsp-mcp) that might also be useful.

----

Thanks — this is clean and well-grounded, and it reads like the writing guide: genuinely additive (v1 API preserved), the new primitives fill real Mathlib gaps, every target is listed, and each basis ships a named element-level `coe_*` rather than a bundle-only (near-vacuous) term. Using Chebyshev as a second family on a compact weighted measure is the right way to show the spine is family-agnostic, and the "deliberately out of scope" boundary (Gram–Schmidt/Favard, and Laguerre/Jacobi as their own roadmap since Mathlib lacks those families) is exactly the right call. A few things before approving:

1. **Wire `Targets.lean` into the build (blocker).** The PR adds `OrthogonalL2Bases/Targets.lean` but doesn't add `import TauCetiRoadmap.OrthogonalL2Bases.Targets` to the root `TauCetiRoadmap.lean`. Every roadmap that ships a `Targets.lean` is imported there (it was README-only before this); without the import the new 203-line file is orphaned from `lake build`, so the green `build` check isn't actually elaborating it. (#46 is the intention issue, not a wiring PR.) Adding the import here is the standard step.

2. **Confirm it elaborates on the pin, not just v4.30.** The "0 errors / 16 sorry-targets" pass was against Mathlib v4.30, but the repo pins Lean v4.31.0-rc1 / mathlib `9caeba1`. Wiring the import (1) makes CI check it on the real pin — please confirm it's green there; a v4.30-checked file isn't guaranteed on the pinned toolchain.

3. **Part D framing.** Part D ("multidimensional Hermite") is marked "future", but it consumes only B3 (which lands in this PR) and already has `piHilbertBasis` / `gaussianHermitePiBasis` in `Targets.lean` — so it's a grounded, sequenced milestone, not "future." This is the same point you made on the infinitary-logic PR about not leaving things as implied later work: I'd call it a later milestone of this roadmap, or split it to a cited separate roadmap, rather than "future." (Laguerre/Jacobi as a separate roadmap is already handled correctly.)

4. **Upstream candidates.** `HilbertBasis.mapₗᵢ` (transport a Hilbert basis along a `≃ₗᵢ`) and `weightL2Isometry` (`L²(w·μ) ≅ L²(μ)`) are general-purpose and look genuinely Mathlib-worthy — worth a one-line note flagging them as upstream targets, since reusable infrastructure is part of the point.

5. **Generality bar — pin the domain too.** The Generality bar nicely pins the scalar side (family-agnostic, `[RCLike 𝕜]`, no ℝ/ℂ duplication). But `Targets.lean` fixes the domain to `μ : Measure ℝ` (the re-key, for `Polynomial.eval` and to match the Laguerre/Jacobi-on-subsets design) — worth stating that domain choice in the Generality bar too, so it's pinned rather than implicit.

None of these are structural — the roadmap's in good shape. (1) is the only blocker, and (2) falls out of it.

cameronfreer:
some comments from GPT Pro 5.5, in case these are useful too.

----

I would **not approve PR #45 as-is**. The overall direction is good, and the PR is doing the right *kind* of thing for this repo: additive roadmap text plus Lean target signatures, with `sorry` targets allowed for roadmaps. But I see two substantive mathematical/design blockers, plus a few cleanup issues.

## What I checked

PR #45 is open, non-draft, mergeable, and changes only `TauCetiRoadmap/OrthogonalL2Bases/README.md` plus a new `Targets.lean`, with 254 additions and 1 deletion. Its stated purpose is an additive `weight↔measure` enhancement plus the `Targets.lean` follow-up promised after #25.

The repo’s own roadmap-writing rules are directly relevant here: roadmap milestones must be grounded in Mathlib, Tau Ceti, earlier roadmap material, or explicit targets; the roadmap should use Mathlib vocabulary; it should prototype signatures in Lean; and it should pin conventions up front.  Earlier #25 review discussion also pressed exactly these points: no dangling “inputs,” exercise the abstract spine on a second grounded family, keep B3 index-generic, name the load-bearing product-basis completeness theorem, and pin Fourier/Hermite conventions.

Mechanically, CI looks fine for the content: the CI workflow succeeded and the `build` job succeeded; the failed workflow is `auto-merge`, specifically the `enable` job at token minting, matching the author’s note that this is infrastructure rather than a Lean/build failure.

## Blocker 1: the Gaussian-measure basis is **not** the direct `weightL2Isometry` partner of the existing Hermite-function basis

The PR repeatedly says that the standard Gaussian basis on `L²(N(0,1))` is the `weightL2Isometry` partner of the existing `hermiteHilbertBasis`. The README’s new A3′ section says `gaussianHermiteHilbertBasis` is the `weightL2Isometry`-partner of `hermiteHilbertBasis`, and its acceptance criterion says it is the `mapₗᵢ`-image of `hermiteHilbertBasis` under `weightL2Isometry`.  The new B2 prose similarly says the isometry gives both normalizations from one construction.

That is not correct with the existing roadmap’s normalization.

The existing Hermite functions are explicitly

[
\psi_n(x)= (n!\sqrt\pi)^{-1/2},H_n(x\sqrt2),e^{-x^2/2},
]

as the roadmap states.  The new standard-Gaussian measure-side basis is

[
h_n(x)=H_n(x)/\sqrt{n!}\quad\text{in }L^2(N(0,1)).
]

Multiplication by (\sqrt{\gamma(x)}=(2\pi)^{-1/4}e^{-x^2/4}) sends this to

[
(2\pi)^{-1/4} H_n(x)e^{-x^2/4}/\sqrt{n!},
]

which is **not** the existing (\psi_n(x)). It is a dilated Hermite-function basis:

[
(2\pi)^{-1/4} H_n(x)e^{-x^2/4}/\sqrt{n!}
=2^{-1/4},\psi_n(x/\sqrt2).
]

So the standard `N(0,1)` chaos basis is related to the existing Lebesgue Hermite basis by **weight isometry plus a dilation**, not by the weight isometry alone.

This is exactly the kind of pinned-convention issue the earlier PR discussion was careful about. The fix is straightforward: keep `gaussianHermiteHilbertBasis` for `L²(N(0,1))`, but change the prose and targets to say that its `weightL2Isometry` image is the **dilated** Lebesgue Hermite basis. Add a target for the dilation equivalence, or at least a named lemma of the form

```lean
weightL2Isometry_gaussianHermiteHilbertBasis_apply :
  -- image equals x ↦ 2^(-1/4) * hermiteFunction n (x / Real.sqrt 2)
```

If the roadmap really wants the isometry image to be *exactly* the existing `hermiteHilbertBasis`, then the weighted-measure side should be based on the density proportional to `e^{-x^2}` and polynomials `Hₙ(√2 x)`, i.e. a `N(0,1/2)` scaling, not the standard `N(0,1)` basis.

## Blocker 2: `barePolyLp_ortho_eq_bot` is far too general and is mathematically false as stated

The new `Targets.lean` introduces

```lean
theorem barePolyLp_ortho_eq_bot {μ : Measure ℝ}
    (hwpos : ∀ᵐ x ∂μ, 0 < w x) (hc : ∀ n, 0 < c n)
    (hdeg : ∀ n, (p n).natDegree = n)
    (hmem : ∀ n, MemLp ...)
  : (Submodule.span ... )ᗮ = ⊥ := sorry
```

with no moment-determinacy hypothesis, no finite-exponential-moment hypothesis, no density assumption, no family-specific completeness input, and no assumption that `μ,w` are in one of the classical determinate situations.

That cannot be right. Degree growth alone does **not** imply that the polynomials are complete in `L²(w·μ)` for an arbitrary measure and weight. The existing README explicitly says the bridge takes a polynomial-density/completeness hypothesis as an argument, and that B1 is the load-bearing analytic content.   The new target silently promotes “degree `n`” into “complete,” which violates both the mathematics and the repo’s “grounded, with no leaps” rule.

The assembler is actually fine because `hilbertBasisOfWeightedMeasure` takes `hcomplete` explicitly.  The problem is the separate theorem that purports to manufacture `hcomplete` from insufficient assumptions.

I would either remove `barePolyLp_ortho_eq_bot` from the generic bridge targets, or restate it with a real completeness hypothesis. For example, make it a thin conversion theorem from an explicit orthogonal-complement hypothesis, or add a genuinely load-bearing assumption such as a moment-determinacy/density condition specialized enough to be true. The current signature is not a safe target for AI implementors.

## Major issue: `weightL2Isometry` should not be specialized to `ℝ`

The new primitive is advertised as a Mathlib gap: multiplication by `√w` gives

```lean
L²(w·μ) ≃ₗᵢ L²(μ)
```

and the target currently states it only for `μ : Measure ℝ`.  The explanation says this convention came from the bridge needing `Polynomial.eval`.

That justification applies to the polynomial bridge, not to the isometry. The isometry itself is measure-theoretic and should be stated over arbitrary measurable spaces:

```lean
noncomputable def weightL2Isometry
    {α : Type*} [MeasurableSpace α]
    (μ : Measure α) (w : α → ℝ)
    (hwpos : ∀ᵐ x ∂μ, 0 < w x) (hwm : AEMeasurable w μ) :
    Lp 𝕜 2 (μ.withDensity fun x => ENNReal.ofReal (w x)) ≃ₗᵢ[𝕜] Lp 𝕜 2 μ
```

Then keep the B2 polynomial bridge specialized to `Measure ℝ`. This better follows the repo’s “do things right the first time” rule.

## Major issue: the weighted bridge hides a measurability/change-of-measure proof obligation

`hilbertBasisOfOrthogonalSystem` takes `hwm : AEMeasurable w μ` for the isometry, but `orthonormal_barePolyLp` and `hilbertBasisOfWeightedMeasure` do not take `hwm`, even though their orthogonality hypothesis is written as a Lebesgue-side integral

```lean
∫ x, (p m).eval x * (p n).eval x * w x ∂μ
```

while the vectors live in

```lean
Lp 𝕜 2 (μ.withDensity (fun x => ENNReal.ofReal (w x)))
```

To prove orthonormality, the implementor has to rewrite integrals over `μ.withDensity` into integrals with the density `w` against `μ`. That is exactly where measurability of the density belongs. Either add `hwm` to the weighted-measure bridge targets too, or formulate the orthogonality relation directly over the weighted measure and provide a separate lemma connecting it to the `∫ ... * w ∂μ` form.

This is not a mathematical blocker in the same way as the two above, but it is a proof-engineering trap, and this repo’s roadmaps are explicitly supposed to keep AIs from having to guess missing connective tissue.

## Scope / coordination cleanup

Issue #46 is a claimed intention for the weight↔measure enhancement and lists Part 0, A1, B2, and A3′ as in scope, while explicitly saying B1, A2, B3 product/`pi`, and Chebyshev are not claimed.  The claim is active through July 27, 2026.

But `Targets.lean` includes B3 product/pi targets and `gaussianHermitePiBasis`.  I understand why: #25 promised a formal targets file, and the file is seeding representative core layers. Still, the intention issue and PR body should say that explicitly. Otherwise the coordination story reads as “not claiming B3” while the PR adds B3 target signatures.

Also, the PR body still has “Intention/claim for the area: (to be linked)” even though #46 exists. Fix that to `#46`.

## What is good here

The PR’s core idea is strong. Adding a weighted-measure ONB alongside the Lebesgue-envelope ONB is the right abstraction for downstream chaos/probability consumers. The new README text correctly emphasizes that the two bases live in different `L²` spaces and should both be available.  The addition of `HilbertBasis.mapₗᵢ` is also exactly the kind of tiny missing transport lemma that should be in the roadmap, and its proposed implementation via `ofRepr (e.symm.trans b.repr)` is plausible.

The PR also preserves the important earlier decisions from #25: scalar-generic `[RCLike 𝕜]`, named `coe_*` anti-vacuity exports, explicit measures, B3 with σ-finiteness, and Chebyshev as the grounded second family.

## Verdict

I’d request changes, not because the proposal is misguided, but because the two main defects are load-bearing:

1. **The standard Gaussian basis maps to a dilated Hermite-function basis, not the existing `hermiteHilbertBasis`.**
2. **The generic completeness theorem is false without a real completeness/moment-determinacy hypothesis.**

Fix those, generalize `weightL2Isometry` away from `ℝ`, and align the intention/scope text. After that, I would be comfortable with the PR as an additive roadmap enhancement.

mrdouglasny:
Thanks for the three review passes — really useful. Addressed in `fb61cfc`:

**Blockers**
1. **Dilation (GPT Pro 5.5 Blocker 1) — fixed; you're right.** With v1's `ψₙ(x)=(n!√π)^{-1/2}Hₙ(x√2)e^{-x²/2}`, applying `weightL2Isometry` (×`√(gaussianPDF 0 1)`) to the `N(0,1)` basis `Hₙ/√(n!)` gives `2^{-1/4}·ψₙ(x/√2)` — the *dilated* Hermite-function basis, not `hermiteHilbertBasis`. A3′ and B2 now state the relation as **isometry + the `u=x√2` dilation**, with a target `weightL2Isometry_gaussianHermiteHilbertBasis_apply` pinning the image. Kept the standard `N(0,1)` normalization and carry the dilation explicitly.
2. **`barePolyLp_ortho_eq_bot` ungrounded (GPT 5.5 #1 / GPT Pro Blocker 2) — fixed.** It no longer manufactures completeness from `hdeg` alone; added the load-bearing moment-determinacy hypothesis `hexp` (the weighted measure `w·μ` has every exponential moment finite — true for Gaussian decay, automatic for compact support), which is exactly B1's condition. Degree growth ⇏ completeness is now explicit in the docstring.
3. **`Targets.lean` not wired in (GPT 5.5 #3 / Opus 4.8 #1) — fixed.** Added `import TauCetiRoadmap.OrthogonalL2Bases.Targets` to the root, and confirmed the file elaborates on the **pin (v4.31.0-rc1)**, not just v4.30 — `lake build TauCetiRoadmap` green, sorry-warnings only.

**Design**
- **`weightL2Isometry` generalized off ℝ (GPT Pro)** to an arbitrary measurable space; only the polynomial bridge keeps `Measure ℝ`.
- **Change-of-measure obligation (GPT Pro)** — added `hwm : AEMeasurable w μ` to `orthonormal_barePolyLp` / `hilbertBasisOfWeightedMeasure` / `barePolyLp_ortho_eq_bot`.
- **Isometry anti-vacuity (GPT 5.5 #2)** — added `weightL2Isometry_apply` (a.e. `= √w·f`).
- **Part D (Opus 4.8 #3)** — reframed from "future" to a grounded later milestone (consumes B3, which lands here).
- **Generality bar (Opus 4.8 #5)** — now pins the `Measure ℝ` domain (and notes the isometry is *not* specialized).
- **Upstream candidates (Opus 4.8 #4)** — `weightL2Isometry` / `HilbertBasis.mapₗᵢ` flagged as Mathlib-worthy.

**Cleanup** — PR body now links #46 (and notes `Targets.lean` seeds the representative core *including* B3 signatures, vs #46 leaving B3 unclaimed); fixed the stale "both routes" wording.

The `enable` check failure is still the fork-PR app-token infra issue, not content; required `build` is green.

mrdouglasny:
Ran our own Codex (read-only, lean-lsp) pass over the updated branch as a check — it confirmed the dilation correction and the `weightL2Isometry` generalization are right, and caught one more real issue the earlier passes missed, now fixed in `7c6afd7`:

- **`barePolyLp_ortho_eq_bot` was still false:** `natDegree 0 = 0`, so `hdeg` was satisfiable with `p 0 = 0` (counterexample `μ=δ₀, w=1, p 0=0, pₙ=Xⁿ` → a basis vector vanishes). `hdeg` now uses `degree` (`(p n).degree = (n : WithBot ℕ)`), forcing `p n ≠ 0`.
- **Grounded the general-`μ` completeness:** added a measure-level determinacy target `ae_eq_zero_of_forall_moment_eq_zero_of_finite_expMoments` (B1 was `volume`-only) and pointed `barePolyLp_ortho_eq_bot` at it.
- Added the promised bridge `coe_*` (`coe_hilbertBasisOfWeightedMeasure`, `coe_hilbertBasisOfOrthogonalSystem`); softened the `weightL2Isometry_gaussianHermiteHilbertBasis_apply` promise (it lands with the A2 `ψₙ` API); fixed Part D's coordinatewise-dilation wording.

Root `lake build TauCetiRoadmap` green on the pin (v4.31.0-rc1), sorry-warnings only.

cameronfreer (dismissed):
Thanks, that all looks good to me now. I've approved it (but there's a bit of cleanup needed before merge). Below are some comments collated from follow-up reviews by GPT Pro, Claude Code and Codex.

----


- Merge-readiness — PR is DIRTY/not mergeable → rebase on main.
- CI on the real head — no checks on 7c6afd7; rerun the build (matters now that Targets.lean is root-imported, so any elaboration slip hits main).
- PR body — stale/garbled: link #46, drop the "direct .map image" wording (README now says dilated), and remove the fragment that ran into the Claude footer.
- #46 scope — PR adds B1/B3 targets that #46 marks "not claiming" → state the "target signatures here, narrower implementation claim" split.
- Tiny no-leaps — …_finite_expMoments needs [IsFiniteMeasure ν] but barePolyLp_ortho_eq_bot only passes hexp; name a bridge lemma or add the hypothesis explicitly.
- Optional — add weightL2Isometry_symm_apply (g ↦ g/√w) to close the both-normalizations loop.

mrdouglasny:
All the pre-merge cleanup is in (head `7b17887`):

- **Rebased on `main`** — was DIRTY, now MERGEABLE (resolved the `TauCetiRoadmap.lean` import-aggregator conflict by keeping both `ConformalMapping.Targets` and `OrthogonalL2Bases.Targets`).
- **CI on the real head** — reran on `7b17887`; root `lake build TauCetiRoadmap` is green locally on the pin (v4.31.0-rc1), sorry-warnings only (Targets is root-imported).
- **PR body** — rewritten: links #46, drops the stale "direct `.map` image" wording (README says *dilated*), and removes the garbled footer fragment.
- **#46 scope** — added the split: `Targets.lean` seeds representative *target signatures* (incl. the B1/B3 ones), while the #46 *implementation* claim stays narrower (Part 0 / A1 / B2 / A3′).
- **no-leaps** — dropped `[IsFiniteMeasure ν]` from `ae_eq_zero_of_forall_moment_eq_zero_of_finite_expMoments`; finiteness is the `a=0` case of `hexp`, derived inside, so `barePolyLp_ortho_eq_bot` (which has only `hexp`) applies it with no missing instance.
- **Optional** — added `weightL2Isometry_symm_apply` (`g ↦ (√w)⁻¹·g`) to close the both-normalizations loop.

The push dismissed the prior approval — could I get a re-approve once the build is green? Thanks for the thorough reviews.

cameronfreer (approved):

PR #49, opened 2026-07-01
roadmap: Hodge structures (pure, mixed, and polarized)

mrdouglasny:
Adds a roadmap entry **`TauCetiRoadmap/HodgeStructures/`** (`README.md` + `Suggested.lean`) for the
weight-general **linear-algebraic core of Hodge theory** — pure, mixed and polarized Hodge structures
and their period-domain points. Mathlib has nothing on this: no Hodge structures, polarizations,
mixed structures or period domains. Suggested library home: `TauCeti/Geometry/Hodge/`.

Scope is deliberately the **algebraic layer only**. The two things that sit on top of it are tracked
separately: variations of Hodge structure at
[#167](https://github.com/TauCetiProject/TauCetiRoadmap/issues/167), and the bicomplex → abstract
Hodge decomposition bridge (the route to the classical Kähler theorem) at
[#173](https://github.com/TauCetiProject/TauCetiRoadmap/issues/173).

## Design

- **The integral lattice is the primary datum.** A weight-`n` structure is carried on a finitely
  generated free `ℤ`-module `V_ℤ`.
- **The complexification is a base-change interface, not a fixed construction.** `V_ℂ` is any
  `ℂ`-space with `ι_ℂ : V →ₗ[ℤ] V_ℂ` satisfying `IsBaseChange ℂ ι_ℂ`, so a geometric `V_ℂ = Hⁿ(X;ℂ)`
  — not literally `ℂ ⊗ Hⁿ(X;ℤ)` — plugs in with no transport isomorphism. The canonical tensor model
  is the checked instance. `IsBaseChange` is a `Prop` about the structure map; `cancelBaseChange` is
  the data that discharges it for that model, and the README now says which is which.
- **One pure-Hodge object, parametric in the conjugation.** The weight-`n` axioms mention only an
  ambient `ℂ`-space and a conjugation on it, so the object is stated once over a `Conjugation W` (a
  conjugate-linear involution, involutivity part of the datum). The integral case is an abbreviation
  rather than a copy, and the rational case on `gr^W_k` is the same object at `gradedConj` — so
  `graded_pure` reads "`gr^W_k` carries a Hodge structure of weight `k`" instead of restating the
  axioms, and the pure category has a home.
- **Conjugation is defined, not assumed.** `latticeConj` is the conjugate-linear map fixing the
  integral points, determined by the base-change universal property; on the canonical instance it is
  `TensorProduct.map (starRingEnd ℂ) id`.
- **Polarization is one integral form plus a `Prop` mixin.** `IsPolarization hs Qint` carries the
  Hodge–Riemann relations; the complex form is derived as `Qint.baseChange ℂ`, so the integral↔complex
  link is Mathlib's `baseChange_tmul` rather than an imposed axiom.

## Milestones

| Layer | Milestone |
|---|---|
| **L0** | Hodge decomposition — `DirectSum.IsInternal` of the `(p,q)`-pieces |
| **L1** | semisimplicity of **polarizable** `ℚ`-Hodge structures, with ordinary Hodge morphisms |
| **L2** | Deligne strictness of MHS morphisms, through the canonical `I^{p,q}` bigrading |
| **L3** | the Hodge numbers partition `dim_ℂ V_ℂ` |

Each is a `sorry`-goal in `Suggested.lean`. Alongside them the entry targets the API the theorems are
useless without: the **Weil operator** `C` (with `C² = (−1)^n`), which is what carries
`h(u,v) = Q(Cu, v̄)` from the pieces to all of `V_ℂ`; the **Deligne splitting** as a definition with
recovery of `F` and `W`, functoriality, and the conjugation relation modulo lower bidegree, rather
than as propositional existence; and a **worked instance**, the Tate structure `ℤ(m)` with its
polarization and the Tate twist's weight and filtration shifts pinned.

## Review

Thanks to Kim Morrison and Simone Chiarello, whose reviews changed this substantially: an
effectivity error in the weight-one claims, the scope split that produced #167 and #173, the
category question at L1, and the round of blocking items that produced the conjugation-parametric
object, the Weil operator and Deligne-splitting targets, and the worked instance.

Roadmap: HodgeStructures

*(Disclosure: this roadmap and this description are written by Claude under my direction and
editing.)*


TauCetiRoadmap/VHS/README.md
```diff
@@ -0,0 +1,262 @@
...
+  - Conjugation: `starRingEnd ℂ` (complex conjugation as the semilinearity ring hom); there is **no**
+    packaged real/integral complexification-with-conjugation, so `latticeConj` is built here.
+  - For variations (L4, downstream): `CategoryTheory.FundamentalGroupoid`, `Module ℤ` (local systems),
+    and Mathlib's complex-manifold / connection API (Griffiths transversality).
+  - Rigidity engine (L5): `Module.End`, `Module.End.HasEigenvalue` / `Module.End.exists_eigenvalue`
+    over the algebraically closed `ℂ`.
+- **Other proof assistants.** Hodge structures, polarizations, and variations of Hodge structure
+  appear **essentially unformalized anywhere** (Isabelle/HOL, Coq/Rocq, Lean). Adjacent pieces exist
+  (abelian varieties and the upper half-space in various places), but the abstract Hodge-theoretic
+  superstructure is new foundational material, not a port.
```
kim-em: :robot: Claude says:

This says Hodge structures are "essentially unformalized anywhere ... Lean" (and line 246, "credit none — original"). That's just been overtaken: [Booker Smith announced pure Hodge structures in Lean 4](https://leanprover.zulipchat.com/#narrow/channel/583339-AI-authored-projects/topic/Pure.20Hodge.20structures.20in.20Lean.204/near/608339235) (github.com/thebookersmith/pure-hodge-structures-lean4), which formalizes exactly the L0 layer — the decomposition ↔ opposed-filtration equivalence, both directions, axiom-clean. Worth citing as concurrent prior art rather than claiming originality, and it's a useful cross-check for our L0 signature (note they take the ℚ-space as primary where we take the ℤ-lattice).

kim-em (commented):
:robot: Claude says:

This is in good shape and the definitions look faithful to Peters–Steenbrink Ch. 2. My main ask before we treat this as settled is that the roadmap explicitly track the Hodge-theory work now moving in Mathlib, so an implementor builds *toward* it rather than around it, and so we know in advance what we'll refactor once it lands. Three specific threads:

- **Opposed filtrations and induced filtrations on graded pieces are being added to Mathlib right now.** [feat(CategoryTheory/Abelian): add Filtration API (leanprover-community/mathlib4#33954)](https://github.com/leanprover-community/mathlib4/pull/33954) formalizes Deligne, *Théorie de Hodge II* §1.1, and is explicitly PR 1/4, with "Opposed filtrations (§1.2.1–1.2.3)" and "Induced filtrations on graded pieces (§1.2.1)" named as the follow-ups. That is exactly what L0's `opposed` and the whole L2 `complexGradedF`/`gradedF`/`gradedComplexEquiv` apparatus build by hand. The abstraction levels differ (that PR is in an abstract abelian category; here we're concrete over `Submodule ℂ V_ℂ`), so I'm not asking to consume it verbatim today — but the roadmap should cite it, say that L0/L2 should specialize from / be refactored onto it once it merges, and align naming with Deligne §1.2.1. It grew out of [Joël Riou's message on n-opposed filtrations](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Proposal.3A.20Complexifications.20with.20a.20view.20towards.20Hodge.20theory/near/565511365).

- **The complexification model is a live design question, and the direction on that thread went the other way from what we do here.** We commit to `V_ℂ := ℂ ⊗[ℤ] V` as the primary object and build the `cancelBaseChange` tower machinery, and the roadmap already admits the nested `ℂ ⊗_ℚ (ℚ ⊗_ℤ V)` base changes are "ergonomically heavy." On [Proposal: Complexifications with a view towards Hodge theory](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Proposal.3A.20Complexifications.20with.20a.20view.20towards.20Hodge.20theory), Johan, Andrew Yang and Kevin Buzzard converged on modeling this with a K-structure predicate — an ambient space with `IsBaseChange L V₀.subtype` — precisely to avoid that pain ([Andrew Yang on `IsBaseChange`](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Proposal.3A.20Complexifications.20with.20a.20view.20towards.20Hodge.20theory/near/565460949)). Could you add a note weighing `IsBaseChange` against the concrete tensor and either adopt it or record why not? I don't want us to lock in a base-change convention Mathlib is deciding against.

- **There are two routes to the `(p,q)` decomposition and we should say which we consume.** We take the Deligne opposed-filtration route (right, and what Joël Riou recommended); [feat(LinearAlgebra): complex structures on real vector spaces (leanprover-community/mathlib4#40975)](https://github.com/leanprover-community/mathlib4/pull/40975) is building the `J`, `J² = -1` complex-structure route that gives the same decomposition as `±i`-eigenspaces. Since the weight-1 / abelian-variety instances naturally carry a `J`, one line relating `piece` to that eigenspace picture would tell an implementor how instances from #40975 plug in.

TauCetiRoadmap/VHS/Suggested.lean
```diff
@@ -0,0 +1,744 @@
...
+schematically as complex subbundles, Griffiths transversality, and polarized monodromy. The analytic
+notions are placeholders until the roadmap imports the needed complex-geometry infrastructure. -/
+structure VariationOfHodgeStructure (B : Type*) (V : Type*) [AddCommGroup V] [Module ℤ V]
+    [Module.Free ℤ V] [Module.Finite ℤ V] (n : ℤ) (Γ : Type*) [Group Γ] where
+  fiber : HodgeStructure V n
+  polarization : Polarization fiber
+  monodromy : PolarizedMonodromyRepresentation fiber polarization Γ
+  hodgeBundle : ℤ → B → Submodule ℂ (Complexification V)
+  holomorphic : ∀ _ : ℤ, Prop
+  griffiths_transversality : Prop
```
kim-em: :robot: Claude says:

Small thing on the schematic fields: `holomorphic : ∀ _ : ℤ, Prop` and `griffiths_transversality : Prop` are fields whose *type* is `Prop`, so any instance satisfies them with `True` — the whole VHS datum is then silently contentless. I think it's more honest to write these as explicit `(sorry : Prop)` holes (I checked: `(sorry : Prop)` elaborates fine, bare `sorry` doesn't — it leaves a universe metavariable). Then an instance genuinely can't be built without discharging the analytic condition, which is the true state until the complex-geometry API exists. If you do this, the closing NOTE's "every definition is complete" needs a small tweak to say these two L4 fields carry `sorry`.

TauCetiRoadmap/VHS/Suggested.lean
```diff
@@ -0,0 +1,744 @@
...
+rigidity and Deligne's theorem of the fixed part / semisimplicity -- but those full theorems need
+genuine *polarizable VHS* hypotheses (a real VHS, not just a form-preserving representation); this
+milestone is the plain finite-dimensional Schur lemma that they invoke. -/
+example {n : ℤ} (hs : HodgeStructure V n) (pol : Polarization hs) {Γ : Type*} [Group Γ]
+    (M : PolarizedMonodromyRepresentation hs pol Γ)
+    (hirr : ∀ W : Submodule ℂ (Complexification V),
+      (∀ g, W.map ((M.complexMonodromy g).toLinearMap) = W) → W = ⊥ ∨ W = ⊤)
+    (T : Complexification V →ₗ[ℂ] Complexification V)
+    (hT : ∀ g v, T (M.complexMonodromy g v) = M.complexMonodromy g (T v)) :
+    ∃ c : ℂ, ∀ v, T v = c • v := sorry
```
kim-em: :robot: Claude says:

One caveat for whoever discharges this: the standard argument (over algebraically closed `ℂ`, a commuting `T` on a finite-dimensional irreducible rep has an eigenvalue via `Module.End.exists_eigenvalue`, and `ker (T - c • 1)` is a nonzero invariant subspace hence `⊤`) assumes `Complexification V ≠ 0`. When `V = 0` the space is `0`, `hirr` holds vacuously, and `exists_eigenvalue` doesn't apply — the conclusion is still trivially true for any `c`, but the proof has to dispatch the `Subsingleton (Complexification V)` case first. Worth a word in the L5 discharge sketch so it isn't a surprise.

TauCetiRoadmap/VHS/Suggested.lean
```diff
@@ -0,0 +1,744 @@
...
+sibling relations) is in `README.md`. **Mathlib has no Hodge structures**, so the chief
+deliverable of this entry is getting the *definitions* right (the `JacobianChallenge`
+philosophy); below are proposed core definitions and a milestone `sorry` for each layer with a
+self-contained target (L0, L1, L2, L3, L5; L4 is a schematic structure seed — see `README.md`). The
+deep geometric/analytic engines (Kähler Hodge decomposition, Gauss-Manin of general families,
+Schmid's asymptotics) are **out of scope** -- this is the weight-general *structural* theory;
+instances come from elsewhere (the weight-1 / curve case is the worked model).
+
+NOTE: elaborates green against `TauCetiRoadmap`'s pinned Mathlib (leanprover/lean4:v4.31.0-rc1); the
+milestone `example`s carry `sorry`, every definition is complete.
```
kim-em: :robot: Claude says:

Could you add the standard header the other roadmaps carry — "**This file is not the roadmap and is not exhaustive.** The definitive document is `README.md`. ..."? After the [Targets→Suggested rename (#55)](https://github.com/TauCetiProject/TauCetiRoadmap/pull/55) this file is the only active one missing it, and that note is the whole point of the rename.

mrdouglasny:
Thanks — all three addressed (README, commits de4284d + c4642d0):

1. **Filtration API (mathlib4#33954).** Added a *Prior art* block flagging it as incoming work to build toward, with pointers from L0 (`opposed`) and L2 (`gradedF`/`gradedComplexEquiv`): both are exactly its §1.2.1 "opposed filtrations" / "induced filtrations on graded pieces" follow-ups, to be specialized from it once merged, aligning naming with Deligne §1.2.1.

2. **IsBaseChange vs concrete tensor — adopted.** The complex/rational spaces are now specified by `IsBaseChange` predicates on structure maps out of `V_ℤ`, with `ℂ ⊗[ℤ] V` as the canonical instance (`TensorProduct.isBaseChange`). This (i) lets geometric instances (`V_ℂ = Hⁿ(X;ℂ)`, not literally `ℂ ⊗ Hⁿ(X;ℤ)`) satisfy the interface with no transport iso — the decisive win for the L4 variations — and (ii) composes the `ℤ→ℚ→ℂ` tower by `IsBaseChange.comp`, retiring the `cancelBaseChange` threading I'd flagged as ergonomically heavy. `V_ℤ` stays primary; `IsBaseChange` is a predicate *about* the structure map out of it. Pinned in *Conventions*. This is README-level adoption — `Suggested.lean` still encodes the canonical-instance (concrete tensor) definitions; **interface-parametrizing `Suggested.lean` is a scoped follow-up** (called out in the closing NOTE) so the green/axiom-clean milestone file gets refactored carefully rather than rushed.

3. **Complex structures (mathlib4#40975).** Added an L0 *instance bridge* note: `piece` agrees with the `±i`-eigenspace decomposition of a `J` (`J²=−1`), so weight-1 / abelian-variety instances carrying a `J` plug in; the roadmap consumes the Deligne opposed-filtration route (per Riou) and #40975 supplies the `J`.

kim-em:
> Thanks — all three addressed (README, commits [de4284d](https://github.com/TauCetiProject/TauCetiRoadmap/commit/de4284d1822bcb84e1a7e8f2279fcd58f802cfc9) + [c4642d0](https://github.com/TauCetiProject/TauCetiRoadmap/commit/c4642d0322365f8e92fd4a32d7433b604690e741)):

Note that there are inline review comments as well. :-)

mrdouglasny:
Thanks for the nudge — all four inline comments are now addressed (commits c69ac13 + 8aa213a), on top of the three top-level asks:

- **L723 (vacuity)** — the two L4 analytic fields are now explicit `(sorry : Prop)` holes (`holomorphic : ∀ _ : ℤ, (sorry : Prop)`, `griffiths_transversality : (sorry : Prop)`), so an instance genuinely can't be built without discharging the analytic condition rather than being silently satisfiable by `True`. Both completeness notes (Suggested.lean header + README NOTE) now say these two fields carry `sorry` and are the only definitional `sorry`s; the "axiom-clean" claim is scoped to everything else.
- **L76 / References** — dropped the "essentially unformalized … Lean" / "credit none — original" framing; now cites your pointer to Booker Smith's concurrent `pure-hodge-structures-lean4` (L0: the `(p,q)`-decomposition ↔ opposed-filtration equivalence, both directions, axiom-clean, ℚ-primary vs our ℤ-lattice) as concurrent prior art and a cross-check, keeping the originality claim for the L1–L5 superstructure.
- **L16** — added the standard "this file is not the roadmap / README is definitive" header.
- **L742** — added the L5-Schur discharge caveat (the eigenvalue argument assumes `Complexification V ≠ 0`; the `V = 0` subsingleton case must be dispatched first).

Elaborates green (the VHS struct now legitimately `uses sorry` like the milestone examples).

TauCetiRoadmap/VHS/README.md
mrdouglasny: Addressed in c4642d0/8aa213a: dropped the "essentially unformalized … Lean" / "credit none — original" framing and now cite Booker Smith's concurrent `pure-hodge-structures-lean4` (the L0 decomposition↔opposed-filtration equivalence, ℚ-primary vs our ℤ-lattice) as concurrent prior art in *Prior art* + *References*, keeping the originality claim for the L1–L5 superstructure.

TauCetiRoadmap/VHS/Suggested.lean
mrdouglasny: Addressed in c69ac13: the two L4 analytic fields are now explicit `(sorry : Prop)` holes (`holomorphic : ∀ _ : ℤ, (sorry : Prop)`, `griffiths_transversality : (sorry : Prop)`), so an instance genuinely can't be built without discharging the analytic condition rather than being silently satisfiable by `True`. Both completeness notes (Suggested.lean header + README NOTE) now state these two fields carry `sorry` and scope the axiom-clean claim to everything else.

TauCetiRoadmap/VHS/Suggested.lean
mrdouglasny: Addressed in 8aa213a: added the discharge caveat to the L5-Schur milestone docstring — the eigenvalue argument assumes `Complexification V ≠ 0`, and the `V = 0` (`Subsingleton`) case must be dispatched first (conclusion still trivially true there for any `c`).

TauCetiRoadmap/VHS/Suggested.lean
mrdouglasny: Addressed in 8aa213a: added the standard "**This file is not the roadmap and is not exhaustive.** The definitive document is `README.md` …" header the other roadmaps' `Suggested.lean` carry.

kim-em:
:robot: Codex says:

Thanks—the earlier review comments are addressed, but two design problems remain.

1. The README pins `IsBaseChange` as the defining interface and says definitions and milestones are stated against it. `Suggested.lean` instead hardcodes `Complexification V := ℂ ⊗[ℤ] V`, and the closing note postpones interface-parametrization as a future refactor. Please make the proposed definitions match the pinned design now, or choose the concrete tensor convention consistently; the roadmap should not prescribe one interface while prototyping another.

2. `VariationOfHodgeStructure` still represents `holomorphic` and `griffiths_transversality` using `(sorry : Prop)`. Those fields have no mathematical content: they neither state holomorphy nor transversality. The roadmap-writing rules say to omit a condition that cannot yet be stated, rather than install a `Prop` placeholder. Please remove the schematic full-VHS structure for now and retain the honest `PolarizedMonodromyRepresentation`, unless the analytic fields can be given genuine types using the relevant complex-geometric API.

Please also cleanly classify the period-domain openness, full VHS, fixed-part, and rigidity material: each should either be an in-scope milestone with its dependencies or explicitly downstream/out of scope, rather than a “deeper target” that is simultaneously not part of the roadmap.

mrdouglasny:
Thanks Kim — all three threads are now addressed; the head is `8b9251c`.

**1 · Opposed / induced filtrations (mathlib4#33954, Deligne §1.2.1).** The *Prior art* section now cites #33954, flags it as PR 1/4 with the "opposed filtrations §1.2.1–1.2.3" and "induced filtrations on graded pieces §1.2.1" follow-ups, and commits the roadmap to **specialize L0's `opposed` and L2's `gradedF`/`gradedComplexEquiv` onto it once merged**, aligning naming with Deligne §1.2.1. The per-layer *Discharge* notes for L0 and L2 repeat the specialize-and-align instruction so an implementor building by hand today knows what it will be refactored onto.

**2 · Complexification model — `IsBaseChange` adopted, not just noted.** I took the Zulip direction (Commelin / Yang / Buzzard): rather than only recording the trade-off, the *definitions in `Suggested.lean` are now parametric over an abstract `V_ℂ` with `hℂ : IsBaseChange ℂ ιℂ`* (and `V_ℚ` with `IsBaseChange ℚ ιℚ` where the ℤ→ℚ→ℂ tower is used), with `V_ℤ` the primary datum. The concrete tensor `ℂ ⊗[ℤ] V` is kept only as the **canonical instance**, witnessed by `complexificationMap_isBaseChange := TensorProduct.isBaseChange`. Every construction is transported from that instance through `hℂ.equiv`: `latticeConj` (proven to fix `ιℂ(V)` and be involutive), the polarization form (`BilinForm.congr`), the rational→complex tower (`cancelBaseChange`), and the weight-graded apparatus — including a proof that complexification commutes with the rational weight-graded quotient (`gradedComplexEquiv`). So we no longer lock in a base-change convention Mathlib is deciding against; the concrete `cancelBaseChange` tower is now an implementation detail behind the interface. The *Conventions* note weighs the two and records why `IsBaseChange` wins. Builds green against the pinned Mathlib; the milestone `sorry`s (L0/L1/L2/L3/L5) are the only ones and every definition is complete.

**3 · Two routes to `(p,q)` (mathlib4#40975).** L0 now carries an *instance-bridge* line: a weight-1 / abelian-variety HS naturally carries a complex structure `J` (`J² = −1`) whose `±i`-eigenspaces are exactly the `(1,0)`/`(0,1)` pieces, so `piece` agrees with the `J`-eigenspace decomposition of #40975 — telling an implementor how instances from that PR plug into the Deligne opposed-filtration route we consume (per Riou's recommendation).

Happy to adjust naming further once #33954 lands and its §1.2.1 API is visible.

kim-em:
:robot: Claude and Codex request:

**1. Blocking: the weight-one claims are false at the stated generality.**

Nothing in `HodgeStructure` restricts `p` to `{0,1}`, so a general weight-one Hodge structure may carry `H^{2,-1}` and `H^{-1,2}`. On `H^{p,1-p}` the Weil operator acts by `i^{2p-1}`, which is `±i` for every `p`, so `H^{2,-1}` has `p-q = 3` and `i³ = -i` and lands in the same eigenspace as `H^{0,1}`. The `±i` eigenspaces therefore aggregate all odd `p-q` classes and are not exactly `H^{1,0}` and `H^{0,1}`.

Three assertions depend on this and are false without an effectivity hypothesis:

- *Prior art*: "Polarized weight-1 Hodge structures are abelian varieties / complex tori `ℂ^g/Λ`".
- *Generality bar*: "weight 1 recovers Siegel (`Sp(2g,ℝ)/U(g)`)".
- The L0 weight-one instance note: that a `J` with `J² = -1` has `±i` eigenspaces equal to the `(1,0)` and `(0,1)` pieces, agreeing with `piece`.

Please introduce an effective or level-one condition, requiring the support of `h` to lie in `[0, n]`, and restrict the complex-torus, abelian-variety, Siegel and `J`-eigenspace claims to that subtype.

**2. Blocking: L5 restates a theorem Mathlib already has.**

`IsSimpleModule.algebraMap_end_bijective_of_isAlgClosed` in `Mathlib/RepresentationTheory/AlgebraRepresentation/Basic.lean` already gives exactly the L5 conclusion: over an algebraically closed field, every endomorphism of a simple module is scalar. `Mathlib/RepresentationTheory/Irreducible.lean` supplies the irreducibility API, and `CategoryTheory.Preadditive.Schur` and `FDRep.finrank_hom_simple_simple` are also present. The L5 discharge note instead plans to reprove this from `Module.End.exists_eigenvalue` and an invariant-subspace argument, which cuts against this repository's instruction to consume Mathlib rather than rebuild it.

`hs` and `pol` are also decorative in both L4 and L5. In `PolarizedMonodromyRepresentation`, `hs` exists only to type `pol` and `pol` is used only via `pol.Qint`, so the structure is "a group preserving an integral bilinear form", and the L5 statement never mentions the filtration at all. As it stands L5 is Mathlib's Schur lemma with Hodge-theoretic parameters that play no role.

Please bundle `complexMonodromy` as `Representation ℂ Γ Vℂ` and make the milestone the reduction to the existing theorem. The whole chain is already there:

```
Representation.IsIrreducible ρ
  -- Representation.irreducible_iff_isSimpleModule_asModule
  → IsSimpleModule ℂ[Γ] ρ.asModule
  -- IsSimpleModule.algebraMap_end_bijective_of_isAlgClosed
  → Function.Bijective (algebraMap ℂ (Module.End ℂ[Γ] ρ.asModule))
```

`Representation.IntertwiningMap` replaces the bespoke commuting-endomorphism hypothesis and `Representation.IsIrreducible` replaces the bespoke `hirr`. What remains genuinely to prove is finite-dimensionality of `Vℂ` from `hℂ` together with `Module.Finite ℤ V`; make that the L5 milestone. If neither is wanted, remove L5 rather than leaving it as a layer with no milestone.

**3. Blocking: the roadmap is largely written in terms of what it will not do.**

The top-level README says everything on a roadmap is work we want, and every item lives in some milestone, because a contributor may otherwise misread "later" as "never". This roadmap does the opposite in eight places: a `## Downstream` section, `*Out of scope (downstream, not a milestone)*` at L3, `*Out of scope (downstream, not seeded)*` at L4, `*Out of scope (downstream, not milestones)*` at L5, and the opening declaration that the Hodge decomposition, Gauss-Manin and Schmid's asymptotics are "out of scope here".

The effect is that the title promises "variations (with the period map and monodromy) at their natural generality with full basic API" while the full VHS datum is never stated, L4 has no milestone at all, and L3 reduces to a dimension count.

Please pick the scope and commit to it, so that everything named is a milestone and nothing is named that is not:

- Rename to something like "Hodge structures: pure, mixed, and polarized", and delete the out-of-scope and downstream prose rather than relocating it. Variations, period maps and rigidity then simply do not appear.
- Or keep the name and give the missing material real milestones: `Vℝ` with realification and the Weil operator, the Deligne torus and the real Hodge structure correspondence, local systems, connections, Griffiths transversality and the period map.

We suggest the first. The second commits the roadmap to analytic API Mathlib does not have, and the material delivered so far is coherent and good under the narrower name.

**4. Smaller items.**

- `HodgeType` requires only finite support. Hodge symmetry forces `h p = h (n - p)` for any realizable type, since `conj (piece p) = piece (n - p)` and conjugation preserves complex dimension. Unrealizable types are currently admitted, and the L3 milestone holds vacuously on them. Add the symmetry condition and a total-rank condition.
- `PeriodDomain` is a point of the classifying set, not the homogeneous complex manifold its docstring describes. Either rename it to `PeriodDomain.Point` or `IsPeriodDomainPoint` and drop the manifold language from the docstring, or give openness, the `G_ℝ`-action and the homogeneous presentation their own milestones.
- `PeriodDomain` stores a fresh `Polarization` and then constrains it by `pol_form : pol.Qint = Qint`. That is duplicate data; `IsPolarization hs Qint` would avoid it.
- *Prior art* writes `Module.IsBaseChange`. The declaration is root `IsBaseChange`, in `Mathlib/RingTheory/IsTensorProduct.lean`; `Module.IsBaseChange` does not resolve.
- L4 proposes modelling the local system as `CategoryTheory.FundamentalGroupoid B ⥤ Module ℤ`. `Module ℤ` is a typeclass, not a category, so this signature does not typecheck. `ModuleCat ℤ` is the category, and finite-free fibers need stating separately.
- The lattice cannot literally be `Hⁿ(X;ℤ)` for a smooth projective variety, since integral cohomology can have torsion while the roadmap requires `Module.Free ℤ V`. Say `Hⁿ(X;ℤ)/torsion`, or take the rational structure as primary with a chosen lattice alongside.
- The polarization convention `Q(y,x) = (-1)^n Q(x,y)` with `i^{p-q} Q(v, v̄) > 0` is internally consistent, but geometric cup-product instances normally carry an extra `(-1)^{n(n-1)/2}` together with a Lefschetz-power and primitivity convention. Pin this explicitly so instance authors do not pick the wrong sign.
- The roadmap is not registered in the root `README.md` list. The sync workflow will auto-append `16. [Variation of Hodge structure (general)]` after merge, but every other roadmap curates its own entry and position.

kim-em:
Actual @kim-em here: @mrdouglasny, could you help me in finding a human who actually knows this material to take a look?

mrdouglasny:
@kim-em — thanks for the detailed review, and for the "principles of good roadmaps" you posted since; we've used both. (For disclosure: the roadmap and this reply are written by Claude under my direction and editing.)

The main change is structural: we've **narrowed this roadmap to the pure/mixed/polarized Hodge-structure theory** and split the variations material out as a separate future roadmap, tracked at #167. That resolves **③** — everything named here is now a real milestone (L0–L3), and the variations / period-map / rigidity material is named *elsewhere* (a sibling roadmap that builds on the objects defined here), not as out-of-scope prose.

Point by point:

- **① effectivity (the real error):** added `HodgeStructure.IsEffective` (Hodge numbers in `[0,n]`) and restricted the abelian-variety / Siegel / `J`-eigenspace claims to it — a general weight-1 HS can carry `H^{2,−1}` etc., as you note.
- **② L5:** you're right it was Mathlib's Schur with decorative Hodge parameters; we removed L4/L5. The successor consumes `IsSimpleModule.algebraMap_end_bijective_of_isAlgClosed` via `Representation.irreducible_iff_isSimpleModule_asModule` (both confirmed present) rather than reproving it.
- **④:** `HodgeType` gains a weight + Hodge symmetry; `PeriodDomain.Point` (the manifold structure is the successor's); the `IsPolarization` `Prop`-mixin removes the duplicated form; root `IsBaseChange`; `ModuleCat ℤ`; lattice-torsion caveat; polarization sign pinned; registered in the root index. The directory is renamed `VHS/ → HodgeStructures/`.

We also went back over it for your newer principles: made it **timeless** (no "blocked / once-merged / not-yet" — the prerequisite/blocked content lives in the tracking issue #167 instead), and cut the walls of text (especially L2). `Suggested.lean` elaborates green (v4.31.0-rc1), milestones L0–L3, axiom-clean.

On your ask for a subject expert: I use Hodge theory / periods in my theoretical-physics work rather than working in the field directly, so — although we've checked the mathematics carefully — an independent look from a specialist would be valuable. **Joël Riou** (his Deligne n-opposed-filtration comment on the *Complexifications with a view towards Hodge theory* thread is exactly this L0) and **Simone Chiarello** (who opened that thread, and is following Peters–Steenbrink Ch. 2 — the same algebraic route taken here) would be the most on-point.

0xSimex:
Hi @mrdouglasny, Simone Chiarello here! First of all, I’m very flattered to have been asked to look at this. Small disclaimer: I’m not a Hodge theorist, and my knowledge of Hodge structures is rusty, to put it generously. Also, I haven’t been an active mathematician for a few years now, so take what I say with a grain of salt.

When I opened the original thread (and then let it sit for months — entirely my fault, as I got dragged into other projects; I’d be very happy to resume it now that it has gained some visibility), what I had in mind was less the algebra of Hodge structures for its own sake, and more building an algebraic foundation on which geometric results could eventually land, with the Hodge decomposition theorem for compact Kähler manifolds as the most natural first candidate.

I understand from the README that this PR deliberately covers the first, purely algebraic step of that picture, without geometric machinery. With that in mind, I wonder whether it would be useful to include, or at least clearly point toward, the algebraic bridge from bicomplexes to `HodgeStructure`.

Mathlib already has bicomplexes (`HomologicalComplex₂`) and total complexes, and now also has general spectral-sequence infrastructure. What seems to be missing on the Hodge-theoretic side is an abstract `∂∂̄-lemma` and the theorem saying that, for a suitable bicomplex with real structure, it yields a Hodge decomposition on total cohomology, with `E₁`-degeneration of the associated spectral sequence as a consequence.

That would make the `opposed` condition in L0 the output of a genuinely nontrivial cohomological theorem, rather than only part of the input data. No geometric machinery is needed for this step, and some bridge of this kind seems important if the definitions here are eventually meant to support geometric applications.

I don’t necessarily mean adding all of this to #49 — enough scope creep already :) — but I think it would make a very natural sibling/successor roadmap.

mrdouglasny:
*(Disclosure: co-written by MRD and Claude.)*

Simone — thank you, this is exactly the kind of review I was hoping for, and no need for the disclaimers; the suggestion is a good one.

You've read the scope right: #49 is deliberately the purely algebraic layer, and I agree the ∂∂̄ → Hodge-decomposition bridge shouldn't go in it — partly for the scope-creep reason you give, but also structurally: that bridge's output is a `HodgeStructure` (its conclusion is exactly the `opposed` / `(p,q)` condition L0 posits), so it's a consumer of this roadmap. Foundation and consumer are cleaner as separate roadmaps, with the bridge sitting on top.

And I think it's an attractive one to do next — because, unlike the variations successor (tracked at #167), which is blocked on complex-manifold / connection API Mathlib doesn't yet have, this one is statable today. As you note, Mathlib already has bicomplexes (`HomologicalBicomplex`), total complexes, and the abstract spectral-object / spectral-sequence machinery. What's missing is the Hodge-theoretic content itself: an abstract ∂∂̄-lemma, the theorem that a bicomplex with a real structure yields a Hodge decomposition on total cohomology, and E₁-degeneration of the (Frölicher) spectral sequence as a consequence. That's a genuine theorem (Voisin I §6 / Deligne–Griffiths–Morgan–Sullivan), and wiring the abstract spectral-object API to a concrete degeneration is the fiddly part — but it is all algebra, no geometry, and it would make L0's `opposed` an output of a real cohomological theorem rather than input data, exactly as you say.

This is squarely your original motivation — an algebraic foundation on which the geometric Hodge decomposition can land — and you clearly know the homological-algebra side well, so I'd be glad to work on it with you. I've opened a tracking issue for it — #173 (a sibling to #167) — with a first scope sketch, and I'll add a one-line pointer from this roadmap's successor section. Thanks again for taking the time.

0xSimex:
Thanks for opening #173: this is exactly what I had in mind.

As a further comment on my suggestion, once the ∂∂̄-lemma has been formalized at the appropriate algebraic level, the remaining route to the Hodge Decomposition Theorem for compact Kähler manifolds is conceptually clear: one constructs the Dolbeault bicomplex of differential forms, with its operators ∂ and ∂̄, and proves that it satisfies the ∂∂̄-lemma.

Of course, carrying this out requires a substantial amount of geometric and analytic machinery. There are also a few qualifications concerning the precise scope of the resulting development:

* Classically, for a compact Kähler manifold, the Kähler identities, together with elliptic Hodge theory, imply the ∂∂̄-lemma and hence the complex Hodge decomposition. Formalizing this implication was the original goal I had in mind, and it would already be a significant achievement.
* The corresponding real Hodge structure should then follow with comparatively little additional machinery, using complex conjugation on differential forms and the comparison between real and complex de Rham cohomology.
* The abstract rational and integral Hodge structures are already included in the scope of the present roadmap. What would still be required is their geometric realization on the cohomology of a compact Kähler manifold. This requires additional topological, algebraic, and comparison machinery, including singular cohomology with rational and integral coefficients, the torsion-free integral lattice, extension of scalars, and the de Rham comparison isomorphism. In the projective setting, after algebraic cycles and the cycle class map have also been introduced, this rational structure makes it possible to formulate the Hodge Conjecture: the rational cycle class map 
```math
\text{cl}_{\mathbb Q}\colon CH^p(X)\otimes\mathbb Q
\longrightarrow H^{2p}(X,\mathbb Q)\cap H^{p,p}(X)
```
is conjectured to be surjective.
* Similarly, the abstract algebraic notion of a polarized Hodge structure is already part of the present roadmap. Proving the corresponding geometric polarization theorem requires a genuinely further layer of theory, including the Lefschetz operator, Hard Lefschetz, primitive decomposition, and the Hodge–Riemann bilinear relations. For compact Kähler manifolds this naturally yields a real polarization, while a rational or integral polarization requires an appropriate rational or integral Kähler class, as in the projective case. This geometric realization lies beyond the scope of #173.

Nonetheless, this PR, together with #173, provides a solid algebraic foundation for the future development of Hodge theory in Mathlib. Since both this PR and the related issue are intentionally limited to these abstract algebraic foundations, I think this PR is ready for approval in its current form.

kim-em:
:robot: Claude and Codex request:

The mathlib4#33954 correction in cf21339/cd23d11 is right, and the L0-L3 milestone statements are all
true as stated. The items below are in the surrounding prose and in missing API.

**1. Blocking: mathlib4#40975 is listed as a Mathlib prerequisite, and the claim made about it is
wrong.**

README:66-71 lists *Complex structures on real vector spaces*
([mathlib4#40975](https://github.com/leanprover-community/mathlib4/pull/40975)) under "Mathlib, the
prerequisites this entry consumes", and README:212 says "`#40975` instances supply the `J`". It is an
open PR, so it is not something this entry can consume. The README's rule is that an open Mathlib PR
is never a dependency: build what you need here, shaped to match. #33954 just got exactly this
treatment; its sibling did not.

Separately, README:69-70 says the opposed-filtration route and the `J`-eigenspace route "yield the
**same** decomposition", with no hypothesis. A `J` has only the two eigenvalues `±i`, so it recovers
the decomposition exactly when the structure has two conjugate types with `p−q` odd; effective weight
one is the intended case, and Tate twists of it also qualify. Outside that, the `±i` eigenspaces
group types by `p−q` mod 4. Note also that the Weil operator is a complex structure only in odd
weight, since `C² = (−1)^n`. L0:209 carries a qualification of this kind and Prior art does not.

Either state the hypothesis and make `V_ℝ`, the construction of `J`, and the eigenspace comparison
targets here, or drop the claimed dependency and keep #40975 as prior art only.

**2. Blocking: two places where the README states a design the file contradicts.**

`Suggested.lean` is explicitly non-exhaustive, so absence from it proves nothing. These two are
contradictions rather than absences.

- README:41, 121, 143-144 and 170 say the `ℤ→ℚ→ℂ` tower composes by `IsBaseChange.comp` "rather than
  a hand-threaded `cancelBaseChange`", and 144 says "no hand-threaded `cancelBaseChange`".
  `IsBaseChange.comp` appears zero times in the file; `cancelBaseChange` is the definitional core of
  `concreteRationalToComplexLinearEquiv` (`Suggested.lean:266`), which every rational-to-complex
  construction routes through. README:333 then lists `cancelBaseChange` in the roadmap's own
  vocabulary, contradicting 144.
- README:237 says conjugation-equivariance of the abstract `fC` is "proved, not assumed". The file
  proves it for the concrete map only (`concreteRationalMapToComplex_conj`, `Suggested.lean:365`);
  after the abstract definition it proves `rationalMapToComplex_maps_WC` and nothing else.

**3. Blocking: nothing is ever instantiated.**

The stated chief deliverable is getting the definitions right, and there is no inhabitant of
`HodgeStructure`, `Polarization`, `MixedHodgeStructure` or `PeriodDomain.Point` anywhere. README:115
says the definitions are "checked on this instance", but the canonical tensor instance exercises the
base-change plumbing, not the Hodge conditions: nothing currently rules out a definition that no
object satisfies.

At least one explicit nonzero inhabitant should be blocking, and the Tate structure `ℤ(m)` is the
cheapest. A fuller worked-examples section, which most roadmaps here carry, would add an effective
weight-one structure from a lattice with a complex structure and its Riemann form, and a pure
structure viewed as mixed. That is also where the L0 companions at README:202 get pinned down: as
written, "`ℤ`-Tate twist, `⊗`/`Hom`/dual" does not fix weights, filtration shifts or signs, starting
with `ℤ(m)` having weight `−2m`.

**4. Say which category L1's conclusion is about, and give it an object.**

README:218 concludes "hence **the category of polarized `ℚ`-HS is semisimple**". Read literally, with
fixed polarizations and polarization-preserving morphisms, that category is not additive, since sums
and the zero map are not isometries, so semisimplicity is not the usual statement about it. The
classical theorem is about *polarizable* rational Hodge structures with ordinary Hodge morphisms.
Please say which is meant.

Either way the object is missing. There is no rational pure Hodge structure type: `HodgeStructure` is
definable only over an integral lattice, which is why `gradedPure` (`Suggested.lean:787`) re-spells
boundedness, antitonicity and opposedness by hand rather than saying "`gr^W_k` carries a Hodge
structure of weight `k`". Morphisms appear only as an unelaborated companion line at README:202, and
the one category targeted is `MHS.Hom` at README:246. Under "no gaps", please give the rational pure
object and the pure category their own targets; how to factor that against the integral case is your
call.

**5. The step from a form on each piece to a form on `V_ℂ` is not stated.**

README:220 reads "the Hodge-Riemann positivity makes `h(u,v) := i^{p−q} Q(u, v̄)` a positive-definite
Hermitian form on each piece, so `V_ℂ` carries a definite Hermitian form". `i^{p−q} Q(u, v̄)` is not
defined until `u` is homogeneous, so the "so" is doing real work. It can be discharged componentwise
off the L0 decomposition, but the reusable packaging is the Weil operator `C`, acting by `i^{p−q}` on
`H^{p,q}`, with `h(u,v) = Q(Cu, v̄)`. Neither `C` nor the extension step appears anywhere.

`C` is worth targeting in its own right: its definition from the decomposition, its action on pieces,
`C² = (−1)^n`, compatibility with `latticeConj` and `Q`, and positive-definiteness of `h`. Everything
downstream that wants a Hermitian form will want it.

**6. The Deligne splitting needs to be API, not a proof hint.**

README:239-240 names it correctly, as the canonical bigrading every MHS morphism respects, but 241
then says "establishing the splitting propositionally (existence of the `I^{p,q}` bigrading)
suffices". Propositional existence is enough to close the milestone and not enough to be usable. The
`I^{p,q}` bigrading is the basic API of mixed Hodge structures: please name it, with recovery of `F`,
recovery of `W`, the conjugation relation modulo lower weights, and functoriality.

**7. Smaller items.**

- README:88-89: the integral symmetry group is `Sp(2g, ℤ)` only when the form is a scalar multiple of
  the standard symplectic one, that is when the elementary divisors agree. A general nondegenerate
  integral alternating form gives a paramodular group, and the intrinsic statement is
  `Aut(V, Qint)`. The same sentence also runs together two correspondences: effective weight-one
  structures give complex tori, and the polarizable ones give abelian varieties.
- README:59-60 now says #33954 "was retired on 2026-08-11 in favour of" #42642. Roadmaps are
  timeless, and that is the class of dated claim
  [#172](https://github.com/TauCetiProject/TauCetiRoadmap/pull/172) removed elsewhere. Mathlib has no
  such API and [mathlib4#42642](https://github.com/leanprover-community/mathlib4/pull/42642) is
  adding one; which PR superseded which is not roadmap content.
- Same rule at `Suggested.lean:24-25`, which pins `leanprover/lean4:v4.31.0-rc1`, and README:338-340,
  which pins the toolchain and records the axiom-cleanliness check. Other roadmaps say only that the
  signatures elaborate against the pinned Mathlib.
- README:94-95, "is being developed concretely": either that work is a target here or the clause goes.
- The *Successor roadmap* section (README:265) is the ambitious tail CONTRIBUTING asks you to label a
  roadmap-for-a-roadmap, with contributors told not to follow it now. The pointer to #167 is right;
  it needs that sentence.
- README:123-125 calls `latticeConj` "the unique conjugate-linear map fixing the integral points". The
  file proves `latticeConj_ι` and `latticeConj_involutive` but not uniqueness, which is the lemma an
  instance author will want.
- README:132-133 pins `Polarization hs` as `{ Qint // IsPolarization hs Qint }`; `Suggested.lean:245`
  is a two-field structure. Equivalent, but the README names an exact spelling.
- `Suggested.lean:798` still calls `MixedHodgeStructure` "(schematic)", which looks stale now that
  `graded_pure` is fully encoded.
- The PR description still describes `TauCetiRoadmap/VHS/`, `Targets.lean`, milestones L0 through L5
  including the L4 variation structure and the L5 Schur lemma, and "Hodge structures and their
  variations". The branch is `HodgeStructures/`, `Suggested.lean`, L0 through L3, no variations.

0xSimex:
Hi all!

I took another look at the newly added Deligne-splitting part. One small but substantive Hodge-theoretic correction: I think the indices in the conjugation congruence are currently in the wrong order. In the standard convention for Deligne's canonical splitting, the relation is

```lean
I p q ≡ conj (I q p) mod ⨁ (r < p) (s < q), I r s
```

Equivalently, if one writes it with conjugation on `I p q`, it should be

```lean
conj (I p q) ≡ I q p mod ⨁ (r < q) (s < p), I r s
```

So in the roadmap's current formulation, the lower-bidegree bounds should be exchanged: `r < q, s < p`, rather than `r < p, s < q`.

See Peters–Steenbrink, *Mixed Hodge Structures*, §3.1 and the references therein.

I also have one minor terminology point in the `IsEffective` documentation: I would drop the phrase “(of level `n`)”. To me, a pure Hodge structure has "weight n" if non-zero terms have `p+q=n`, "effective" means that it's supported on non-negative indexes, and "level n" means that the `n` is the minimum `k` such that there are no nonzero terms with `|p-q|>=k`. Therefore, if such a structure is effective and has weight n (as in the definition), then we can only conclude its level is _at most_ n. For example, a pure Hodge structure of type (1,1) is effective, has weight 2, but level 0. Please correct me if I misinterpreted your definitions.

Apart from these minor corrections, I've nothing more to say about this PR from a human standpoint.

mrdouglasny:
Thanks Simone — both corrections are right and are in (`75cc9ed`).

You read `IsEffective` correctly: it is Hodge numbers supported in `[0, n]` (`F^0 = ⊤`,
`F^{n+1} = ⊥`), with nothing about level, so "(of level `n`)" was simply wrong and is gone.

The Deligne congruence now states `I^{p,q} ≡ conj (I^{q,p})` modulo `⨆_{r<p, s<q} I^{r,s}`, noting
that writing it with the conjugation on the other side swaps the bounds to `r<q, s<p`, and citing
Peters–Steenbrink §3.1. That was my error, introduced when the splitting was promoted from
propositional existence to a defined bigrading.

I hadn't appreciated when we started how close this is to your own work — a thesis on P=W in
non-abelian Hodge theory ([arXiv:2002.08929](https://arxiv.org/abs/2002.08929)) is rather more than
"a bit rusty". Thank you for catching a real error in it.

kim-em (dismissed):
Round-three items are all addressed, and the two remaining textual fixes are pushed as ee1b35e.

kim-em (approved):
Re-approving after the v4.34.0-rc1 adaptation; full `lake build` is green locally and in CI.

PR #53, opened 2026-07-03
Add roadmap: Dense graph limits and graphons (Lovász LNGL Part 3)

cameronfreer:
## Summary

Adds a roadmap for dense graph limits and graphons, following Part 3 of Lovász, *Large Networks
and Graph Limits*. Its spine is

```text
Graphon → homDensity → cutNorm → cutDist → GraphonSpace
        → counting → regularity → compactness → separation → convergence
```

The roadmap consumes Mathlib's finite-graph and measure-theory APIs and develops the graphon theory
that connects them. The principal endpoint is the equivalence between cut-distance convergence and
convergence of all homomorphism densities.

## Architecture

The specification pins five load-bearing choices:

- strict measurable graphon representatives, with the public quotient taken at `GraphonSpace`;
- a coupling-primary, cross-carrier `cutDist`;
- finite simple graphs with `Sym2` edges and injective density normalized by
  `Nat.descFactorial`;
- explicit carrier generality for each layer;
- neutral namespaces and Mathlib-compatible object kinds.

The development is organized into layers covering the core analytic objects, counting and weak
regularity, the a.e. view, compactness, coupling/pullback comparison, separation and convergence,
finite-graph validation, Lovász–Szegedy representability, sampling, and exchangeable graph-law
representation.

## Cross-roadmap boundary

The Exchangeability roadmap owns the generic relational-law and exchangeable-array API. This
roadmap owns graph-specialized laws, the graphon-mixture proof, and adapters to the generic API.
The graphon-mixture representation is independent of the functional Aldous–Hoover converse; the
generic representation theorem owes a specialization theorem showing that its graph case induces
the same infinite graph laws and quotient-valued graphon mixing measure.

## Grounding

The roadmap cites pinned snapshots of
[`math-commons/graphons`](https://github.com/math-commons/graphons) and
[`cameronfreer/graphon`](https://github.com/cameronfreer/graphon) as secondary proof-route and API
evidence. The roadmap itself specifies the target intrinsically and records the exact hypotheses,
normalizations, carrier choices, and compatibility obligations.

## Files changed

```text
README.md
TauCetiRoadmap.lean
TauCetiRoadmap/DenseGraphLimits/README.md
TauCetiRoadmap/DenseGraphLimits/Suggested.lean
```

## Verification

CI runs `lake build`; `sorry` warnings are expected in the human-owned roadmap specification.

## Authorship

Authored by [Michael R. Douglas](https://github.com/mrdouglasny/) and
[Cameron Freer](https://github.com/cameronfreer/), with assistance from GPT Pro, Claude Code,
Codex, Gemini, lean4-skills, and lean-lsp-mcp.

🤖 Directed by Cameron Freer, revised by GPT 5.6 Sol (xhigh reasoning)


TauCetiRoadmap/DenseGraphLimits/README.md
```diff
@@ -0,0 +1,402 @@
...
+lemma, compactness, separation) are milestones inside the fuller development, not the whole of
+it; each object gets its complete basic API.
+
+**Suggested home:** `TauCeti/Combinatorics/DenseGraphLimits/`.
+
+## Conventions (pinned up front)
+
+Decided now so contributors don't oscillate between incompatible designs. Extended rationale for
+the carrier and cut-distance choices is in two design notes in the
+[`math-commons/graphons`](https://github.com/math-commons/graphons/tree/main/docs) repo.
```
kim-em: :robot: Claude says:

The rationale for the two most consequential pinned conventions — coupling-primary `cutDist`, and the strict-carrier-with-quotient-on-top design — is offloaded to design notes in an external `math-commons/graphons` repo. That won't do. The roadmap is the definitive, self-contained specification, and the reasoning behind a pinned convention is exactly the part a reviewer and an implementor need in front of them: it's where they'd push back, and where they'd look when a downstream choice seems forced. A link out isn't durable — the repo can move, be rewritten, or vanish — and it splits the spec across two places under two owners. Please inline the substance of both notes here: the actual argument for coupling-primary over map-primary, and for the strict carrier over an AE/`AEEqFun` carrier — not a pointer to the argument. If a convention is worth pinning, its justification is worth carrying in the roadmap.

TauCetiRoadmap/DenseGraphLimits/README.md
```diff
@@ -0,0 +1,402 @@
...
+Named extremal consequences as acceptance tests (**Goodman**, **Mantel**, **Sidorenko-`C₄`**),
+the W-random sampling-expectation lemma `E[t(F, G(n,W))] → t(F,W)`, and concrete rational density
+checks. These keep the definitions honest and give visible checkpoints before the deeper layers
+close.
+
+### Layer 8 — Lovász–Szegedy representability
+A graph parameter equals `t(·, W)` for some graphon iff it is multiplicative, normalized,
+reflection-positive, and `[0,1]`-bounded (LNGL Thm 5.54 / the moment problem for graphs). Best
+proved in coordination with a reflection-positivity development rather than re-derived here; it
+is sequenced late because it depends on that material, and it is required work.
```
kim-em: :robot: Claude says:

Layer 8 as written is a leap of exactly the kind the guide rules out. It rests on a "reflection-positivity development" that is not in Mathlib, not a target here, and not in any roadmap this one cites — an object named but never made a target, with "best proved in coordination with" standing in for grounding. That's not allowed: no layer may depend on material nobody is committed to building. Two acceptable resolutions, no third: either reflection positivity becomes an explicit target — here, or in a separate roadmap that this one cites as a dependency and that actually exists — or Layer 8 comes out until such a target does. Please don't leave it resting on undefined work.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
```diff
@@ -0,0 +1,492 @@
...
+are pinned here too — as are the endpoint milestones: Frieze–Kannan weak regularity,
+compactness/completeness of `GraphonSpaceI`, the coupling↔map `cutDistPullback`, the Layer-6b
+convergence equivalence, finite-graph compatibility (`finiteGraphGraphon`), and quotient-level
+separation.
+
+Objects whose precise Lean shape would force a premature API choice — the weak-regularity
+`Finpartition` adapter and the exact mod-null transport bundle — are described in `README.md` instead.
+(An `IsCoupling` structure/class is deliberately
+avoided: couplings aren't canonical, so a typeclass would pick an arbitrary one.)
+-/
```
kim-em, line 42: :robot: Claude says:

Open this file with the standard header every other roadmap's `Suggested.lean` carries, which this one is currently missing:

> **This file is not the roadmap and is not exhaustive.** The definitive document is `README.md`. The statements here suggest Lean forms for particular milestones, so that contributors and reviewers converge on names and signatures; discharging all of them finishes neither a layer nor the roadmap.

This is now written into the roadmap-writing guide (the "Write Lean code" bullet in the top-level `README.md`), so it's the settled convention rather than a one-off request.

TauCetiRoadmap/DenseGraphLimits/README.md
cameronfreer: Done — inlined, and dropped the external `math-commons/graphons/docs` link. The Conventions block now carries a **"Why these two choices"** subsection with the actual rationale for both load-bearing choices:

- **Coupling-primary `cutDist`.** It is cross-carrier by construction — well-defined for graphons on arbitrary probability spaces, needing no common carrier and no standard-Borel/atomless hypothesis even to be *stated*; symmetry is a direct swap-coupling theorem (`cutDist_comm`) and the triangle inequality is the gluing lemma. The measure-preserving-map form is not more general: it **equals** the coupling form under atomless standard-Borel hypotheses (Layer 5, `cutDist_eq_cutDistPullback`), so making it primary would only bake those hypotheses and a common carrier into the basic object.
- **Strict carrier.** The strict kernel is a pointwise `ℝ`-module, so `U − W` and `c • W` are literal kernels (what the cut norm acts on); the `AEEqFun` view enters once, at the named Layer-3 bridge, rather than pervading Layers 1–2.

The Provenance/Acknowledgements pointers to the source repos stay (migration sources, not the rationale). Inlined in f193d2d.

---

🤖 written by Claude, reviewed by Codex, directed by Cameron

TauCetiRoadmap/DenseGraphLimits/README.md
cameronfreer: Done — pinned it as a real target here rather than deferring to an external development. Layer 8 is now split into **8a** (quantum graphs / reflection positivity) and **8b** (representability), with the objects added to `Suggested.lean` as sorry-signatures:

- `LabeledGraph k` — a finite simple graph with an *injective* `k`-labeling — and `LabeledGraph.glue`;
- `GraphParam` with an explicit `IsIsoInvariant`; `connectionMatrix f A` over a finite index family `A : ι → LabeledGraph k` (with the entry law `connectionMatrix_apply`); `IsReflectionPositive` = every finite `connectionMatrix` is `PosSemidef` (finite principal blocks, **not** one infinite matrix); `IsMultiplicative`; `IsNormalized`;
- `lovasz_szegedy_representability` — the iff, with `IsIsoInvariant` among the hypotheses, stated over the canonical carrier `Graphon I (volume)`.

Connection matrices aren't in Mathlib, so they're built here. Synced the Compiled-there list, the reviewer checklist, and the header docstring. Pinned in f193d2d; the injective-labels / finite-connection-matrix / `IsIsoInvariant` / canonical-carrier refinements are in c920cef; the universe-polymorphic (`{ι : Type*}`) reflection positivity and the `connectionMatrix_apply` entry law are in 54cd88b. `lake build` green.

---

🤖 written by Claude, reviewed by Codex, directed by Cameron

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: Already in place — the standard header note was added in 95110d0 (this review was on the pre-note diff). `Suggested.lean` now opens with **"`README.md` is the definitive roadmap document … This file is not the roadmap and is not exhaustive …"**, matching the convention the other roadmaps' `Suggested.lean` carry. No further change needed here.

---

🤖 written by Claude, reviewed by Codex, directed by Cameron

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
```diff
@@ -0,0 +1,553 @@
...
+yielding a finite simple graph (the unlabeled result). -/
+def LabeledGraph.glue {k : ℕ} (G₁ G₂ : LabeledGraph k) : Σ m, SimpleGraph (Fin m) := sorry
+
+/-- **Layer 8 (graph parameter).** A real-valued parameter of finite simple graphs (indexed over
+`Fin`-representatives); isomorphism invariance is imposed separately as `IsIsoInvariant`. -/
+abbrev GraphParam := (n : ℕ) → SimpleGraph (Fin n) → ℝ
+
+/-- **Layer 8 (isomorphism invariance).** `f` agrees on isomorphic graphs — the standing hypothesis
+that makes `f` a genuine graph parameter rather than a labelling-sensitive function on `Fin n`. -/
+def IsIsoInvariant (f : GraphParam) : Prop := sorry
```
kim-em: :robot: Claude says:

`IsIsoInvariant`, `IsMultiplicative`, and `IsNormalized` are `def … : Prop := sorry`. The body is `sorry`, so each is `sorryAx Prop` — a named proposition with no content, and three of the five hypotheses of `lovasz_szegedy_representability` therefore assert nothing.

The guide already forbids this failure mode — "never a field of type `Prop` … the condition asserts nothing" (top-level `README.md`); a `def _ : Prop := sorry` is the same thing one construct over. `sorry` stands in for a proof you can't yet supply, or a statement whose Mathlib API doesn't exist yet — never for the definition of a proposition you can state today. These are one-liners (`≃g`, disjoint union, the one-vertex graph are all in Mathlib), like their neighbours `IsReflectionPositive` and `IsCoupling`, which carry real bodies. Writing `IsMultiplicative` out also forces the disjoint-union reindex (`Fin n₁ ⊕ Fin n₂ → Fin (n₁+n₂)`) into the statement instead of the docstring.

Either define all three, or drop them: it is better to simply omit a condition than to write `sorry` in place of an actual `Prop`.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
```diff
@@ -0,0 +1,553 @@
...
+/-- **Layer 8a.** The connection-matrix entry law: `f` evaluated on the glued graph of `A i` and
+`A j` (its size and simple graph read off the gluing). -/
+theorem connectionMatrix_apply (f : GraphParam) {k : ℕ} {ι : Type*} [Fintype ι]
+    (A : ι → LabeledGraph k) (i j : ι) :
+    connectionMatrix f A i j = f ((A i).glue (A j)).1 ((A i).glue (A j)).2 := sorry
+
+/-- **Layer 8a (reflection positivity).** `f` is reflection-positive when every finite connection
+matrix is positive semidefinite — i.e. every finite principal block of each `M(f, k)` is PSD. Stated
+over finite index families (in any universe) rather than as one infinite matrix. -/
+def IsReflectionPositive (f : GraphParam) : Prop :=
```
kim-em, line 1100: :robot: Claude says:

`IsReflectionPositive` quantifies over `{ι : Type*}` in every universe, but the rest of the roadmap pins finite graphs to `SimpleGraph (Fin n)` representatives (the reviewer checklist even asks for "no universe-restricted `{V : Type}`"). Any finite `ι` reindexes to `Fin (Fintype.card ι)` with PSD preserved, so `∀ n, (connectionMatrix f (A : Fin n → LabeledGraph k)).PosSemidef` is equivalent and matches the convention. Is the universe generality load-bearing, or should this be `Fin n`?

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: Done — all three now carry real bodies (in 7c08925):

- `IsIsoInvariant`: agreement along `≃g` — `∀ (n₁ n₂) (F₁ : SimpleGraph (Fin n₁)) (F₂ : SimpleGraph (Fin n₂)), Nonempty (F₁ ≃g F₂) → f n₁ F₁ = f n₂ F₂`;
- `IsMultiplicative`: `f (n₁ + n₂) ((F₁ ⊕g F₂).map finSumFinEquiv.toEmbedding) = f n₁ F₁ * f n₂ F₂` — with, as you predicted, the disjoint-union reindex `Fin n₁ ⊕ Fin n₂ → Fin (n₁ + n₂)` in the statement rather than the docstring;
- `IsNormalized`: `f 1 ⊥ = 1`.

Also added a reviewer-checklist item barring the failure mode outright ("Do the structural predicates carry real bodies — never `def … : Prop := sorry`, which asserts nothing?"). `lake build` green.

---

🤖 written by Claude, reviewed by Codex, directed by Cameron

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: Not load-bearing — changed to `Fin n` (in 7c08925). `IsReflectionPositive` now reads

```lean
∀ (k n : ℕ) (A : Fin n → LabeledGraph k), (connectionMatrix f A).PosSemidef
```

matching the `Fin`-representative convention; as you note, any finite `ι` reindexes to `Fin (Fintype.card ι)` with PSD preserved, so nothing is lost. `connectionMatrix` itself stays polymorphic in `ι` — the generality is harmless in a def, and the convention constrains the statements.

---

🤖 written by Claude, reviewed by Codex, directed by Cameron

kim-em:
:robot: Codex says:

Thanks—the five earlier review threads are addressed in the current head, and I have resolved them. I think three specification gaps remain before approval:

1. **The Aldous–Hoover endpoint needs an explicit dependency and a graphon-specific target.** The Exchangeability roadmap already assigns exchangeable arrays and Aldous–Hoover to its Layer 8, describing it as a substantial tower with its own prerequisites. This roadmap currently names the same development without citing that dependency. Please make DenseGraphLimits consume Exchangeability Layer 8, then pin what this roadmap builds on top of it: the bridge between graphon sampling and symmetric jointly exchangeable Bernoulli arrays or infinite exchangeable random graphs.

2. **Layer 9 does not pin the joint sampling architecture needed for its almost-sure claims.** `sampleGraph W n` gives a marginal law for each `n`; an almost-sure convergence statement across `n` needs all the samples represented on one probability space, or it must instead be stated through per-`n` concentration bounds from which such a coupling and Borel–Cantelli argument are constructed. Please pin the intended object and theorem shapes, distinguish almost-sure convergence from convergence in probability or distribution, and name the concentration route. The weak-convergence stack currently cited does not by itself supply that specification. This can naturally be resolved together with the exchangeable-graph bridge above.

3. **Layer 8 still jumps from the reflection-positivity predicates to the full representability theorem.** Layer 8a defines labeled graphs, gluing, connection matrices, and the four structural predicates, but saying the gluing algebra’s laws are “expressible” does not make their theory a target. Please pin a concrete intermediate proof spine. For example, the cited Lovász–Szegedy route passes from connection-matrix positivity to positivity of the finite fully-labelled matrices, through the Möbius transform `f† ≥ 0`, then constructs random finite graph laws and uses concentration/Borel–Cantelli to obtain a convergent graph sequence before producing its graphon representation. A quantum-graph or moment-algebra route would instead need its own basic algebra and intermediate results pinned. Whichever route is intended, its dependencies—including any use of Layer 9—need to appear in the roadmap’s ordering rather than being hidden inside the headline theorem.

cameronfreer:
:robot: Fable 5 (directed by Cameron and reviewed by GPT 5.6 Sol) says:

Thanks — the three gaps are addressed across the six new commits (plus a merge of `main`), with one deliberate deviation on gap 1 explained below. (Independently useful context: [`cameronfreer/graphon`](https://github.com/cameronfreer/graphon) has advanced substantially and is now sorry-free with no custom axioms; the provenance section is refreshed and pinned at commit [`6eccca5`](https://github.com/cameronfreer/graphon/tree/6eccca5bbe5c9df46d7129bf59575b8b9b1d6699), which supplies proved routes for the Layer-9 stack and part of the Layer-8 algebraic substrate.)

**1. The Aldous–Hoover endpoint** — pinned, but **decoupled rather than made a dependency**. Layer 9 now ends at a fully pinned graph-level target: `ExchangeableGraphLaw` (consistent finite marginals under every label injection), `sampleExchangeableLaw` with the anchor `upperMass_sampleExchangeableLaw : P(F ≤ G(k,W)) = t(F,W)`, `IsDissociated` with the extremality target `exists_graphon_of_isDissociated`, the infinite form + compactness extension, and the summit `graphonMixtureLawEquiv : ProbabilityMeasure GraphonSpaceI ≃ InfiniteExchangeableGraphLaw` — uniqueness explicitly on the graphon quotient, never among raw kernel representatives. The correspondence is pinned **beyond its Dirac fibers**: alongside the Dirac anchor, the general mixture-coordinate law `graphonMixtureLawEquiv_upperMass` gives, for *every* mixing measure `P`, `upperMass F = ∫ t(F,·) dP` on the quotient — upper masses determine each finite marginal by Möbius inversion, so an arbitrary permuting `Equiv` cannot satisfy it (the precedent is the library's `infiniteMixtureLawEquiv_law_map_restrictFin`). The roadmap names this the **Diaconis–Janson graphon-mixture representation** (a graph-level Aldous–Hoover *consequence*), deliberately not "Aldous–Hoover". The reason we did not make DenseGraphLimits consume Exchangeability Layer 8: the prior formalization proves the graph-level correspondence (`infiniteMixtureLawEquiv`, `InfiniteRepresentation.lean`) by compactness extension plus the mixture representation, **with no array-level input** — a dependency edge would overstate the proof dependency. Instead the array-level theory is declared independent parallel work, joined by the bridge you asked for, recorded as a **documented future interface** — proposed identifier `graphLawArrayLawEquiv`, equating exchangeable graph laws with symmetric, irreflexive, jointly exchangeable Boolean-array laws — prose-only until the Exchangeability roadmap's array API exists (per the roadmap guide's rule for conditions that cannot yet be stated). Diaconis–Janson (2008) added to References.

**2. Joint sampling architecture** — pinned. `infiniteSampleLaw W` is the law of the infinite `W`-random graph on one probability space (i.i.d. positions + one uniform per pair; Mathlib's `MeasurableSpace (SimpleGraph V)` covers `V = ℕ`), with `infiniteSampleLaw_map_restrictFin` identifying every `G(n,W)` as a restriction of this single object, and `infiniteSampleLaw_eq_extension` identifying the explicit sampler with the compactness extension of the sampling marginals — which certifies the sampler's exchangeability and welds the sampling, graph-law, and mixture subsections into one architecture. The two convergence modes are now distinct pinned theorems with distinct named routes: **in probability** (`sampleGraph_cutDist_tendsto_inProbability`, LNGL 10.16, on the marginals, via the two-stage point-sampling/Azuma + rounding/union-bound decomposition) and **almost surely** (`infiniteSampleLaw_ae_tendsto_cutDist`, on the joint space, via the per-coordinate concentration bound `sampleGraph_injHomDensity_concentration` — LNGL Prop 11.32 shape, summable tails — Borel–Cantelli per fixed `F`, countable intersection, and the Layer-6b convergence equivalence). Neither proof consumes the other, and the weak-convergence stack previously cited here is no longer load-bearing for these targets.

**3. The Layer-8b spine** — pinned as a chain of named targets, and the dependency on Layer 9 is now explicit in the ordering. The route: reflection positivity → `graphParamMobius_nonneg` (`f† ≥ 0`, via PSD fully-labelled blocks and the Möbius congruence) and `graphParamMobius_sum_eq_one` (`∑ f† = 1`) → the random graph law `paramExchangeableLaw` (`L_f`) → `paramExchangeableLaw_upperMass` (`upperMass L_f = f`) → `isDissociated_paramExchangeableLaw` → Layer 9's `exists_graphon_of_isDissociated` gives `L_f = sampleExchangeableLaw W` → the sampling anchor closes `f F = t(F,W)`. Note the precise scope of the Layer-9 dependency: the spine consumes the **graph-law representation/extremality infrastructure only, not the graphon-sampling concentration theorems** — while the spine runs, no representing graphon exists yet, so the classical random-graphs-plus-convergent-subsequence route would additionally need an `f†`-specific variance or simultaneous-selection lemma; the extremality route needs none. The Ordering section now reads Layer 8a → Layer 9's graph-law layer → Layer 8b.

`lake build` is green on the new head (`Suggested.lean` imports only Mathlib; the new declarations are the pins listed above, with `graphParamMobius`, `upperMass`, `IsDissociated`, `restrictFin`, and the structures `ExchangeableGraphLaw` and `InfiniteExchangeableGraphLaw` given real bodies).

*Directed by Cameron Freer, written by Fable 5 high, reviewed by GPT 5.6 Sol xhigh.*

cameronfreer:
> Thanks—the five earlier review threads are addressed in the current head, and I have resolved them. I think three specification gaps remain before approval:

Thanks for this review. As noted in the longer reply, my [cameronfreer/graphon](https://github.com/cameronfreer/graphon) repo is now farther along, and it contains a lot of new relevant details.

Also I think it makes more sense for this roadmap to target the graphon-specific variant of Aldous-Hoover, while the jointly (or separately) exchangeable array versions could be later extensions to the Exchangeability roadmap and eventually meet up with this one somehow.

The other two points were more straightforward.

cameronfreer:
A further external review of the head prompted another revision round — eight new commits (plus merges of `main`; the index entry is now 16, after Representation Theory and Modular Forms landed). The organizing principle is the one your reviews kept enforcing: no important theorem concealed inside a constructor or an endpoint. One position from my previous reply is partially walked back below.

- **Layer 9 is split into 9a/9b/9c** (sampling / graph laws and mixtures / convergence), so "Layer 9's graph-law layer" in the Ordering section is now a real numbered unit with its own dependencies (9b consumes 9a, Layer 4's compactness, and the hom-density continuity/separation algebra of Layers 1–2 and 6a, and feeds 8b; 9c consumes 9a + 6b only).
- **The empirical-mixing existence spine is pinned (9b).** The summit `graphonMixtureLawEquiv` pinned its endpoint, Dirac anchor, and mixture coordinates but not how the mixing measure is *obtained* — the same gap-shape as your Layer-8b objection, one layer over. The route is now named targets matching the proved provenance repo: the mixture map with its coordinate law and Dirac fiber, `empiricalMixing`, the `k²/(n+1)` collision estimate, the compactness extraction (the exact point where Layer 4's `CompactSpace GraphonSpaceI` is consumed — previously invisible), limit identification, existence, and uniqueness. The finite-level equivalence is assembled with a **real body** by `Equiv.ofBijective` from the pinned existence and injectivity (so its `_apply` law is definitional), the summit is its transport along the finite↔infinite extension — also a real body — and `exists_graphon_of_isDissociated` is now positioned after the spine and documented as its short consequence rather than a second unexplained representation theorem. The prose that read "compactness extension plus the mixture representation" conflated those two constructions; they are now separate named steps.
- **The Möbius consistency calculus is exposed (8b).** `paramExchangeableLaw := sorry` absorbed a real theorem: `f† ≥ 0` and `∑ f† = 1` give a probability mass at each fixed level only, and the identity binding the levels under every label injection is the actual content. It is pinned at two strengths — the raw mass identity `graphParamMobius_sum_comap` (iso-invariance + multiplicativity + normalization; reflection positivity deliberately not a hypothesis) and its measure packaging `paramGraphLaw` + `paramGraphLaw_map_comap` (reflection positivity enters only to keep `ENNReal.ofReal` faithful) — and `paramExchangeableLaw` is now a visible assembly with a real body. The new bridge `isDissociated_iff_upperMass_mul` (a genuine theorem: dissociation equates product laws, upper-mass multiplicativity constrains only upper events, two-window Möbius inversion closes the gap) makes the dissociativity arrow a visible computation.
- **The concentration route is migrated (9c).** The pinned `t₀` concentration described its proof as "changing one sampled vertex moves `t₀` by at most `k/n`" — but the sampled graph also carries one edge coin per pair, so that is not yet a McDiarmid setup. Replaced by the proved padded-exposure route on the **ordinary** hom density: exposure source (position + padded coin row per vertex), the law identification `map_exposedSample`, the `q/n` oscillation bound, the `2·exp(−ε²n/(2q²))` tail with its summability bridge. `injHomDensity` keeps its unbiasedness anchor in 9a but is no longer the load-bearing engine, and the in-probability endpoint gains the Mathlib-idiomatic `TendstoInMeasure` joint packaging.
- **The cross-roadmap boundary is revised — this is the partial walk-back on your first point.** "Independent parallel theories, nothing shared in either direction" was too strong, and it ducked the ownership question your suggested dependency was really about. The paragraph now separates the two questions: the generic exchangeable relational-law API (consistent finite/infinite laws, relabeling invariance, dissociation, ergodicity, extremality) is **owned by the Exchangeability roadmap**, and this roadmap's graph-specialized structures carry an explicit **refactor-onto commitment** once that shared API lands — an interface obligation rather than a build dependency, so the proved graphon-mixture theorem is not gated on the (genuinely hard, still open) generic functional converse, which instead owes a specialization/compatibility result. The carrier level of the meeting point is also pinned **now** rather than left as prose: `EdgeIndex`, `graphCoordEquiv : SimpleGraph ℕ ≃ (EdgeIndex → Bool)` with measurability both ways, and the relabeling commuting square; only the law-level `graphLawArrayLawEquiv` still waits, for its target type, on the array API.
- **Provenance:** the `6eccca5` migration pin is retained (every proved route the revised Layer 9 pins is checkable there, and the file list now names them), and a second immutable snapshot `bbc82c4` records the newer generic relational framework — including that its hard functional converse is explicitly open (`RankRepresentation` is an interface with no existence theorem at any rank), which is the factual ground for the boundary above.

`lake build` is green on the new head (expected `sorry` warnings only); the PR description is updated to match. Attribution is also sharpened to theorem granularity (mixture existence = the symmetric Boolean specialization of Aldous–Hoover; Diaconis–Janson Thm 5.3 = the law correspondence; Cor 5.4 / Thm 5.5 = the extreme/dissociated case), keeping the settled "Diaconis–Janson graphon-mixture representation, a graph-level Aldous–Hoover consequence" wording.

*Directed by Cameron Freer, written by Fable 5, reviewed by Codex.*

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
```diff
@@ -0,0 +1,1133 @@
...
+  prob : IsProbabilityMeasure law
+  /-- Invariance under every relabeling. -/
+  exchangeable : ∀ e : Equiv.Perm ℕ, law.map (SimpleGraph.comap ⇑e) = law
+
+/-- **Layer 9b (finite ↔ infinite).** Consistent finite marginals extend uniquely to an exchangeable
+law on infinite graphs (a compactness extension), and restriction along `restrictFin` recovers the
+marginals — the packaging that lets the representation below be stated in the infinite form. Prior
+formalization: `InfiniteLaw.lean` + `InfiniteExchangeability.lean`. -/
+def exchangeableGraphLawEquivInfinite :
+    ExchangeableGraphLaw ≃ InfiniteExchangeableGraphLaw := sorry
```
kim-em, line 771: :robot: Claude and Codex say:

`exchangeableGraphLawEquivInfinite` is an opaque `Equiv := sorry` that no **compiled** theorem constrains beyond sampling laws — the docstring and README prose do prescribe projective-limit extension with marginal recovery, but nothing in Lean pins it — and the two summit anchors that appear to constrain it are vacuous about it:

- `graphonMixtureLawEquiv_upperMass` is stated about `exchangeableGraphLawEquivInfinite.symm (graphonMixtureLawEquiv P)`. Since `graphonMixtureLawEquiv := mixtureExchangeableLawEquiv.trans exchangeableGraphLawEquivInfinite`, that expression collapses by `Equiv.symm_apply_apply` to `mixtureExchangeableLawEquiv P`, which is `mixtureExchangeableLaw P` definitionally — so the lemma is the finite-level `upperMass_mixtureExchangeableLaw` restated, and it holds for any `exchangeableGraphLawEquivInfinite` whatsoever.
- `graphonMixtureLawEquiv_dirac` equates `exchangeableGraphLawEquivInfinite (mixtureExchangeableLawEquiv (dirac ⟦W⟧))` with `exchangeableGraphLawEquivInfinite (sampleExchangeableLaw W)`, which follows from the finite-level Dirac fiber `mixtureExchangeableLaw_diracProba` by congruence — again for any `exchangeableGraphLawEquivInfinite`.

The only compiled constraint on the extension is `infiniteSampleLaw_eq_extension`, and it touches only sampling laws. So an implementation could compose the true extension with an arbitrary permutation of `InfiniteExchangeableGraphLaw` fixing the images of sampling laws: every compiled theorem stays provable, while `graphonMixtureLawEquiv P` need not be the infinite law whose finite marginals are `mixtureExchangeableLaw P` — the named bijection is then not pinned as the intended graphon-mixture correspondence. This is the same arbitrary-`Equiv` failure mode the previous rounds closed for the finite-level equivalence, one construction over.

One lemma fixes it — the analogue of the provenance repo's `infiniteMixtureLawEquiv_law_map_restrictFin` that the README already cites:

```lean
theorem exchangeableGraphLawEquivInfinite_law_map_restrictFin
    (L : ExchangeableGraphLaw) (k : ℕ) :
    ((exchangeableGraphLawEquivInfinite L).law).map (restrictFin · k) = L.law k := sorry
```

Every cylinder in Mathlib's σ-algebra on `SimpleGraph ℕ` (comapped from `Adj`) involves finitely many adjacency coordinates, hence lies in some initial `k × k` window, so equality of all `restrictFin` pushforwards determines the law on a generating π-system — this pins the forward map uniquely, hence the whole `Equiv`, and makes the transported summit canonical rather than merely bijective. Optionally also state its summit-level transport, the direct analogue of the cited provenance theorem:

```lean
theorem graphonMixtureLawEquiv_law_map_restrictFin
    (P : MeasureTheory.ProbabilityMeasure GraphonSpaceI) (k : ℕ) :
    ((graphonMixtureLawEquiv P).law).map (restrictFin · k)
      = (mixtureExchangeableLaw P).law k := sorry
```

Please also recalibrate the "pinned beyond its Dirac fibers" reviewer-checklist item to match: the finite-level equivalence is fully pinned (a real body via `Equiv.ofBijective`), but the extension half of the summit currently is not, and the summit-level `_dirac` / `_upperMass` anchors, being formal consequences of the finite-level lemmas, cannot do that pinning.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
kim-em, line 1: :robot: Claude and Codex say:

Three unpinned constructions, in decreasing order of weight. None blocks on its own — in each case the README/docstring prose is unambiguous, and the README is the definitive document — but the first is the same genre this review has been closing and deserves a compiled contract.

- `LabeledGraph.glue` is a `sorry` def with no characterization lemma, yet `connectionMatrix_apply` characterizes the connection matrix only *relative to whatever `glue` happens to be*, so the definition of reflection positivity — and the downstream 8b branch consuming that hypothesis — formally rests on an unconstrained operation. By this file's own principle (`stepGraphon` gets `stepGraphon_apply` because the bare constructor doesn't expose the API), `glue` needs a compiled contract: canonical vertex maps from both sides (or an equivalent quotient construction) with (i) corresponding labels identified, (ii) no other identifications, (iii) joint surjectivity, and (iv) adjacency the union of the two images. A cardinality law alone would not pin adjacency or reindexing.
- `overlayDiff` could carry the fully stateable pin
  `∀ p q : Ω₁ × Ω₂, (overlayDiff μ₁ μ₂ U W π hπ).toFun p q = U.toFun p.1 q.1 - W.toFun p.2 q.2`
  rather than prose only (symmetry follows from `U`'s and `W`'s, so this is consistent with `symm'`).
- `exposureMeasure` leaves the coin-row measure as an inner `sorry`, though the signature already fixes the coin carrier to `ℝ`, so it is stateable today:
  `Measure.pi fun _ : Fin n => (volume : Measure ℝ).restrict (Set.Icc 0 1)`.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
```diff
@@ -0,0 +1,1133 @@
...
+/-- **Layer 9a (finite-graph hom density).** `t(F, G) = hom(F,G) / m^{|V(F)|}` for a finite target
+graph `G` on `Fin m`. Defined via `Nat.card` (no `Fintype`/decidability on the hom type or on `G`). -/
+def homDensityFin {V : Type*} [Fintype V] (F : SimpleGraph V) {m : ℕ} (G : SimpleGraph (Fin m)) : ℝ :=
+  (Nat.card (F →g G) : ℝ) / (m ^ Fintype.card V : ℝ)
+
+/-- **Layer 9a (injective hom density `t₀`).** The *ordered injective* hom count over the **falling
+factorial `(m)_k = m.descFactorial k`** (`k = |V(F)|`) — **not** `Nat.choose m k`, which would bias
+the sampling estimator by `k!`. Via `Nat.card`; no decidability on the target graph `G`. -/
+def injHomDensity {V : Type*} [Fintype V] (F : SimpleGraph V) {m : ℕ} (G : SimpleGraph (Fin m)) : ℝ :=
+  (Nat.card {φ : F →g G // Function.Injective φ} : ℝ) / (m.descFactorial (Fintype.card V) : ℝ)
```
b-mehta, line 515: [SimpleGraph.Copy](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Copy.html#SimpleGraph.Copy) or its related things may help here

b-mehta: [SimpleGraph.labelledCopyCount](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Copy.html#SimpleGraph.labelledCopyCount), actually

kim-em:
The README grew a "Working with Mathlib" group after this PR opened (https://github.com/TauCetiProject/TauCetiRoadmap/pull/165, doc: state the Mathlib relationship and the missing roadmap principles), and one of its three rules is **never push work to Mathlib**: Tau Ceti material is built here and stays here, Mathlib contributors are welcome to take any of it at any time, and deciding what Mathlib absorbs is solely theirs. So a roadmap shouldn't schedule its own upstreaming, or hold an item back for it. The merged roadmaps had the same problem and got the same sweep in https://github.com/TauCetiProject/TauCetiRoadmap/pull/169 (doc: conform roadmaps to "never push work to Mathlib"), so this isn't aimed at yours in particular.

Three places to change:

- **`### Upstream to Mathlib`** (`TauCetiRoadmap/DenseGraphLimits/README.md:520`) should go entirely. The inventory inside it is good and worth keeping — the measure-preserving map from `(I, volume)`, the dyadic-martingale `L¹` lemmas, the `Measure.pi` curry–uncurry lemmas, the `AEEqFun` ergonomics — but as a statement about *where they live in Tau Ceti*: each is independent of graphons, so each belongs in a general home rather than under `DenseGraphLimits/`. That is the useful content; the destination isn't ours to set. "No upstreaming is scheduled before Layers 1–4 are complete in `TauCeti/`" should just go, since nothing here is scheduled.
- **`:147`**, "each a strong upstream candidate once its API is stable". Same substitution: say they are reusable beyond graphons, which is what makes the sequencing argument work.
- **`:641`**, "The Mathlib upstreaming follows." Drop the sentence.

Worth saying explicitly, since the section is well argued: the *sequencing* instinct behind it is right. Stabilizing an API before spreading it around is good practice, and that argument survives intact once it's about where the material sits in `TauCeti/`.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: Thanks, agreed and addressed in d2ba497.

🤖 Claude summary:

`exchangeableGraphLawEquivInfinite_law_map_restrictFin` is now a named `@[simp]` theorem. It is the composition of the migration source's `infiniteLaw_map_restrictFin` with the equivalence application law, so every finite marginal is pinned. The summit-level form follows formally from this together with the definition of `graphonMixtureLawEquiv`, so it is derived rather than duplicated.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: Thanks, that is useful. See d2ba497.

Now `injHomDensity` requires either restating the numerator through `SimpleGraph.labelledCopyCount` or else proving the explicit equivalence.

I'll leave this thread unresolved for the moment, since I'm testing one more related use of it.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: Thanks, addressed in 1cec4e5:

🤖 Claude says:

`overlayDiff` now carries its pointwise eliminator `overlayDiff_apply`; the exposure coin-row measure is given outright (n independent `[0,1]`-uniforms, no inner `sorry`); and `LabeledGraph.glue` is pinned by `glueInl`/`glueInr` with label identification, joint surjectivity, and faithful adjacency on both sides. The README's new constructor requirements (adversarial regressions with every contract; explicit acting family for every coupling) make this a standing rule rather than a per-comment fix.

cameronfreer:
> The README grew a "Working with Mathlib" group after this PR opened (#165, doc: state the Mathlib relationship and the missing roadmap principles), and one of its three rules is **never push work to Mathlib**: Tau Ceti material is built here and stays here, Mathlib contributors are welcome to take any of it at any time, and deciding what Mathlib absorbs is solely theirs. So a roadmap shouldn't schedule its own upstreaming, or hold an item back for it. The merged roadmaps had the same problem and got the same sweep in #169 (doc: conform roadmaps to "never push work to Mathlib"), so this isn't aimed at yours in particular.

Thanks, see 1cec4e5.

🤖 Claude says:

The "Upstream to Mathlib" section is now "Reusable beyond graphons — general Tau Ceti homes" — the inventory is kept, stated as where the material sits in TauCeti/; "strong upstream candidate" and "the Mathlib upstreaming follows"/"no upstreaming is scheduled" are gone at all three sites. The sequencing argument survives intact as API stabilization before relocation.

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: ok `injHomDensity` is now more fully fixed to use mathlib in ad91ec0: `card_injective_hom_eq_labelledCopyCount` shows that the numerator equals `G.labelledCopyCount F`

TauCetiRoadmap/DenseGraphLimits/Suggested.lean
cameronfreer: further update:

🤖 Claude says:

Fixed in a57d885 + c9f6924: `glueInl`/`glueInr` are embeddings, `glueInl_eq_glueInr_iff` pins the intersection at exactly the corresponding labels, `glue_surjective` gives coverage, and `glue_graph` identifies the glued graph with `G₁.graph.map glueInl ⊔ G₂.graph.map glueInr` — the full adjacency iff follows by `sup_adj`/`map_adj`.

kim-em (approved):
:robot: Reviewed at `3169c1462d3df1c0a88c431e9348b17158fe022a`, including the conflict-resolution merge from current `main`; no blocking defect found. Approving.

PR #56, opened 2026-07-05
refactor(ContourIntegration): tighten three HW-summit signatures

CBirkbeck:
Three signature refinements to the suggested Lean forms in `ContourIntegration/Suggested.lean`, all surfaced while building the contour-integration layer in [TauCeti](https://github.com/TauCetiProject/TauCeti) and each closing a latent gap in the stated milestone:

### 1. `hungerbuhlerWasem_residueTheorem` — add `hbase : γ a ∉ S`
HW Thm 3.3 (and the AINTLIB reference development it ports) require the basepoint `γ a` to avoid the singular set: the principal-value excision is defined by removing `ε`-neighbourhoods of `S` at the *interior* crossings, and a singularity at the closed cycle's basepoint/endpoint makes that excision ill-posed. It is a harmless normalization — any cycle can be cyclically reparametrized so its basepoint lands off the finite set `S`.

This is the roadmap-side half of the corresponding TauCeti PR: without `γ a ∉ S` in the contract, the on-cycle Condition-B milestone cannot state its interior-only crossing hypotheses honestly.

### 2. `ConditionAprime` — add the function `f`
Condition (A′) is *flatness to the pole order of `f`* at each on-cycle singularity — the required sector count/order is read off from `meromorphicOrderAt f s`. The order cannot be named without `f`, so the previous `ConditionAprime γ a b S` was under-specified. Now `ConditionAprime γ a b f S`.

### 3. `homologyCauchyTheorem` — add `hγC1 : ContDiffOn ℝ 1 γ (Set.Icc a b)`
The Dixon/FTC argument needs `deriv γ` continuous for `∫ deriv γ • f (γ ·)` to be the genuine contour integral. The bare "closed cycle" hypothesis omitted this. A piecewise-`C¹` generalization can follow later; `C¹` is the honest minimum for the stated integral.

Docstrings updated to match each change. These are *suggested forms* only (README remains the definitive roadmap); the file still builds with its `sorry`-targets and no new elaboration errors (`lake build TauCetiRoadmap.ContourIntegration.Suggested` green).

### DIFF
--- a/TauCetiRoadmap/ContourIntegration/Suggested.lean
+++ b/TauCetiRoadmap/ContourIntegration/Suggested.lean
@@ -100,10 +100,10 @@ def IsNullHomologous (γ : ℝ → ℂ) (a b : ℝ) (Ω : Set ℂ) : Prop :=
   ∀ w ∉ Ω, windingNumber γ a b w = 0
 
 /-- HW condition **(A′)**: the cycle `γ` approaches each on-cycle singularity in `S` transversally,
-meeting it as a finite union of model sectors with a prescribed pole order (HW §3). One of the two
-regularity conditions that make the principal value exist; stated explicitly so the summit is
-honest. -/
-def ConditionAprime (γ : ℝ → ℂ) (a b : ℝ) (S : Finset ℂ) : Prop := sorry
+meeting it as a finite union of model sectors, flat to the pole order of `f` there (read off from
+`meromorphicOrderAt`, HW §3). One of the two regularity conditions that make the principal value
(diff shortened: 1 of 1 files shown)


CBirkbeck:
Closing per maintainer decision: we'll stay on the existing roadmap signatures and build HW 3.3 faithful to them, rather than refining the suggested forms ahead of the code. The three adjustments proposed here — `ConditionAprime`'s `f`, HW 3.3's `γ a ∉ S` basepoint avoidance, and `homologyCauchyTheorem`'s `C¹` regularity — can be revisited if and when the corresponding targets are actually built and their exact hypotheses are concrete.

PR #57, opened 2026-07-06
feat: add pivotal and spherical categories roadmap

kim-em:
This PR adds a roadmap for pivotal and spherical monoidal categories. It discharges the standing TODO in Mathlib's `CategoryTheory/Monoidal/Rigid/Basic.lean` ("Define pivotal categories (rigid categories equipped with a natural isomorphism `ᘁᘁ ≅ 𝟙 C`)") and specifies the theory it opens onto: the dual functor `C ⥤ (Cᵒᵖ)ᴹᵒᵖ` and the double-dual endofunctor built from it, pivotal structures as monoidal trivializations of the double dual, the left and right categorical traces and dimensions, spherical categories, quantum and Frobenius–Perron dimension, the two structural examples (`FDRep G` with its standard spherical structure, and the pointed categories `Vec^ω_G` whose pivotal structures form a torsor over `Hom(G, kˣ)`), Frobenius–Schur indicators, the classification of pivotal structures by characters of the universal grading group (Drinfeld–Gelaki–Nikshych–Ostrik), and the full synoptic chart of tensor categories from Henriques–Penneys–Tener (arXiv:1509.02937, Figure 2), including the Drinfel'd-centre arrows and the `balanced+rigid ≃ braided+pivotal` equivalence.

The core definitions are stated over a `RightRigidCategory` (both HPT trace formulas and the double dual use only right duals, and Mathlib registers `FDRep k G` as right rigid via `FGModuleCat`); the classification results are stated at the fusion-category-over-an-algebraically-closed-characteristic-0-field bar. `README.md` is the definitive spec; `Suggested.lean` pins the load-bearing objects and named milestones as `sorry`-targets and elaborates against the pinned Mathlib with only `sorry` warnings.

🤖 Prepared with Claude Code


CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `f04a2c1f6c00b6b2a528aa7c2c3c15b29ae9d9ca`, including the README and the proposed Lean interfaces. This is a mathematical/specification review, not a claim to have run a local Lean build.

The duality conventions, monoidality requirement on pivotal structures, finite-support condition on `VecTwisted`, and separation of the fusion hypotheses are useful. I found the following remaining issues.

### 1. The unrestricted spherical-to-ribbon assertion is false

**Locations:** README Layer 10 (including “Ribbon corresponds to the spherical pivotal structures under it” and the spherical-centre arrow); `centerBalancedOfSpherical` and `ribbon_center_of_spherical` in `Suggested.lean`.

The stated spherical condition is equality of left and right traces of all endomorphisms. Without additional hypotheses, it does not imply that the twist associated to a braided pivotal structure is ribbon. HPT, Appendix A.2, Proposition A.4, states the implication from ribbon to spherical, and adds semisimplicity for the converse. The current roadmap drops that restriction.

Here is an explicit counterexample to the proposed generality. Let `k` have characteristic three and let `C = FDRep k (C₃)` with its symmetric braiding. Twist the standard pivotal structure by the central generator `g` of `C₃`. On every representation write `ρ(g) = I + N`; then `N³ = 0`. Every equivariant endomorphism `f` commutes with `N`, so `Nf` is nilpotent. Consequently

`Tr(ρ(g) f) = Tr(f) = Tr(ρ(g⁻¹) f)`.

These are the two pivotal traces, so this pivotal structure is spherical in the roadmap's precise sense. Its associated twist is `ρ(g)` (or its inverse for the other convention). It is not ribbon: on the two-dimensional representation with `ρ(g) = [[1,1],[0,1]]`, the action on the dual is `ρ(g⁻¹)ᵀ`, whereas the dual of the twist is `ρ(g)ᵀ`; they differ in characteristic three.

This also obstructs the **induced** spherical-centre assertion at the displayed generality. The symmetric embedding `C → Z(C)` preserves the induced pivotal structure and braiding, and the induced twist restricts to the non-ribbon twist just described. Thus the induced twist on `Z(C)` cannot be ribbon either.

**Repair:** keep the definitions general, but state the spherical/ribbon converse under the actual semisimplicity/splitting hypotheses used in HPT. For the centre theorem, a safe intended scope is the fusion-category-over-an-algebraically-closed-characteristic-zero-field setting, with the requisite centre semisimplicity supplied. Alternatively, use an explicitly stronger compatibility hypothesis rather than trace equality alone.

There is a second interface issue here even after restricting the theorem: `centerBalancedOfSpherical` accepts an arbitrary `[PivotalCategory (Center C)]`, whereas its documentation promises the pivotal structure induced from `C`. Construct a named `centerPivotalOfPivotal`, prove that its components forget to those of the given structure on `C`, and build the twist with that instance. Remove the unrelated centre-pivotal input, or require an explicit compatibility equation. Merely choosing some centre pivotal structure through a `Nonempty` theorem does not express this contract.

Source: [HPT, Appendix A.2, Proposition A.4](https://arxiv.org/pdf/1509.02937).

### 2. The proposed Grothendieck-ring multiplication is not valid over the fields admitted by its signature

**Locations:** README Layers 3–4; `fusionCoeff` and `grothendieckRing`.

`fusionCoeff` is the `k`-dimension of `Hom(X_i ⊗ X_j, X_l)`, and `grothendieckRing` promises multiplication by these coefficients with unit `[𝟙]`. But `grothendieckRing` has neither `[IsAlgClosed k]` nor a splitting hypothesis on simple endomorphism rings. The README's preceding multiplicity formula has the same qualification missing locally.

Take finite-dimensional **real** representations of `C₃`, and let `V` be its irreducible two-dimensional rotation representation. All the displayed hypotheses of `grothendieckRing` hold, but `End(V) ≅ ℂ` as a real algebra. Thus

`fusionCoeff 𝟙 V V = dim_ℝ Hom(𝟙 ⊗ V, V) = 2`,

although `V` occurs in `𝟙 ⊗ V` with multiplicity one. The prescribed multiplication would therefore fail the unit law.

**Repair:** carry `[IsAlgClosed k]` through this interface, consistently with the advertised fusion bar, or use an explicit split-semisimple hypothesis `End(S) ≅ k` for each simple `S`. A genuinely nonsplit theory needs multiplicities over the relevant division endomorphism algebra, not the raw `k`-dimension. Add the real rotation example as a negative test for applying the split formula.

Source: [EGNO, §4.16](https://math.mit.edu/~etingof/egnobookfinal.pdf) discusses precisely the distinction between arbitrary-field and split tensor categories.

### 3. The adjoint-subcategory generators must be restricted to simple objects

**Location:** README Layer 9, definition of `C_ad` as generated by `X ⊗ Xᘁ`.

Please make `X` **simple** in this definition, ideally by indexing the generators by `SimpleClasses C`. Allowing arbitrary objects changes the subcategory and destroys the subsequent universal-grading claim.

For example, in `Vec_{C₂}`, let `g` denote the nontrivial simple object and put `X = 𝟙 ⊕ g`. Then

`X ⊗ Xᘁ ≅ 𝟙 ⊕ 𝟙 ⊕ g ⊕ g`.

The subcategory closed under direct summands generated by this object contains `g`, hence is the whole category. The actual adjoint subcategory of a pointed category is the neutral component. More generally, allowing `X = 𝟙 ⊕ Y` would put every simple `Y` in the generated subcategory.

**Repair/test:** use the simple-object generators of EGNO Definition 4.14.5, and check `C_ad(Vec_{C₂}) = Vec` and `U(Vec_{C₂}) = C₂` together. This is an important quantifier to pin before constructing the quotient/group.

### 4. The pivotal-functor equation has the composition order reversed for Lean's `≫`

**Location:** README Layer 1, “Pivotal functors”.

With `δ_X : F(Xᘁ) ≅ (F X)ᘁ`, the displayed

`F(φ_X) = δ_{Xᘁ}⁻¹ ≫ (δ_X)ᘁ ≫ φ_{F X}`

does not compose in Lean's left-to-right convention. The order was retained from a right-to-left `∘` formula. In `≫` notation the composable path is

`φ_{F X} ≫ (δ_X)ᘁ ≫ δ_{Xᘁ}⁻¹`,

from `F X` through `(F X)ᘁᘁ` and `F(Xᘁ)ᘁ` to `F(Xᘁᘁ)`. Use the corresponding `.hom`/`.inv` maps in a representative definition of a pivotal functor, and check identity/composition with that actual definition.

These changes leave the overall programme intact, but prevent false targets and incompatible multiplicity/composition conventions from entering its foundations.

CBirkbeck (commented):

**Assessment: the principal mathematical corrections from the previous review are present. Two focused specification amendments remain before I would recommend approval.**

The current head correctly restricts the spherical-to-ribbon converse, names the induced centre pivotal structure and gives its forgetful comparison, puts `IsAlgClosed k` on `fusionCoeff` and `grothendieckRing`, restricts adjoint generators to simple objects, and fixes the pivotal-functor composition order. I checked the corresponding actual signatures, not only the explanatory warnings. Those earlier objections should not be repeated against this version.

**1. Carry the splitting hypothesis into the Layer 3 multiplicity statement too.** Layer 3 still describes its setting as a k-linear abelian category and promises “the multiplicity of a simple S in X as `finrank k (S ⟶ X)`”. The correction made in Layer 4 and Suggested.lean needs to reach this earlier supplier. In real representations of C₃, the rotation simple V occurs once in itself but `dim_ℝ End(V)=2`. State that formula under Hom-finiteness and a split-simple hypothesis, or explicitly impose algebraic closure for that theorem. The nonsplit decomposition theorem itself may stay general; it uses dimension over the division algebra End(S), not raw k-dimension. The standing fusion bar should not have to be guessed for a layer explicitly presented as general semisimple-category infrastructure.

**2. Give centre semisimplicity an owned proof milestone.** The revised `ribbon_center_of_spherical` now assumes `IsFiniteSemisimpleCategory (Center C)`. That is a legitimate conditional theorem, but the announced unconditional centre-of-a-spherical-fusion-category result needs a producer for the instance. Layer 10 currently says only “with the centre's semisimplicity supplied”, and the ordering still lists just Layers 1–2 and the existing centre as its inputs. No specific supplier or construction of centre semisimplicity is identified there.

Please add the actual centre-fusion theorem and its prerequisites to the dependency plan, or point to an existing roadmap that owns it. In a proof through separability/averaging, spell out the induction/forgetful construction and the nonzero global-dimension input needed to split extensions; do not assume semisimplicity of the centre merely from semisimplicity of C. This is the extra work introduced by the correct restriction to the fusion setting. The ENO/Müger references already listed are appropriate, but the claimable construction needs an owner.

Nonblocking cleanup: the PR body and policy item 4 still describe a deliberate `RightRigidCategory` divergence, although the present classes and standing conventions use `RigidCategory`. Also remove the claim that a Prop-valued hypothesis cannot fix a coefficient field: a predicate can depend on k, and data may be obtained by choice from proved existence; the sound design reason for explicit instances is API clarity and avoiding hidden assumptions, not that logical impossibility.

The base dual/trace conventions and finite-support requirement for `VecTwisted` look coherent. I checked the current Mathlib rigid-directory contents as well; this is not a local Lean build. Resolve the present branch conflicts separately from these mathematical/specification points.

CBirkbeck (changes requested):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `2dce8fe44cd5604a04cb3ada959d296f5cc91181`.

I read both complete files and the earlier reviews. The earlier fixes should remain accepted: the spherical-to-ribbon converse has its semisimple scope; the centre's induced pivotal structure is named and compared on underlying objects; split multiplicities and simple-object adjoint generators are correct; and the pivotal-functor composite has the right order. The new centre layer assigns a previously missing prerequisite, but needs the following repairs.

**1. `globalDim_center` quantifies over an unrelated monoidal structure on `Center C`.** Its explicit `[MonoidalCategory (Center C)]` argument shadows the canonical centre tensor structure. All the later centre hypotheses are then about that arbitrary structure. The dimension-squaring theorem is not true at this generality. For example, over ℂ take C = Vec_{C₂}. Its centre has four simple objects. As a linear semisimple category it is equivalent to Rep(D₁₀), where D₁₀ has order ten and four irreducibles of dimensions 1,1,2,2. Transport that symmetric fusion tensor structure onto the same underlying centre category. The displayed hypotheses hold, but its global dimension is 10, whereas `(dim C)^2 = 4`.

Remove the free centre-monoidal instance and use Mathlib's canonical centre structure. Likewise use the canonical induced linear/additive structure, or supply explicit compatibility with the forgetful functor, wherever the theorem compares it with C. This is a target-signature error, not an objection to the true dimension formula.

**2. Supply the actual global-dimension input rather than citing Perron–Frobenius positivity.** Layer 5 proves positivity of FP dimensions. It does not show that the pivotal products `d(S)d(Sᘁ)` are positive, nor identify them with FP dimensions; that identification would incorrectly make every fusion category pseudo-unitary. Add the squared-norm/pivotal-dimension comparison and nonvanishing/positivity result, with its needed proof infrastructure. ENO, *On fusion categories*, Theorem 2.3, Proposition 2.9 and §3 are the relevant statements, distinct from the FP character construction. For the existing pivotal scope a shorter proof may be chosen, but its common-eigenvector/conjugation argument still needs to be stated. For an arbitrary algebraically closed characteristic-zero field, also specify the finite-data descent/base-change step rather than calling field elements 'positive reals'.

This also affects Layer 6's pseudo-unitarity target: `globalDim` currently requires a chosen pivotal structure, so it cannot by itself be the premise of the general theorem constructing a pivotal structure. Define the pivotal-independent squared-norm dimension first, or explicitly narrow the target to an already-pivotal input. State the positive-dimensional structure's uniqueness as well as existence (ENO Proposition 8.23), and specify the coefficient-field comparison used in `dim C = FPdim C`.

**3. Make the centre foundation and averaging identity explicit.** Before talking about exact adjoints and split short exact sequences, assign the abelian/linear structure on the canonical centre and the exactness of its forgetful functor. The two adjunctions alone do not identify their unit/counit composite with `dim C · id`; that graphical/algebraic calculation is the required normalization for the proposed retraction and should be a named sub-target. Also fix the Suggested.lean comment's composition: the endofunctor of Z(C) is `Center.forget ⋙ centerInduction`, not the reverse.

No local Lean build was run. These are mathematical and interface issues in the new layer, not requests to prove the roadmap's `sorry` targets before merge.

PR #58, opened 2026-07-06
feat: add roadmap for Temperley-Lieb diagrams, categories, and algebras

kim-em:
This PR adds a roadmap for Temperley-Lieb diagrams, categories, and algebras: planar matchings with their encodings, counting, and dihedral operations; the diagram category (circles as data) with its monoidal and rigid structure and its universal property as a presented monoidal category; the linear category TL(R, δ) with pivotal, spherical, braided, and ribbon instances; the algebras with the e_i presentation theorem, the unnormalized Markov trace, and conditional expectations; cell modules with the Gram determinant formula and the semisimplicity/nondegeneracy classifications; Jones-Wenzl projections with both recursions and the coefficient formula; and the Karoubi envelope with the generic Clebsch-Gordan classification and the root-of-unity quotients, stated as the first fusion categories in a proof assistant. Conventions pin δ = q + q⁻¹, with quantum integers as evaluations of Mathlib's rescaled Chebyshev polynomials, so genericity hypotheses are ring-level statements about δ; they also pin the two composites `cup ≫ cap = δ · 1` and `cap ≫ cup = e₁`, and the multiplication order `TLAlg` inherits from Mathlib's `End`, since both are easy to get backwards in a way that still typechecks. Suggested.lean elaborates against pinned Mathlib (sorry-stubbed milestones, allowed in the roadmap library).

Reusable infrastructure kept here, in three pieces: monoidal categories presented by generators and relations; cellular algebras; and the monoidal, linear and rigid structure of the additive-and-idempotent envelope together with quotients by tensor ideals. That last piece is not optional scaffolding. Mathlib supplies `Mat_` and `Karoubi` as plain categories only, with no monoidal and no `Linear` instance for either, and it has no monoidal quotient at all, so every claim Layer 7 makes about tensor products, duals or quantum dimensions in the envelope rests on material that has to be built first. Tensor ideals are accordingly stated for a general monoidal linear category rather than for TL alone, so the negligible ideal can be pushed along `Mat_` and `Karoubi` and quotiented out; the same closure conditions impose the relations on a presented monoidal category in Layer 2.

Field hypotheses are carried where they change the mathematics rather than assumed away. The root-of-unity dictionary needs ζ ≠ ±1 and characteristic zero, since [n](2) = n and [n](-2) = (-1)^(n-1)·n never vanish in characteristic zero and both vanish at n = p. The semisimplicity classification is stated outright — semisimple iff the quantum order exceeds n, or δ = 0 with n odd — and marked characteristic zero, because det G_{5,1} = [3]⁴(δ² - 2) is -2 at δ = 0 but vanishes in characteristic two; TL_5 over the algebraic closure of F_2 is an acceptance test alongside TL_2(0) and TL_3(0). Goodman-Wenzl carries the same hypothesis. The Gram determinant identity is stated universally over ℤ[X] and specialized, because cross-multiplied and then evaluated it says nothing whenever a denominator vanishes, which is exactly the root-of-unity regime it is wanted for.

The general pivotal/spherical/ribbon API is deliberately not owned here: this roadmap consumes it from the pivotal and spherical categories roadmap ([feat: add pivotal and spherical categories roadmap, #57](https://github.com/TauCetiProject/TauCetiRoadmap/pull/57)) and repays it with its first diagrammatic instances, so the cross-links in this README resolve once that PR merges; the README says so rather than presenting that API as already available. The fusion hypotheses at the Layer 7 summit are #57's verbatim, including `End 𝟙 ≅ K` and an algebraically closed field of characteristic zero, so its results apply to these categories. Braid groups, the Kauffman bracket, and the Jones polynomial are non-goals, as is the modular theory: the δ-level statements built here are what a characteristic-p roadmap would start from.

🤖 Prepared with Claude Code


kim-em:
`TauCetiRoadmap/TemperleyLieb/README.md:128`: "`qInt` is a one-definition evaluation layer over it (see the conventions), and the underlying polynomial identities are **natural candidates for upstreaming into that file**."

Against **never push work to Mathlib**, added in https://github.com/TauCetiProject/TauCetiRoadmap/pull/165 (doc: state the Mathlib relationship and the missing roadmap principles) after this PR opened; the merged roadmaps got the same sweep in https://github.com/TauCetiProject/TauCetiRoadmap/pull/169 (doc: conform roadmaps to "never push work to Mathlib").

The useful content is that the identities are about `Polynomial.Chebyshev.S` and not about Temperley-Lieb, so say that: they are statements about the Chebyshev polynomials, stated at that generality, with no diagram algebra in sight.

CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `1f7ec04704c30380fd3ab24bf469d201c422e924`, including the README and `Suggested.lean`. This is a mathematical/Lean-interface review; I have not run a local Lean build.

The universal Gram-polynomial formulation, characteristic-zero qualifications, and distinction between trace nondegeneracy and algebra semisimplicity are well motivated. The following issues remain.

### 1. Counting circles without their placement imposes more than the two snake relations

**Locations:** the definition of `TLDiagram`, and README Layer 2's universal property into an arbitrary monoidal category.

The diagram carrier stores a matching and just a natural-number circle count. Therefore a circle to the left of the identity strand and a circle to its right are equal morphisms `1 → 1`. In a target interpretation, writing `d = cup ≫ cap : 𝟙 → 𝟙`, this forces the left and right scalar actions of `d` on the generating object to coincide.

The two snake equations alone do **not** imply that equality in an arbitrary monoidal target.

An explicit test is the monoidal category of bimodules over `A = ℚ × ℚ × ℚ`, with tensor product over `A`. Let `M` have one one-dimensional summand `M_ij` for each directed edge `12, 21, 23, 32` of the three-vertex path. Define evaluation to send a reverse-edge pair `m_ij ⊗ m_ji` to the idempotent `e_i` and to kill nonclosed paths. Define coevaluation by

`e_i ↦ Σ_(j adjacent to i) m_ij ⊗ m_ji`.

Both snake equations hold: in either composite on `m_ij`, only the matching reverse edge survives. However, `d` is multiplication on `A` by `(1,2,1)`. On `M_12`, its left scalar action is multiplication by `1`, while its right scalar action is multiplication by `2`. The required functor from the proposed `TLDiagCat` therefore cannot exist.

**Repair:** distinguish the free category on a self-duality from its quotient which forgets circle placement. Either retain the extra planar region/circle data in the genuinely free carrier, or include the additional relations and their target-side compatibility in the universal property. A separate `R`-linear universal property with `cap ∘ cup = δ · id` and a monoidally `R`-linear target is a useful alternative for the later linear category. Do not claim the current count-only category is presented by exactly the two snakes into unrestricted monoidal targets. Pin the corrected presentation as an actual representative Lean signature and include the bimodule example as a negative test.

### 2. The unconditional idempotent-completeness instance for a quotient is false

**Location:** `Envelope`, immediately after `toMonoidalQuotient`:

```
instance (I : MonoidalIdeal S C) [IsIdempotentComplete C] :
    IsIdempotentComplete (MonoidalQuotient S C I) := sorry
```

The README correctly allows taking an idempotent completion after quotienting, but this signature instead asserts that every such quotient is already idempotent complete.

Take `C` to be finite-rank free abelian groups, an idempotent-complete symmetric monoidal `ℤ`-linear category, and let `I` consist of morphisms divisible by `6`. Its quotient has the same objects and matrices over `ℤ/6` as morphisms. The endomorphism `3` of the rank-one object is idempotent, since `3² = 3 mod 6`. It does not split in this quotient: under `ℤ/6 ≅ 𝔽₂ × 𝔽₃`, its image has ranks `(1,0)`, whereas every available object has ranks `(n,n)`.

**Repair:** remove this instance at its present generality. Use `Karoubi (MonoidalQuotient S C I)`, or prove idempotent lifting/splitting under explicit additional hypotheses in the particular TL quotient. The generic infrastructure must not assume the property it is intended to restore.

### 3. Pin the cell-form orientation and the horizontal reflection's tensor variance

**Locations:** README Layer 5's half-diagram basis/form, and Layer 3's reflection paragraph.

Layer 5 calls its half-diagrams **injective** maps `k → n`, but defines the form by `ū ≫ v = scalar · 1_k + ...`. With those sources and targets, `ū ≫ v` is an endomorphism of `n`, not of `k`. With an injective basis the composable pairing at `k` is `u ≫ v̄`. Alternatively, use the surjective `n → k` convention already used in the definition of `halfCount`, and state the corresponding module action consistently. The literal matrix/action/form comparison should be pinned before deriving the Gram formulas.

Also, the left-right (horizontal) reflection reverses juxtaposition:

`reflect(f ⊗ g) = reflect(g) ⊗ reflect(f)`.

It is covariant but **tensor-reversing**, not an ordinary monoidal involution of the category as the paragraph currently suggests. Package the target using the monoidal opposite, or state the anti-monoidal comparison explicitly. The distinction is already visible in the two different diagrams `cap ⊗ id` and `id ⊗ cap : 3 → 1`.

### 4. Specify the actual coefficient formula and its sign conversion

**Location:** README Layer 6's claim of a coefficient formula “as a product of quantum-integer ratios”, attributed to Morrison.

[Morrison, Proposition 5.1](https://arxiv.org/pdf/1503.00384) gives a **sum over admissible removal sequences** of products, divided by a quantum factorial. Its precursor is the sum in Proposition 4.1. Thus the proposed computational target needs the finite set of admissible sequences, their reindexing, and a sum; it should not silently replace that formula by one product read from the diagram. A different product formula would need to be identified and justified separately.

The paper also uses loop value `−[2]`, whereas this roadmap uses `+[2]`; its Wenzl recursion consequently has the opposite displayed sign. Record the translation of diagram generators/coefficient signs when importing the algorithm. The existing `f₂` and `f₃` checks are useful, and a diagram with two different admissible removal choices should be added to test that the contributions are actually summed.

### 5. A square root is needed for the specified braiding, not for the basic rank-two evaluation functor

**Location:** README Layer 3's evaluation representation parenthesis claiming that the square root supplied by `A` is necessary for cup and cap with circle value `q + q⁻¹`.

For any unit `q ∈ R`, take the cap matrix on `R²` to be

`B = [[0,q],[1,0]]`

and the cup coefficient matrix to be

`B⁻¹ = [[0,1],[q⁻¹,0]]`.

The two snakes are `B B⁻¹ = B⁻¹ B = I`, and the closed circle is `Σ_ij (B⁻¹)_ij B_ij = q + q⁻¹`. No square root is required. It is reasonable to retain an `A`-specialized representation for compatibility with the chosen Kauffman braiding, but the nonexistence claim is incorrect; separate this stronger braided parameterization from the more general monoidal evaluation construction.

The two first findings are foundational blockers. The later points fix concrete source/target and formula contracts rather than adding unrelated scope.

CBirkbeck (commented):

**Assessment: the earlier foundational repairs are substantially complete, but the newly added circle-centrality predicate needs one quantifier correction before approval.**

I checked the revised README and representative interfaces. The false quotient-idempotent-completeness instance is gone; the cell-form convention is now consistently based on string-surjections; horizontal reflection is tensor-reversing; Morrison's coefficient formula now includes its sum and sign translation; and the rank-two evaluation no longer claims to require a square root. The universal Gram-polynomial identity, characteristic-zero semisimplicity qualification, `TL₃(0)`/`TL₅(𝔽₂,0)` tests, and signed Chebyshev indexing in the actual `qInt` definition remain sensible.

### Restrict `IsCircleCentral` to the generating object

The README correctly requires the two actions of `d=η≫ε` to agree on **V**. However, the new Lean predicate says

```
∀ X : C, rightAction_X(d) = leftAction_X(d).
```

That is stronger and is not a necessary condition for interpreting `TLDiagCat` in an arbitrary target C. The image consists of the tensor powers of V and the unit; it imposes no condition on unrelated objects of C.

A small test separates the two. In the monoidal category of A–A-bimodules for `A=ℚ×ℚ`, take `V=Ae₁`, coevaluation `A→V⊗_A V` the projection onto e₁, and evaluation the inclusion into A. The two snakes hold, and `d=e₁` acts identically on both sides of V. This gives the intended count-only diagram interpretation: the two loop positions agree on V and hence on its tensor powers. But on the bimodule `M₁₂`, supported by e₁ on the left and e₂ on the right, d acts by 1 on the left and 0 on the right. Thus the current `∀ X : C` predicate fails even though the interpretation exists.

**Repair:** remove the universal quantifier and state the equality at V. Derive it for tensor powers as part of the presentation proof. A stronger all-objects predicate could be kept under a different name as a sufficient hypothesis, but it cannot be the claimed characterization. In the README's displayed equation also reverse the unitor maps to match the correctly typed Lean expression: start with `ρ_V⁻¹`/`λ_V⁻¹` and end with `ρ_V`/`λ_V` in `≫` notation.

Two smaller API/ownership improvements:

* Layer 7 should consume the semisimple-category predicate from #57, rather than say that defining it is part of this roadmap. TL owns the decomposition/classification proofs and envelope structure, not a second general semisimplicity class.
* With half-diagrams `n→k` and Mathlib's reversed End multiplication, plain precomposition gives the right cell action. For the displayed left `Module (TLAlg … n)` instance, pin the standard star-twisted precomposition formula (or choose the equivalent opposite-module convention). This will keep the corrected form orientation consistent with the action as well as with its scalar tower.

The remaining changes are focused and do not call for reopening the corrected Gram or Jones–Wenzl mathematics. No local Lean build was run. #57 remains an explicit dependency, and this branch also has merge conflicts to resolve.

CBirkbeck (changes requested):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `18fd1375a79bb634a880eff5378a909d03d577c0`.

The previous mathematical repairs are accepted: circle centrality is now required at the generating object rather than throughout the target; the false quotient-idempotent-completeness instance is removed; the surjective half-diagram convention and star-twisted left action agree with Mathlib's reversed End multiplication; horizontal reflection is tensor-reversing; and Morrison's coefficient formula is a sum with an explicit sign-conversion task. The rank-two evaluation construction no longer incorrectly requires a square root. I also checked the Gram formula against Ridout–Saint-Aubin, Theorem 4.7, and the root-of-unity ideal argument against Goodman–Wenzl.

**One concrete Lean-interface error remains:** `PresentedTL (R : Type*) ... : Type` places the presented algebra in universe zero even when R is in an arbitrary universe. But `presentedTLEquiv` identifies it with `TLAlg R δ n : Type u`. Already at n = 0, the intended algebra is R itself (the one-dimensional algebra on the empty diagram). Thus these targets silently demand a small representative of every coefficient ring, with no `Small` assumption. An opaque `sorry` definition hides this problem; elaboration alone does not repair it. Make the carrier `PresentedTL (R : Type u) ... : Type u`, consistently with `TLAlg` and `CellModule`, and check the presentation equivalence at an explicitly nonzero universe. No change to the mathematical scope is needed.

The README also explicitly promises representative signatures for the two universal properties, but the file still stops at `IsCircleCentral` and its propagation lemmas. Please add the interpretation/existence and uniqueness-up-to-monoidal-isomorphism targets, specifying the image of the generator, cup and cap. The corrected predicate is good, but it is not itself a universal-property statement. This is the remaining part of the earlier presentation-interface request, not a demand to prove the construction before merging.

Nonblocking proof-planning note: when implementing Goodman–Wenzl, expose the evaluable path/block-idempotent and constant-dimension inputs of its §2, or name an alternative proof. Its displayed Hom-to-End retractions divide by δ, so the included δ = 0 / ℓ = 2 case needs a separate argument rather than verbatim reuse. I am not asserting the δ = 0 conclusion is false.

This is close after the earlier substantive fixes, but I would make the carrier/universal-property corrections before approval. The dependency on #57 is clearly acknowledged. No local Lean build was run.

PR #61, opened 2026-07-06
fix(ContourIntegration): give condition (A′) the pole orders it needs

CBirkbeck:
### The problem: A′ is internally inconsistent

`ConditionAprime` is documented as requiring `γ` to approach each singularity
"as a finite union of model sectors **with a prescribed pole order**" — i.e. HW
condition (A): *flat of order n at a pole of order n*. But its signature is

```lean
def ConditionAprime (γ : ℝ → ℂ) (a b : ℝ) (S : Finset ℂ) : Prop := sorry
```

`S : Finset ℂ` is a bare set of **points** — it carries no orders, and there is
no `f` — so **no faithful definition can express the order-matching the
docstring (and the summit theorem) require.** The best a filler can do under
this signature is order-1 transversality, which is too weak for higher-order
poles, so `hungerbuhlerWasem_residueTheorem` (stated for general meromorphic
`f`) is unprovable from it. The definition and the theorem are inconsistent as
written.

This surfaced concretely while implementing `ConditionAprime` in TauCeti: the
correctness review [correctly] blocked an order-1 A′ as not faithful to HW (A′),
and every fix it suggested requires order data the signature can't hold.

### The fix (minimal)

Add `f` so A′ can read the pole orders (via `meromorphicOrderAt`), and thread it
through the summit's `hA`:

```lean
def ConditionAprime (γ : ℝ → ℂ) (a b : ℝ) (f : ℂ → ℂ) (S : Finset ℂ) : Prop := sorry
--                                          ^^^^^^^^^^^ added
...
    (hA : ConditionAprime γ a b f S) (hB : ConditionB γ a b f) :
--                              ^ added
```

This mirrors how condition (B) already couples with `f`, and is the smallest
change that lets a faithful A′ (*flat of order = pole order*) be stated — so the
general HW Thm 3.3 the roadmap already states becomes provable. No other targets
are touched. Elaborates green (`sorry` placeholders intact).

### DIFF
--- a/TauCetiRoadmap/ContourIntegration/Suggested.lean
+++ b/TauCetiRoadmap/ContourIntegration/Suggested.lean
@@ -100,10 +100,11 @@ def IsNullHomologous (γ : ℝ → ℂ) (a b : ℝ) (Ω : Set ℂ) : Prop :=
   ∀ w ∉ Ω, windingNumber γ a b w = 0
 
 /-- HW condition **(A′)**: the cycle `γ` approaches each on-cycle singularity in `S` transversally,
-meeting it as a finite union of model sectors with a prescribed pole order (HW §3). One of the two
-regularity conditions that make the principal value exist; stated explicitly so the summit is
-honest. -/
-def ConditionAprime (γ : ℝ → ℂ) (a b : ℝ) (S : Finset ℂ) : Prop := sorry
+meeting it as a finite union of model sectors, **flat of order equal to the order of `f`'s pole
+there** (HW §3). One of the two regularity conditions that make the principal value exist; stated
+explicitly so the summit is honest. The prescribed pole orders come from `f` (read off by
+`meromorphicOrderAt`), so `f` is part of the data: the point set `S` alone cannot supply them. -/
+def ConditionAprime (γ : ℝ → ℂ) (a b : ℝ) (f : ℂ → ℂ) (S : Finset ℂ) : Prop := sorry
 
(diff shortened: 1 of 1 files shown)


CBirkbeck:
So this it seems was a mistake/oversight in the roadmap that needs updating for the results to get past the AI reviewers.

mrdouglasny (approved):
Approving — correct, complete, and a genuine faithfulness fix on the summit's critical path.

**Correctness (the load-bearing point).** A′ is documented (and used by `hungerbuhlerWasem_residueTheorem`, stated for general meromorphic `f`) as requiring the cycle to be *flat of order = the pole order* at each on-cycle singularity. Under the old signature `(γ a b S)` with `S : Finset ℂ` a bare point set, that order data is simply unrepresentable — the best any filler could do is order-1 transversality, too weak for higher-order poles, so the summit was unprovable from it. Adding `f` so A′ can read `meromorphicOrderAt` is the minimal faithful fix, and it's genuinely HW-faithful (principal-value existence for an on-cycle pole depends on matching the contour's flatness to `ord_s(f)`).

**Completeness.** `ConditionAprime` occurs exactly twice in the file (def + the summit's `hA`); both are updated, no dangling references, and `f` is bound before `hA` in the summit signature so it's well-scoped. Build is green.

**API / docs.** Now mirrors the sibling `ConditionB γ a b f`, and the docstring both states the order-matching precisely and explains why `f` is data ("the point set `S` alone cannot supply them") — good design rationale. Scope is minimal and single-concern.

No change requests.

PR #47, opened 2026-06-28
Add roadmap: Modular forms — Hecke theory, newforms, and L-functions

CBirkbeck:
Adds a roadmap under `TauCetiRoadmap/ModularForms/` (README + `Suggested.lean`): the classical
arithmetic theory of modular forms on top of Mathlib's analytic foundation. It supersedes the
closed #36, rewritten to be self-contained and grounded all the way down.

**What it covers (Layers 0–11).** Diamond operators and modular forms with character (nebentypus)
→ the valence formula at general level → the Hecke algebra → the Petersson inner product →
newforms and strong multiplicity one → Atkin–Lehner/Fricke → L-functions (Euler product,
functional equation) → the **coefficient field is a number field** → the LMFDB invariants → the
modular curve as the analytic quotient `Γ\ℍ` and the **dimension formulas** — valence-formula
upper bounds, with the lower bounds by analytic Riemann–Roch **built inside Layer 10** (Forster's minimal chain: finiteness of `H¹`, Serre duality, Riemann–Hurwitz) → the level-one **Eichler–Selberg trace formula**. The modular curve is the compactified analytic quotient — no functor, no
representability, no moduli problem — and the cusp/elliptic-point counts and genus of `Γ\ℍ` are
built analytically within the roadmap.

**Notes**
- **Character spaces (Layer 0)** are defined exactly as in AINTLIB: the simultaneous
  `χ`-eigenspace of the slash-defined diamond operators inside `M_k(Γ₁(N))` (`modFormCharSpace`,
  a `Submodule`), with the classical transformation law as a bridge theorem
  (`modFormCharSpace_iff_nebentypus`) and the proved internal `DirectSum.IsInternal` nebentypus
  decomposition migrated as-is — **not** re-founded via a character-twisted slash action.
- **Hecke algebra (Layer 2)** consumes Mathlib's new abstract Hecke ring
  (`NumberTheory/HeckeRing/Defs.lean`, leanprover-community/mathlib4#41251; the convolution /
  ring-structure stack leanprover-community/mathlib4#41253–#41328, upstreamed from AINTLIB's
  `AbstractHeckeRing/*`) and adds the `GL_n` theory (Shimura §3.2: transpose commutativity, diagonal cosets, Theorem 3.20's polynomial structure, with the hand-off milestones the automorphic-representations roadmap consumes) and the `Γ₀(N)`/`Γ₁(N)` congruence pairs at `n = 2`,
[...]

PR #66, opened 2026-07-08
Add roadmap: graph regularity, finite weak regularity, and arity-3 hypergraph complexes

cameronfreer:
## Scope

Adds a self-contained finite regularity roadmap: graph regularity, finite Frieze–Kannan regularity, strong graph regularity, and an arity-three regular-approximation and induced-counting theorem. The regularity changes are in two files:

- `TauCetiRoadmap/Regularity/README.md` — the mathematical specification, layer dependencies, conventions, and completion criteria.
- `TauCetiRoadmap/Regularity/Suggested.lean` — suggested interfaces and definitions, importing only Mathlib. Theorem targets deliberately retain `sorry`.

Induced removal, arithmetic applications, and exchangeability are consumers, not endpoints. Finite–analytic comparison results are required separately; neither regularity proof depends on them.

## Mathematical interfaces

- Graphs and partitions use `SimpleGraph`, `Finpartition`, `IsEquipartition`, and `IsUniform`. The energy is size-weighted and includes diagonal blocks. Layer 1 states the exact weighted-square identity and the weighted absolute-deviation bound from Cauchy–Schwarz, without equitability or regularity assumptions.
- Layer 4 specifies a labeled induced-counting estimate for every graph on `Fin 3`, with error `(16τ + 3√ε)n³ + 6mn²`. Its coarse prediction is defined and transversal. The README defines the intermediate all-cell prediction and accounts separately for fine regularity, energy transfer, and repeated-cell contributions, with explicit induced-count examples.
- A triadic complex chooses its pair palette and has computed complexity. Its uncolored face levels and conversions to Mathlib's nonempty simplicial-complex vocabulary are specified. Pair and top densities have named bounds and normalization laws, with explicit support hypotheses.
- Layer 8 accepts an error schedule and a rank schedule `R : ℕ → ℕ`, then obtains a complexity bound. Top regularity uses `R C.complexity` at the returned complex.
- Layer 9 defines its rank schedule by a finite maximum over palette sizes. The adequacy lemma does not assume a regularity bound. A composed theorem returns the approximant, controlled cells, regularity, and the induced-count estimate together.
- The local prediction is an explicit product of the cell-respecting injection count, canonical pair densities, and top-color densities. Repeated cells use falling factorials. Global and discarded predictions are explicit sums of that formula. Sparse/dense counting and the six-charge assembly are separate targets.

The secondary provenance section compares these statements with `regularity-lemmata` at [315ef97](https://github.com/cameronfreer/regularity-lemmata/tree/315ef979f55f31cc43cd791302519d9a34cc2dc0). That reference is unchanged and is not a Lake dependency.

## Branch and validation

Regularity revisions through [b1372af](https://github.com/TauCetiProject/TauCetiRoadmap/commit/b1372af548a235dee88ecef54fbf83bb867e3daf). Upstream `main` at [aafd59c](https://github.com/TauCetiProject/TauCetiRoadmap/commit/aafd59c921100a7dc9fa20eb0d708c5ec1f508e8) has been merged, without rewriting the earlier branch history. This PR is stacked on [#374](https://github.com/TauCetiProject/TauCetiRoadmap/pull/374), the one-line, declaration-local AlgebraicTopology heartbeat fix; merge that first. The toolchain is [Lean 4.34.0-rc2](https://github.com/leanprover/lean4/releases/tag/v4.34.0-rc2), Mathlib [2b1308e](https://github.com/leanprover-community/mathlib4/commit/2b1308ee396f215169fdaa6807ea1d0f91d0001f), and Tau Ceti [be0b30f](https://github.com/TauCetiProject/TauCeti/commit/be0b30f3963c7f5158674dca96d7fe3fdd0dcd20).

- `lake build TauCetiRoadmap.Regularity.Suggested` passes at these pins. Roadmap `sorry` and linter warnings are not treated as errors.
- The full local `lake build` passes at [7ffda51](https://github.com/TauCetiProject/TauCetiRoadmap/commit/7ffda51c239fcceaee33c6f3e4e3b49a3b56f756), with #374 included.
- Relative to #374, the diff contains only the two regularity files; `git diff --check` passes.

Compilation checks the roadmap interfaces; it is not a claim that the remaining regularity and counting targets have been proved.

## Authorship

Authored by Cameron Freer with AI assistance. Earlier contributions used GPT 5.5 Pro and Sol 5.6 Pro, Claude Code (Opus 4.8 and Fable 5), and Codex (GPT 5.5 and 5.6 Sol), with lean4-skills and lean-lsp-mcp.

the most recent revision prepared by Codex with Astra xhigh and with advice from Astra Pro.


README.md
```diff
@@ -24,6 +24,7 @@ Tau Ceti is being incubated by the [Lean FRO](https://lean-lang.org/fro/) and th
 11. [Conformal mapping and the geometric theory of holomorphic functions](TauCetiRoadmap/ConformalMapping/README.md)
 12. [Weighted orthogonal L² bases: completeness, Hilbert bases, and products of orthogonal systems](TauCetiRoadmap/OrthogonalL2Bases/README.md)
 13. [Contour integration and the Hungerbühler–Wasem generalized residue theorem](TauCetiRoadmap/ContourIntegration/README.md)
+14. [Graph regularity, finite weak regularity, and arity-3 hypergraph complexes](TauCetiRoadmap/Regularity/README.md)
```
CBirkbeck: Rebase needed: this branch predates roadmaps 14 (Representation theory) and 15 (Modular forms) on `main` — the PR is currently conflicting, and this entry collides with both of them in addition to the acknowledged race with #53. After rebasing this becomes 16 (or 17 if #53 lands first).

TauCetiRoadmap/Regularity/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+## Prior formalization (secondary — reviewers judge the mathematics, not this map)
+
+[`cameronfreer/regularity-lemmata`](https://github.com/cameronfreer/regularity-lemmata) is a public
+Lean 4 library of finite regularity, counting, and approximation infrastructure — `sorry`-free with
+no custom axioms (CI-enforced by its `scripts/check.sh`). Its partition/graph layers are developed
+for **directed relations on an arbitrary `Finset` host** (subsuming `SimpleGraph`); its hypergraph
+development is Boolean and unordered. The declaration-level claims in this section and the
+per-layer notes were checked at commit
+[`f7403c9`](https://github.com/cameronfreer/regularity-lemmata/tree/f7403c9943fa47b83d071e6d622a87fe14122f79)
```
CBirkbeck: The pinned commit `f7403c9943fa47b83d071e6d622a87fe14122f79` does not exist on GitHub — this tree link 404s, so every declaration-level claim anchors to a dead link. The claims themselves look right at today's HEAD (`7c5a28e`): I spot-checked `frieze_kannan`, `energy_mono`, `exists_strongWitness`, `exists_regular_refinement_and_almostRefining_equipartition`, `exists_goodColoring`, `exists_triadic_regular_approximation`, `IsPolyadRegularAt`, and `cutDiscrepancy_le_iff`, plus the claimed file layout — all present. So this is a mis-pin (unpushed or rewritten history), not a content problem: please re-pin to a reachable SHA, ideally a tag.

TauCetiRoadmap/Regularity/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+## Interfaces exported to other roadmaps
+
+This roadmap exports finite regularity and counting interfaces that later roadmaps may consume —
+deterministic regularity inputs for exchangeable-array statements and removal-style / arithmetic
+hooks. These are **downstream consumers, not local endpoints**. In particular, once the
+exchangeable-array API exists, the finite sampling lemmas here should provide deterministic
+regularity inputs for random-array statements; this roadmap does not own the representation theorem.
+
+**Optional interoperability (not gating any layer).** The finite and analytic developments may later
```
CBirkbeck: The repository writing guide is categorical: “**Nothing is ‘optional’.** Don't use the word, and don't imply it. Everything on a roadmap is work we want … every item lives in *some* milestone.” This section's title plus eight further uses across the file (“optional downstream adapter”, “optional interoperability” ×5, …) run directly against that. Suggested fix: either give the adapters a milestone home — a late layer here, or make them cited targets of the dense-graph-limits roadmap — or move them squarely into *Non-goals*, and retitle this section accordingly (e.g. “Interoperability adapters (owned downstream)”).

TauCetiRoadmap/Regularity/README.md
```diff
@@ -0,0 +1,447 @@
...
+## References
+
+- E. Szemerédi, *Regular partitions of graphs* (1978).
+- A. Frieze, R. Kannan, *Quick approximation to matrices and applications*, Combinatorica 19 (1999).
+- L. Lovász, B. Szegedy, *Szemerédi's Lemma for the Analyst*, GAFA 17 (2007).
+- T. Tao, *Szemerédi's regularity lemma revisited*, Contrib. Discrete Math. 1 (2006) — the
+  strong-regularity (energy-gap stopping) iteration Layer 4 follows.
+- Y. Zhao, *Graph Theory and Additive Combinatorics* (2023), ch. 2 — graph regularity, counting, and
+  the strong-regularity exposition.
+- V. Rödl, M. Schacht, *Regular partitions of hypergraphs* (2007); B. Nagle, V. Rödl, M. Schacht,
```
CBirkbeck: Bibliographic: the exact title is Rödl–Schacht, “Regular Partitions of Hypergraphs: Regularity Lemmas”, *Combin. Probab. Comput.* **16**(6) (2007), 833–885. (The other references verified exact — including Terry arXiv:2404.02030's full title and the NRS RSA **28** (2006) data.)

TauCetiRoadmap/Regularity/README.md
```diff
@@ -0,0 +1,447 @@
...
+- **Gate.** The 3-uniform worked example computes a relative triple density over a triad and over a subpolyad.
+
+### Layer 7 — top-layer regularity over polyads
+- **Consume.** Layer 6.
+- **Build.** `IsTopRegularOverPolyad H P δ r` — the **`(δ, r)` NRS-rank form**: for every top color and every union of at most `r` subpolyads (`unionSupport` — arbitrary subgraphs of the parent pair graphs, *not* just vertex boxes) carrying a `δ`-fraction of the parent support, relative-density stability. `r = 1` is the disc-regular form; the rank the counting summit needs is pinned by pattern size in Layer 9 (`inducedCountingRank3`). The honest **weaker** predicate `IsVertexBoxRegularOverPolyad` (shrink the three cells, keep the full pair graphs) is kept as an intermediate target and for the `r = 2` shadow gate — it is **not** the predicate the induced-counting summit consumes: vertex-box discrepancy alone is generally not counting-ready strength. Plus the most-polyads / exceptional-mass and slicing/inheritance targets.
```
CBirkbeck: Minor: “disc-regular” is descriptive usage from the hypergraph-regularity literature — NRS themselves write `(δ, r)`-regular. A parenthetical would save an implementer a fruitless name hunt in the primary source.

TauCetiRoadmap/Regularity/Suggested.lean
```diff
@@ -0,0 +1,853 @@
...
+  regP : P.IsUniform G ε
+  regQ : Q.IsUniform G (F P.parts.card)
+  energyClose : weightedEnergy G Q - weightedEnergy G P ≤ ε
+  boundedFine : Q.parts.card ≤ strongGraphRegularityBound ε F l₀
+
+/-- **Layer 4 (summit).** Strong graph regularity, **compositional form**: against a starting
+equipartition `P₀` and a requested minimum complexity `l`, coarse/fine equipartitions with the
+`StrongRegular` properties exist, with the coarse partition **almost-refining** `P₀` (the chosen
+Layer-2 wrapper guarantees only almost-refinement of the *input* partition; the exact nesting
+`exists_regular_exact_refining_equipartition` provides holds between the partitions the iteration
```
CBirkbeck: Docstring garble: “…the exact nesting `exists_regular_exact_refining_equipartition` provides holds between the partitions…” — two phrasings merged; presumably “…provides the exact nesting between the partitions the iteration itself constructs”.

TauCetiRoadmap/Regularity/Suggested.lean
```diff
@@ -0,0 +1,853 @@
...
+the exceptional-mass parameter `εmass` and the schedule's counting error `εcount` are **separate**
+(the global theorem supplies the mass bound at `inducedCountingParameter3` but the schedule at the
+final counting error — a single shared `ε` could not be instantiated without an unpinned
+comparison); the slack charges the schedule at the **lower complexity**
+`#vertex-cells + #pair-colors` — the evaluation point `hlower` actually supplies, not
+`C.complexity`; and the slack carries the **route-count factor** `q₂^(k choose 2)` — per-route
+pair-counting errors accumulate over that many routes, so a bare `k²·schedule` would not account
+for them. Without this lemma the global assembly would bound only the actual side of the discarded
+routes. -/
+theorem exceptional_route_prediction_mass_le (H' : Colored3Graph κ₃ V)
```
CBirkbeck, line 1029: The one pinned *shape* in the file I'd flag before someone burns a week on it: the same function `inducedCountingSchedule3` is both the regularity hypothesis (`hlower` runs `LowerSkeletonRegular` at the schedule) and the conclusion slack (`q₂^(k choose 2) · schedule(ℓ)`) — which bakes in a **linear** modulus from pair regularity to predicted-mass control. Regularity-to-counting conversions often lose a power (`O(η^{1/4})`-type moduli), and with any fractional modulus this exact statement is unprovable for small schedules (`η^{1/4} ≥ η` there). I believe linear is genuinely achievable here — the derivation needed is tripartite counting against `η`-regular pairs, whose error is linear in `η` as in the classical triangle counting lemma — but the docstring should say that is the load-bearing reason, or else decouple the hypothesis strength from the conclusion slack (two schedules, or a modulus target), so the target is visibly provable rather than accidentally optimistic.

TauCetiRoadmap/Regularity/Suggested.lean
```diff
@@ -0,0 +1,853 @@
...
+`inducedCopyCount_edit_transfer`. The error is the **per-route budget at the `ε/6` placed-counting
+charge** `routeBudget3 C F₀.k (ε/6)` — not a bare `ε`, for two stacked reasons: a placement admits
+up to `q₂^(k choose 2)` routes (so per-route errors carry the route-count divisor), and the global
+`ε` splits into **six explicit `ε/6` charges** (placed counting; actual discarded mass; predicted
+discarded mass; predicted lower-route slack; diagonal; edit transfer) — allocating the full `ε`
+here would exhaust the budget before the other steps contribute. The regularity hypothesis runs at
+the genuine **schedule** `inducedCountingSchedule3` (local strengths), with the **global**
+edit/exceptional mass at the separate `inducedCountingParameter3`; the route is required
+top-regular at the schedule's value at the complexity. -/
+theorem placed_induced_counting3 (H H' : Colored3Graph κ₃ V) (C : TriadicComplex3 κ₃ V)
```
CBirkbeck: The six-charge decomposition carries no charge for polyads built from *very sparse* pair graphs, where relative-density stability over a near-empty support says little. I believe the statement survives as written because the error is at the absolute `routeBudget · ∏ᵢ |cellᵢ|` scale — if any pair density along a route falls below the budget, both the actual and the predicted count fall below it too, so sparse routes self-bound — but that case split is invisible in the file. Worth one docstring sentence here (or an explicit fold-sparse-into-exceptional convention in Layer 8) so implementers meet it by design rather than mid-proof.

CBirkbeck (commented):
*(Posted by Fable, Chris's AI assistant, at Chris's request — the verification described below is mine.)*

An adversarial pass over the roadmap: every checkable claim checked against the pinned Mathlib, the cited primary sources, and the prior-formalization repo, plus an independent adversarial review of the mathematical architecture (with a second-opinion model), and hand-computed counterexamples where claims were about counterexamples.

**What held up** — more than usually survives this treatment:

- The **Mathlib inventory is exact**: `szemeredi_regularity`'s quoted signature matches `Regularity/Lemma.lean` verbatim; `IsEquipartition`, `equitabilise`, `exists_equipartition_card_eq`, `Copy`/`IsContained`/`Free`/`copyCount`/`labelledCopyCount`, `triangleRemovalBound`, `SzemerediRegularity.bound` are all where claimed; no hypergraph type in Mathlib, as claimed.
- The **load-bearing energy claim is true**: Mathlib's `Finpartition.energy` (offDiag sum over `#parts²`) is genuinely not refinement-monotone. Concretely: `G = K_{X,Y}` on `Q = {X, Y}` has energy `1/2`; refining `X` into `m ≥ 2` parts drops it to `2m/(m+1)²` (I computed the `m = 2` case independently before seeing the general family). The `weightedEnergy` choice is vindicated.
- The **finite Frieze–Kannan bound shape** `4^(⌈1/ε²⌉+1)` is safe (each violating rectangle at most quadruples the parts and boosts weighted energy by `≥ ε²`; the `+1` is merely generous).
- The **NRS reading is right**: the rank-`r` test against unions of arbitrary lower-cylinder subgraphs matches NRS Definition 9, and vertex-box tests are *strictly* weaker (pair-label parity constructions look uniform on boxes with nonrandom configuration counts) — so keeping `IsVertexBoxRegularOverPolyad` out of the summit's diet is correct.
- The **one-orientation-per-pair route design** is sound as stated, precisely because no polyad/route/density ever consults the reverse orientation's color.
- The **six-`ε/6`-charge accounting is internally consistent**: I traced each charge to its pinned closing lemma, checked the route-budget summation recovers its `ε/6`, checked the three uses of `inducedCountingParameter3_charge` (edit, actual-discarded, predicted-mass), and checked that actual copies fiber exactly into (placement, route) pairs with `polyad_mem` automatic under `IsPolyadDecomposition` — no uncovered case at this resolution. The edit-transfer constant is safe (`C(k,3)·6·disc/6·n^k ≤ k³·disc·n^k`).
- The self-reported numbers are exact: **81** top-level declarations, **16** data + **17** theorem `sorry`s, zero `Prop := sorry`, build green.

The line comments carry the actionable findings: two mechanical must-fixes (a dangling prior-formalization pin; the index/rebase collision), one writing-guide violation ("optional" ×8), one bibliographic correction, and — the two comments I'd weigh most — a pinned target whose *shape* bakes in a linear regularity-to-counting modulus that deserves a stated justification, and a sparse-pair-block case the charge decomposition handles only implicitly.

One style note, not a rule: "summit" appears 21 times; recent review rounds on other roadmaps pushed toward plainer register for exactly this kind of vocabulary.

cameronfreer:
@CBirkbeck, thanks for the very thorough review! I think I've addressed it all (along with a bunch of other things that came across in the process). Please see the several new commits and updated PR description. Details below.

---

🤖 Summary written by Fable 5 and edited by GPT Sol 5.6

Every item is addressed on the branch, which is now rebased onto current `main`. One consolidated summary rather than per-thread replies:

- **Rebase / index.** Rebased; Representation Theory stays 14 and this entry now takes the next available index at merge time (15 as of today — Modular Forms #47 and Dense Graph Limits #53 are still open, so the number is assigned at merge, not predicted). Build re-run green after the rebase.
- **Provenance pin.** The pinned commit was indeed unreachable. Every declaration-level claim was re-checked at `a18f98c96770967bc9f3ab5a81fdf642d9f68b99` and the roadmap is re-pinned to that full SHA plus the tag `tauceti-roadmap-pin-1` (protected against update and deletion by a tag ruleset; the link uses `/tree/<tag>`, which resolves for a plain tag). The summary map gained a column stating each row's exact relationship to the targets here, so the prior-work evidence is inspectable per layer.
- **"Optional."** Gone from the README, `Suggested.lean`, and the PR description. The interoperability paragraph is retitled **Interoperability adapters (owned downstream)** with an explicit ownership boundary: a consumer roadmap that needs a comparison owns it as a named milestone there, and no layer or gate here depends on one.
- **Bibliography / terminology / docstring.** Rödl–Schacht corrected to the exact title and data (CPC **16**(6) 2007, 833–885), with the companion *Counting Lemmas* paper (887–901) distinguished; NRS page range added. "(δ, 1)-regular" parentheticals added at both first uses of "disc-regular". The merged Layer-4 docstring sentence is repaired as you suggested.
- **The two mathematical comments** — both taken, and resolved more strongly than a docstring note. Layer 9 now has a named **lower-route counting bridge**, applied on both the kept and the discarded side:
  - Concrete `lowerRouteCountAt` / `expectedLowerRouteCountAt`, and an explicit sparse/dense split (`IsSparseRoute`) in which sparse routes **self-bound** exactly as your comment argued — kept as an easy local case, deliberately not a seventh charge and not folded into the exceptional polyads.
  - An input-strength threshold `pairRouteRegularityThreshold3` **decouples** pair-regularity input strength from counting output error, so no target bakes in a linear regularity-to-counting modulus (linear and power-loss conversions both instantiate it). The dense counting theorem gives the threshold its mathematical meaning; positivity makes it admissible to the approximation theorem; calibration theorems ensure the schedule is strong enough to supply it at the route-budgeted ε/6 charge.
  - The discarded-route prediction bound (`exceptional_route_prediction_mass_le`) had the same schedule-in/schedule-out shape and is now decoupled too: input regularity at the threshold, conclusion slack the separate output function `exceptionalPredictionSlack3` with its own route-count-weighted charge — the schedule-shaped charge inequality is retired. The discarded predicted mass is restricted to **transversal placements** (the bridge's dense theorem requires transversality; the repeated-cell predicted side is owned by the diagonal gate, so the predicted-side partition stays overlap-free).
  - The dense conversion requires a genuine pattern/error-dependent rank, `requiredTopCountingRank3 … ≤ r` (never a bare `1 ≤ r`, which would assert `k`-vertex counting at rank 1). Its calibration against the global `inducedCountingRank3` is stated **for complexity-bounded complexes** — an explicit fixed-point target for this roadmap's constant-rank specialization, with the rank being calibrated appearing inside the complexity bound itself; Rödl–Schacht instead permit a rank schedule (`r : ℕ → ℕ`), which remains the documented fallback if no fixed point exists.
  - `placed_induced_counting3` now takes only its local contract (transversality, lower-skeleton regularity at the schedule, route-local top regularity, and rank adequacy — extracted from `hreg`'s complexity bound through the rank calibration) instead of the full approximation bundle, and the bookkeeping you verified by hand is exposed as targets: the fibration identity `inducedCopyCount_eq_sum_placed`, its predicted mirror, and a **proved** six-charge arithmetic assembly lemma.
- **Style.** "Summit" is down to one register-setting use in the README opening; elsewhere the text names the thing meant (the regular-approximation theorem, the counting theorem, a layer endpoint).
- **One update since your review:** Mathlib master has meanwhile gained a set-based `Hypergraph` carrier (`Mathlib/Combinatorics/Hypergraph/Basic.lean`, absent at the Tau Ceti pin `9caeba1`), so the "no hypergraph type in Mathlib" claim is now pin-sensitive, and Layer 0 owes an explicit `toHypergraph` migration boundary once the pin catches up.

kim-em:
@cameronfreer, would you mind going through the open conversations above and marking them as "resolved" if appropriate? This makes review easier. (This holds everywhere, not just this PR, not just this repo!)

TauCetiRoadmap/Regularity/Suggested.lean
cameronfreer: Resolved by decoupling — the stronger of the two remedies you offered, with the docstring rationale kept as well. The schedule no longer plays both roles:
- Input strength is now the named threshold pairRouteRegularityThreshold3 k ρ δ: the dense theorems (lowerRoute_counting3, placed_induced_counting3_of_denseRoute) take pair regularity at the threshold, and IsPairColorRegular.mono + the calibration theorems are the glue by which LowerSkeletonRegular at the schedule supplies it — so the schedule feeds the hypothesis side only.
- Output slack is the separate parameter δ at the placement scale on the kept side, and the separate function exceptionalPredictionSlack3 on the discarded side; the schedule-shaped conclusion charge is retired.
- The docstring states exactly the load-bearing point you asked for: "a linear conversion (≈ δ/C(k), as in the classical triangle counting lemma against regular pairs) and a power-loss conversion (≈ (δ/C(k))⁴) both instantiate this definition, so the targets do not silently bet on the linear modulus (explicit value is a target)." If the linear modulus turns out achievable, the threshold instantiates linearly; if a power is lost, the same statements remain provable.

🤖 written by Fable 5, directed by Cameron

cameronfreer:
@kim-em, sure happy to do so. I've resolved them all above here, as they were addressed in a single big reply. Sorry about that (in some other projects I've seen workflows where the reviewer resolved them only when verifying that they were addressed, and I didn't want to preempt that if it was the workflow here).

kim-em:
I opened a stacked editorial PR against this branch: https://github.com/cameronfreer/TauCetiRoadmap/pull/1

It reorganizes Layers 4–9, turns Layer 9 into four short sublayers with a six-charge table, removes review-history and rejected-design narration, trims the duplicated `Suggested.lean` docstrings, and condenses the reviewer checklist. The Lean declarations and proof bodies are unchanged.

cameronfreer:
@kim-em wrote:
> I opened a stacked editorial PR against this branch: https://github.com/cameronfreer/TauCetiRoadmap/pull/1
>
> It reorganizes Layers 4–9, turns Layer 9 into four short sublayers with a six-charge table, removes review-history and rejected-design narration, trims the duplicated Suggested.lean docstrings, and condenses the reviewer checklist. The Lean declarations and proof bodies are unchanged.

Thanks for the commits. I'll incorporate that in a moment along with a bunch of other changes (some based on recent work on cameronfreer/regularity-lemmata that derisks more of this PR), described below.

----

🤖 Claude says:

Structural revision, replayed on top of cameronfreer/TauCetiRoadmap#1. Summary of what changed and, as importantly, what deliberately did not.

**Layer 5 — pair-color signature.** `PairColorSystem` now carries an involutive `rev` and the coherence law `color_rev : color (reversePair p) = rev (color p)`. This was a genuine inconsistency, not a tidiness issue: Layer 6 read coordinate pairs at *both* orientations while Layer 9 budgeted one color per *unordered* pattern pair. Agreement on canonically oriented pairs does not determine polyad support — and it still fails when each system is merely swap-coherent for some involution of its own, so the involution has to be shared and therefore belongs in the signature. With it, canonical routes carry one color per pattern pair and the `ℓ ^ (k.choose 2)` factor in `routeBudget3` becomes honest arithmetic. `PairColorSystem.ofRaw` records that arbitrary raw directed data still embeds via joint colors — expressiveness preserved, palette cardinality squared (`ℓ ↦ ℓ²`), which is a real cost since the palette enters both the complexity measure and the route budget.

**Layer 6 — de-bundling.** `Polyad3.support` and `Subpolyad3.support` were proof-carrying fields pinned by `mem_support_iff`; they are now computed definitions. Dot notation is unchanged for every consumer. This yields data extensionality, `DecidableEq`, `Fintype`, and the enumeration bound `#parts³ · ℓ³` — the last three being prerequisites of `TriadicComplex3.polyads : Finset (Polyad3 skeleton)` that simply were not available before. `Subpolyad3.ofSubcells` is no longer a target: it is a definition with its intended characterization proved. The repository adapter's key dictionary is named explicitly as `faceKey = ![k₁₂, k₀₂, k₀₁]`, since `polyadBlock` keys a face by the coordinate it *omits* while `Polyad3` keys a *role pair* — exactly the kind of silent mismatch that would otherwise survive until a counting theorem failed.

**Layer 2 — reclassified.** Not a consumption of `szemeredi_regularity` but a build over Mathlib's exported iteration. The summit takes no seed, and the seed cannot be reintroduced afterwards because uniformity is not hereditary (finite counterexample: a regular pair with an irregular refined sub-pair), while common refinement also destroys equitability. Mathlib's increment does exactly refine its input and supplies the analytic content. Two pieces are built here over it: an initial equitable partition almost-refining the seed, and seeded execution of the energy induction. **[updated 2026-08-13]** Both are now stated as Tau Ceti constructions; the roadmap no longer speculates about upstream supplying either.

**Layer 9.** Routes are indexed by canonical orientation, so the `ℓ ^ (k.choose 2)` divisor in `routeBudget3` bounds one color assignment per unordered pattern pair; `faceKey` names the index-reversed adapter dictionary.

## Deliberately unchanged

- **The pair-regularity predicate stays coordinatewise.** The `L¹`-in-palette variant would let the route divisor go, but it rests on an open question: whether `L¹` regularity of two pairs controls the cherry/covariance sum by a palette-free modulus. Settled so far: `L¹` gives palette-free control of every *rectangle* test, and the generic cut-to-entrywise conversion costs exactly `ℓ`. Adopting half the trade would be unsound.
- **The fixed-rank calibration is not a refuted claim.** A cofinal falsification is vacuous — complexity is the computed sum `#cells + pairColorCount + #polyads`, so bounded complexity bounds the palette. The honest form is pointwise and reduces to `1 + L ≤ regularityBound3 …`. **[updated 2026-08-13]** The roadmap now commits to fixed rank: establishing `regularityBound3 … < 1 + L` is work to be discharged, not a gate to wait on, and the rank-schedule alternative is recorded in the consolidation comment below rather than in the document.
- **The sparse/dense split is retained.** The aggregate route identities settle the lower skeleton only, not the conversion from top regularity over a small polyad support.
- **The predicted repeated-placement contribution stays with the diagonal gate.** The upstream generic weight bound supplies its mathematical core but not TauCeti's expected-count definitions.

`lake build` remained green throughout. **[updated 2026-08-13]** The declaration and `sorry` figures given here are superseded by the consolidation comment below: the scan behind them missed headers carrying a `noncomputable` or attribute prefix.

**[updated 2026-08-13]** The fresh immutable pin (`tauceti-roadmap-pin-2`) and the consolidated inventory update are done; see the comment below.

cameronfreer:
I've done some further work downstream in [cameronfreer/regularity-lemmata](https://github.com/cameronfreer/regularity-lemmata), and repinned and revised this PR accordingly; I think it's now ready for re-review.

----

🤖 Claude says:

Consolidation pass. The prior-formalization claims are now pinned at [`315ef97`](https://github.com/cameronfreer/regularity-lemmata/tree/315ef979f55f31cc43cd791302519d9a34cc2dc0) / tag `tauceti-roadmap-pin-2` (protected by a tag ruleset, as pin-1 was), and the temporary post-pin paragraph is gone — those five results are folded into the capability map row by row.

**One substantive correction to the Layers 5–8 note.** The library's regularization is not arity-specific, so describing it as an "arity-3 precursor" understated it. It is stated for a coloring of `j`-sets against an arbitrary decidable observable on ordered `(j+1)`-tuples, normalized by `|V|^{j+1}`, with the triadic statements as the `j = 2` instance and the complexity recurrence `K ↦ K·2^{O(K^{j+1})}` (cameronfreer/regularity-lemmata#77, closing cameronfreer/regularity-lemmata#69). The qualifiers that do survive are the ones that matter here: a single Boolean observable rather than color-indexed relative densities, unordered pair colors, no vertex partition, and one disc atom rather than the bounded unions the rank-`r` tests quantify over.

A consequence worth flagging for the orientation work: permutation invariance of the observable is a hypothesis on the badness-closure lemma alone, so **orientation coherence is not forced by the regularization step**. It is forced from Layer 6 on, where polyad support reads coordinate pairs at both orientations — which is where `Suggested.lean` already locates it.

**Toolchain and `Hypergraph`.** Now Lean `v4.34.0-rc1` / Mathlib `de5ce8a`. `Mathlib/Combinatorics/Hypergraph/Basic.lean` is present at the pin, so the three "no carrier available" passages were stale. It is a set-based carrier (`vertexSet`, `edgeSet`, `Adj`, `EAdj`, `map`, `IsLoop`, `IsComplete`/`completeOn`, a bottom element) with **no uniformity predicate and no finiteness, density, or counting API**. The finite representation therefore stays public on the merits, and Layer 0 now states `UniformHypergraph.toHypergraph` with vertex-set and edge-set agreement as targets rather than as a contingency.

**Two research questions moved off the roadmap.** The former "Open design gates" section is now "Two pinned interfaces", stating the selected specifications definitely. The alternatives it used to hedge toward belong in discussion, not in a timeless document, so they are recorded here:

1. *An `L¹`-in-palette pair-regularity variant.* `∑_c |d_c(A',B') - d_c(A,B)| ≤ ε` would let the route-count divisor be dropped from `routeBudget3`, since per-route errors would be mass-weighted rather than uniform in cell volume. Adopting it turns on whether `L¹` regularity of `(A,B)` and `(A,C)` bounds the cherry covariance `∑_{c,c'} |Cov_a(β_a(c), γ_a(c'))|` by a modulus independent of the palette size `ℓ`. What is settled either way, and is why the roadmap keeps the coordinatewise predicate and the divisor together: `L¹` regularity gives palette-free control of every *rectangle* test of that covariance matrix, while the generic cut-to-entrywise conversion costs exactly `ℓ`.

2. *A rank-schedule formulation.* If no constant-rank fixed point exists, Rödl–Schacht permit a rank function evaluated at the complexity in place of a rank fixed in advance. The roadmap commits to fixed-rank, so the condition `regularityBound3 … < 1 + L` becomes work to be discharged rather than a gate to be waited on. This supersedes the "blocked gate / fallback" framing in the 11 August comment above, which has been marked accordingly.

**Inventory.** 139 top-level declarations: 58 targets (19 data, 37 theorem, 2 `Fintype` instances) and 81 complete (46 definitions, 19 proved lemmas, 12 structures, 4 instances). The sorry count is per-declaration from the compiler; `grep -c sorry` reads 60 because two docstrings discuss placeholders.

These figures supersede the 133 / 55 given on 11 August. Only three declarations were added in this pass — `toHypergraph` and its two agreement statements, which accounts for the whole `sorry` delta — and the rest of the gap is a correction: the earlier scan skipped declaration headers carrying a `noncomputable` or attribute prefix.

cameronfreer:
Refreshed this branch over current `main` after DenseGraphLimits and the subsequently merged roadmaps. The root index/import union and both issue-template dropdowns retain every roadmap. The cross-roadmap boundary is now explicit and consistent: Layers 0–9 prove finite regularity without analytic prerequisites, while this roadmap owns the finite-facing DenseGraphLimits comparison adapters as non-gating interfaces. The PR description has been synchronized, and `lake build` completes successfully (8,781 jobs). Ready for re-review.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

CBirkbeck (changes requested):
🤖 **GPT-5.6 Pro review, prepared at @CBirkbeck's request.**

I reviewed the current head `f74bccc634cb9699f4ea222df594cb284f55fb44`: the full README and `Suggested.lean`, the resolved review threads, the cited `regularity-lemmata` snapshot, the current roadmap rules, and the Rödl–Schacht/Nagle–Rödl–Schacht regularity and counting statements.

**Request changes.** The previous round has been addressed very well. In particular, the finite Frieze–Kannan layer, weighted-energy choice, subpolyad test surface, explicit approximant, sparse/dense split, separated input/output moduli, and six-charge assembly are now carefully designed. I did not find a counterexample to the concrete compiled theorem signatures I checked. I do, however, see three specification-level gaps and one branch-state blocker that prevent approval.

## 1. The fixed-rank calibration has an ungrounded—and probably wrong—quantifier order

The current architecture chooses one host-independent number

```lean
inducedCountingRank3 q₃ k ε
```

before the output complex is known. That same number is fed into

```lean
regularityBound3 q₃ η F r t₀
```

and must later dominate

```lean
requiredTopCountingRank3 k (routeBudget3 C k (ε / 12))
```

for every complex below that bound. But `routeBudget3` decreases with `C.pairColorCount`, while the only bound on `C.pairColorCount` comes from `regularityBound3`, which itself depends on the chosen rank. Abstractly, the calibration asks for a finite solution of a circular inequality of the form

```text
r ≥ requiredRank (ε / B(r)^choose(k,2)).
```

There is no general fixed-point principle giving such an `r`; making both sides monotone does not help, and the README's sentence that the functions are “chosen jointly” is not a proof route.

The cited sources use a different quantifier order for exactly this reason. Rödl–Schacht I, Theorem 2.3, takes a **function** `r : ℕ^(k-1) → ℕ` and returns top regularity of rank `r(a_P)`, evaluated at the output partition complexity. Rödl–Schacht II, Theorem 1.3, chooses the required rank only after the lower densities have been fixed.

The clean repair is to make Layer 8 accept a rank schedule, for example

```lean
R : ℕ → ℕ
```

and require top regularity at `R C.complexity` (or at a more precise lower-complexity tuple). Layer 9 can then define `R ℓ` as a maximum of the required ranks for all pair-palette sizes bounded by `ℓ`. This removes the circularity and follows the source theorem's quantifiers.

An alternative is to retain a fixed rank but add and prove an explicit finite fixed-point theorem for the actual `regularityBound3` and `requiredTopCountingRank3`. That would be genuinely new mathematics and needs a stated proof route; it cannot be left inside an opaque calibration target.

Until this is repaired, `exists_strong_regular_approximation3` and the final counting theorem have not been shown to compose.

## 2. Layer 4's counting deliverable is not a theorem statement

Layer 4 currently asks for

> “a counting lemma whose hypotheses include `StrongRegular`”

and its gate is

> “Prove at least one counting result from `StrongRegular`.”

This is not unambiguous enough to guide implementation or review. A trivial lemma which ignores the witness would satisfy the words. It also leaves unclear what quantity the coarse partition is meant to predict, which patterns are covered, and how the fine regularity and energy gap enter the error.

Please pin one exact counting theorem and its prediction. The existing binary-palette result in `regularity-lemmata/Relational/GraphCounting.lean` gives a useful model: for a fixed three-vertex induced pattern it compares the actual induced embedding count with a named coarse estimate, with an explicit error depending on the fine regularity, an auxiliary density threshold, the energy gap, and the diagonal-cell term.

The Tau Ceti target need not copy that exact constant, but it should state something of that specificity in both the README and `Suggested.lean`, including:

- the pattern carrier;
- the coarse predicted count;
- labeled versus unlabeled and induced versus non-induced normalization;
- the exact hypotheses consumed from `StrongRegular`;
- an explicit error bound.

This is also the acceptance test that shows the proposed `StrongRegular` object has the right API rather than merely existing.

## 3. `expectedInducedCountAt` is still not pinned by an exact formula

The central prediction is currently an opaque `sorry` definition whose docstring says only:

> “the injection-corrected cell-size factor, times the pair-color densities … times the required relative top-color densities.”

That does not determine the object. In particular, “injection-corrected cell-size factor” is ambiguous when several pattern vertices are assigned to the same cell, and the precise top-color factor and its indexing are not stated anywhere as an equation.

Please define, or characterize by a theorem, the exact local prediction. A clean interface would first define

```lean
placementInjectionCount φ
```

as the number of injective maps respecting the assigned cells, prove that it is the product of falling factorials over repeated cells, and prove that for a transversal placement it is `∏ i, |φ.vertexCell i|`. Then state literally

```text
expectedInducedCountAt
  = placementInjectionCount
    × ∏_{i<j} pairColorDensity(...)
    × ∏_{i<j<l} relativeDensity(H', patternColor(i,j,l), ψ.polyad(i,j,l)).
```

The exact equality is load-bearing for the sparse-route bound, the exceptional predicted-mass estimate, the diagonal estimate, and the global predicted-sum identity. A prose product shape is not enough to verify those later targets or to prevent an implementation from choosing an unrelated prediction.

## 4. The branch must be rebased and rebuilt before approval

This head is currently **248 commits behind `main`**. Its successful CI run used the old dependency graph (`Mathlib 05ae0103…`, `TauCeti e8af08d0…`), whereas current `main` uses much newer Mathlib and Tau Ceti revisions. The current green check therefore does not establish that `Suggested.lean` elaborates against the toolchain into which it would merge.

Please rebase onto current `main` and obtain a fresh green `lake build`. The PR description should then be refreshed as well: it still says four files are changed, including the root README and `TauCetiRoadmap.lean`, while the current diff correctly contains only the two files in `TauCetiRoadmap/Regularity/`.

## Further API improvements

These are smaller than the three blockers above, but I would address them in the same revision.

### Reconcile `HypergraphComplex` with Mathlib's simplicial-complex vocabulary

Mathlib already has `PreAbstractSimplicialComplex`, a downward-closed family of finite nonempty faces. The proposed `HypergraphComplex` differs by retaining the empty face, a dimension bound, and finite level sets, so it may deserve a separate bundled type; however, the roadmap currently neither reuses the Mathlib object nor gives a bridge explaining the difference.

Please either build the finite bounded complex as an augmentation/refinement of the Mathlib carrier, or provide explicit conversion and membership theorems. Likewise, `TriadicComplex3.toHypergraphComplex` should have characteristic equations for its 0-, 1-, 2-, and 3-faces rather than remain an unrelated opaque object.

`TriadicComplex3 κ₃ V` also has a phantom `κ₃`: none of its fields mention the top palette. It would be cleaner as `TriadicComplex3 V`, with `κ₃` appearing only in predicates involving a colored top graph, unless top-color data is intended to become part of the structure.

### Add the elementary density API explicitly

The counting proofs use, and the docstrings repeatedly appeal to, facts such as

```lean
0 ≤ pairColorDensity ...
pairColorDensity ... ≤ 1
∑ c, pairColorDensity ... c = 1
0 ≤ relDensityOn ...
relDensityOn ... ≤ 1
∑ c, relDensityOn ... c = 1
```

with the appropriate positive-denominator or nonempty-support hypotheses. These should be named Layer 5/6 API, together with the density transformation under `rev`. They are basic theory of the objects, not proof-local arithmetic.

### Add the composed existence-and-counting endpoint

After the rank issue is repaired, add a theorem which instantiates `exists_strong_regular_approximation3` with the counting parameter, schedule, rank schedule, and diagonal floor and returns `H'`, `C`, and the final induced-count estimate together. The current conditional counting theorem is useful, but the one-line composition is an important regression test that the two parameter interfaces actually match.

## What looks good

The earlier concerns about the linear pair-regularity modulus and sparse routes have been resolved properly rather than papered over. The roadmap now has a real lower-route bridge, separates input regularity from output slack, treats repeated-cell and exceptional-route contributions on both the actual and predicted sides, and exposes the fibration and arithmetic assembly. The weighted-energy and finite Frieze–Kannan layers remain well grounded in Mathlib and the pinned prior formalization. I also found no target or completion criterion that pushes work to Mathlib: the implementation is correctly owned by Tau Ceti, with Mathlib used only for existing vocabulary and substrate.

Once the rank quantifiers, Layer 4 gate, and local prediction formula are fixed and the branch is rebased, I would be happy to re-review this for approval.

cameronfreer:
:robot: prepared at Cameron's direction by Astra xhigh in codex with feedback from Astra Pro.

Response to [CBirkbeck's September 11 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/66#pullrequestreview-5179571135), with the regularity revisions through b1372af.

1. **Rank calibration.** bb963ca replaces the fixed rank with a schedule `R : ℕ → ℕ` in Layer 8, evaluated at the returned complex's complexity. Layer 9 defines `inducedCountingRankSchedule3` by a finite maximum over palette sizes. `requiredTopCountingRank3_le_inducedCountingRankSchedule3` has a proof using membership in that finite range, with no regularity-bound hypothesis. This supersedes the earlier comments defending a fixed-rank calibration. `exists_strong_regular_approximation3_counting` now has a composition proof connecting the existence and counting interfaces; its underlying mathematical ingredients remain roadmap targets.

2. **Layer 4 counting.** ce96172 defines the labeled induced-injection count for every graph on `Fin 3` and its transversal coarse prediction. b1372af states the bound

   ```text
   |inducedGraphCount3 J G - coarseInducedGraphEstimate3 J G P|
     ≤ (16τ + 3√ε)n³ + 6mn²,
   ```

   where `τ = F #P.parts` and every coarse cell has size at most `m`. The docstring identifies the four witness fields consumed. Layer 1 now states `weightedDensityDeviation_sq` and `weightedDensityDeviation_abs_le`: the exact weighted-square identity and its Cauchy–Schwarz consequence, including diagonal blocks and without equitability or regularity assumptions. The README defines the all-cell intermediate prediction and gives the transfer argument, including why actual minus predicted repeated-fine-cell contributions cost the repeated-cell volume only once. The threshold parameter is absent. The triangle, edgeless, and one-edge examples have explicit expected labeled counts.

3. **Prediction formula.** ce96172 defines `placementInjectionCount` as an actual count and `expectedInducedCountAt` as the specified product of that count, canonical pair densities, and top-color densities. Falling-factorial and transversal identities are named targets. 24a51cd defines the global and discarded predictions as explicit sums of this local formula. The repeated-cell acceptance example gives two injections for two labeled vertices assigned to a two-element cell, not four.

4. **Basic APIs.** 568b589 removes the phantom top palette from `TriadicComplex3`, supplies the Mathlib simplicial-complex conversions and face equations, and names the density bounds, normalization, and reversal laws. b1372af also corrects the orientation docstring: supports and routes use canonical pairs; reversal provides compatibility under exchanging roles. Layer 0 now distinguishes homomorphisms, injective non-induced copies, and induced injections. 407d994 keeps the prior-formalization comparison secondary and removes revision history from the roadmap prose.

Upstream main was merged in 458db8d. This PR is stacked on [#374](https://github.com/TauCetiProject/TauCetiRoadmap/pull/374), the one-line, declaration-local heartbeat fix; merge that first. The full local `lake build` passes with it included. Compilation checks the interfaces, not the proofs of the remaining `sorry` targets.

Ready for another review.

PR #77, opened 2026-07-19
Add complexity theory roadmap

BoltonBailey:
This is a roadmap for the formalization of basic computational complexity theory
based on Turing Machines that operate over binary strings.


felixpernegger:
@BoltonBailey FYI, @kesslermaximilian is working on this (in particular Cook Levin theorem) as his master thesis

BoltonBailey:
See discussion about Max's work here: [#CSLib > Plans for P / NP, Turing machines and Cook's theorem @ 💬](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Plans.20for.20P.20.2F.20NP.2C.20Turing.20machines.20and.20Cook.27s.20theorem/near/612609322)

Looks like he is taking a Lambda Calculus based approach, I think the Rocq version did that also. I have been thinking maybe another direction to take this would just to be to ask the AI to produce proofs of polytimeness or non-polytimeness for every Prelude definition that fits into the framework. On the other hand, I don't know how well that lines up with the rest of the projects being proposed on this repo. Also, https://github.com/SamuelSchlesinger/complexitylib was recently brought to my attention, and it seems like A lot of this could just be done there.

So I think I will consider this a please-adopt/something a maintainer can close if they think it's not a good fit.

PR #80, opened 2026-07-22
Add roadmap: adic spaces, up to the Fargues–Fontaine curve

CBirkbeck:
A roadmap for the foundations of Huber's theory of adic spaces (`TauCetiRoadmap/AdicSpaces/`, README + `Suggested.lean`): Huber and Tate rings, the valuation spectrum `Spv A`, continuous valuations and the adic spectrum `Spa (A, A⁺)`, rational localization and the structure presheaf, the sheafiness theorem with Čech acyclicity for strongly noetherian Tate rings (Wedhorn Theorem 8.28(b) / Huber), the definition of adic spaces, and — as the final application (Layer 7 of Layers 0–7) — the adic Fargues–Fontaine curve `𝒳 = 𝒴/φ^ℤ` for `E = ℚ_p` and a perfectoid field `F` of characteristic `p`.

**Notes**

- **Coordination with Mathlib.** Open PRs now develop this material — mathlib4#38009 (`Spv` and its topology on `ValuativeRel`), mathlib4#42312 (Huber rings), mathlib4#42315 (`Spa` as a topological space). The roadmap follows the upstream decisions: `Spa` is a subspace of `Spv`, points are valuations up to equivalence, and the Layer-0–2 names align with upstream at migration.
- **`A⁺` is explicit data.** Ring properties (`IsHuberRing`, `IsTateRing`, `IsStronglyNoetherian`, `IsUniform`) are `Prop` mixins; theorems take `Aplus : Subring A` explicitly; a bundled `HuberPair` (with morphisms carrying continuity and `f(A⁺) ⊆ B⁺`) serves the categorical uses. No global `A⁺` typeclass, and no globally bundled pair.
- **The categories are Wedhorn's ladder**: `𝒱^pre` (presheafed spaces with stalk valuations) → affinoid pre-adic spaces → pre-adic spaces → `𝒱` (the sheaf subcategory) → adic spaces. Sheafiness comes in three named notions (`IsSheafyPair`, `IsSheafyRing`, `IsStablySheafyRing`), and Layer 4 proves the pair-level sheafiness theorem plus the all-degrees Čech acyclicity actually used by the applications (the `H^q`-vanishing form of 8.28 needs a Čech-to-sheaf-cohomology comparison; sheaf cohomology is not developed here).
- **Layer 6** states the Buzzard–Verberkmoes theorem and a **new counterexample** (uniform, sheafy, not stably uniform — answering Hansen–Kedlaya Remark 3.16), proved in AINTLIB's `FJP/` directory; its formal status inherits the Layer-4 dependency cone like every downstream result.
- **Layer 7** fixes the direct wandering-slice route for the chart isomorphism in `𝒱^pre` (the one genuinely new theorem; the topological charts exist in provenance), lists every hypothesis needed to apply Layer 4 to Kedlaya's interval rings `B^I`, and states the power-comparability lemma the rank-free window covering rests on.
- **Out of scope**: étale cohomology, formal models/GAGA, Berkovich spaces, fibre products, almost/condensed methods, perfectoid spaces and tilting beyond Layer 7's perfectoid-field input, Huber's noetherian-ring-of-definition sheafiness case, and the curve's structure theory (bundles, Harder–Narasimhan, classification, `B_dR`, general `E`, the relative and schematic curves) — a successor roadmap.
- **Provenance.** AINTLIB `dev/adic-spaces @ 59bbbe8ba14a` (2026-07-28), with a per-milestone status table in the README separating direct `sorry` counts from the transitive `#print axioms` audit required at migration. The sheafiness capstone and `Spv` spectrality are the main incomplete parts; the quotient-curve topology and interval-ring strands are implemented but inherit that cone.


kim-em (commented):
:robot: Claude+Codex say:

This is a strong document, and several things in it are the best I've seen in a roadmap here: the
`A⁺`-is-data decision stated with its reasons, the Hausdorff convention gap called out explicitly,
the honest `sorry` policy in `Suggested.lean`, and the separation of direct `sorry` counts from the
transitive audit. The coordination with the three in-flight Mathlib PRs is real coordination rather
than a gesture, and Layers 3 and 4 name their proof steps at a granularity a contributor can pick up.

Most of the specifics are inline. This body carries the blocking summary, the citation checklist
(which spans too many places to anchor), a non-blocking note on shape, and a list of the claims we
checked and found good.

## The blocking list

1. **Wedhorn §7.1 (`Spv(A, I)`) is missing, and the stated route to spectrality of `Cont` is not
   available.** Remark 7.6 says the inclusion `Spv(A, I) ↪ Spv(A)` is not spectral in general.
   (Inline at Layer 1.)
2. **Rational localization for a general Huber ring needs Wedhorn's weighted restricted series
   (5.48) and topological localization (5.51), not the Tate-algebra quotient.** This is exactly the
   case Layer 7 needs, since `A_inf` is Huber and not Tate. (Inline at Layer 3.)
3. **Two prerequisites exist in neither Mathlib nor the roadmap:** boundedness of a subset of a
   topological ring, and pro-constructibility. (Inline.)
4. **Layer 4's five-step flatness route cannot work as written**, and it departs from the proof in
   the source it cites. (Inline.)
5. **The Hausdorff convention is stated and then not followed in Layers 2, 3 and 6**, and one
   resulting statement is false as it would be read in Lean. (Inline.)

## Wedhorn citations: incorrect, or correct with mismatched hypotheses

We checked every Wedhorn number against arXiv:1910.05934v1. Your claim that v1 is the only version
is correct. Each row is also flagged inline at its location.

| Roadmap says | Wedhorn actually has |
|---|---|
| `Spa (A, A⁺) = ∅ ↔ A = 0` for complete `A` (Proposition 7.32 shape) | 7.32 is a Corollary about a unit `π` with `\|π(y)\| < \|s(y)\|`. Wanted: **Proposition 7.49(1)**, for an arbitrary affinoid ring, `Spa A = ∅ ⟺ A/closure{0} = 0` |
| Iterated localization: a rational subset of a rational subset is rational (Lemma 7.54 = [Hu2] Lemma 2.6) | 7.54 is the **cover refinement** lemma, and Wedhorn cites **[Hu3]** Lemma 2.6 for it. Iterated localization is **Proposition 8.2(2)** |
| `A⁺ = {f : ∀ v, v f ≤ 1}` for **complete** pairs (Proposition 7.52 shape) | 7.52(1) holds for an **arbitrary** affinoid ring. Only 7.52(2), the unit criterion, needs completeness. Drop the hypothesis rather than adjust it |
| `IsUniform A` (`A°` bounded, **Definition 7.36**) | 7.36 is a Corollary about constructible subsets |
| `IsStablyUniform` (**Definition 7.37**) | 7.37 is an Example about `Spa(K, A(v))` for a microbial valuation |
| Cover normalization (Wedhorn **Lemma 8.31**/8.34 shape) | 8.31 is flatness of `A⟨X⟩`. Use **7.54** for refinement to a standard rational cover, **8.33** for the two-piece Laurent base case, **8.34** for reduction of general rational covers. Reserve **7.55** for Proposition 8.30 |
| Rings of definition are exactly the open, bounded, adically topologized subrings (**Corollary 6.4, Lemma 6.6**) | That is **Lemma 6.2**. 6.4 is intersections/products of rings of definition and `A°` as their union, which does support your neighbouring claims; **Lemma 6.6** is the open-ideal criterion, needed in Layer 2 |
| Completion ... ([Hu2] Lemma 1.6, **Wedhorn §6.4**) | §6.4 is "Examples". Completion, dense image, kernel `closure{0}` and the Hausdorff embedding are **§5.4, Proposition and Definition 5.32 with Example 5.33**; **Remark 6.8** is preservation of the f-adic structure; [Hu2] Lemma 1.6 is cited by Wedhorn only for **6.9(1)** |
| (Wedhorn states 8.28 for the completion) | 8.28(b) is stated for `A` a strongly noetherian Tate ring, with no completion. Definition 8.26 and 8.28(a) are the ones going through `Â` |

**Larger than a citation slip:** Wedhorn has **no notion of uniformity at all**. The word does not
occur outside "uniform space" and "uniformizing element". The definitions should be cited to
**[HK] Definition 2.3** (uniform: `A°` bounded) and **[HK] Definition 3.13** (stably uniform).

**Correct, for the record:** 6.1, 6.10, 6.17 and 6.18 (both marked "Proof. Missing"), 7.14,
7.30(2) and 7.30(3), 7.35, 8.10, 8.22, 8.26, 8.27, 8.28(b), 8.32, 8.33, 8.34. Add **Theorem 6.16**,
which is the open mapping statement itself and also marked "Proof. Missing", alongside 6.17-6.18.

## Non-blocking: shape

Layer 7 is about a quarter of the document for one application, while the general theory that the
rest of the subject reuses is comparatively thin. Whether to be deep or broad is your call, so treat
this as a suggestion. Things that are not targets and that we would expect a future roadmap in this
area to want:

- **Properties of morphisms.** You have `𝒱`-morphisms and open adic subspaces, but no finite type,
  separatedness, closed immersions, or a systematic open-immersion API.
- **Analytic points**, `(Spa A)^a` (Definition 7.39, §7.5, §8.3). Layer 7 correctly notes `𝒴` is not
  the analytic locus, but the notion is never defined; Proposition 7.49(2) and Remark 7.50 are the
  reason it matters.
- **Affinoid algebras over a nonarchimedean field** (Definition 7.56, Example 7.57 following):
  `k⟨X₁,…,Xₙ⟩` with `A⁺ = A°`, discs, points of the closed disc. Classical rigid geometry, and
  highly reusable.
- **Closed subspaces from quotient pairs** (Proposition 7.38).
- **The support map `Spv A → Spec A`** and its section by trivial valuations (Remark 4.6).
- **Fibre products** and the completed tensor product. We understand the exclusion, but the completed
  tensor product is basic reusable Huber-ring theory rather than geometry.

One naming note that doesn't anchor anywhere in particular: `IsUniform`, `IsStronglyNoetherian`,
`IsSheafyRing` and `IsRingOfIntegralElements` are very generic for the root namespace. There is no
actual clash (Mathlib's `IsUniform`s are all namespaced), but a `Huber` namespace would match
Mathlib convention.

## What we checked and found good

- Wedhorn arXiv:1910.05934 has exactly one version, as claimed. `Suggested.lean` compiles against
  the pinned Mathlib.
- The three Mathlib PRs exist and match your descriptions, and were open when checked on 2026-08-06:
  https://github.com/leanprover-community/mathlib4/pull/38009 "feat(RingTheory/Valuation): define
  the valuation spectrum and its topology",
  https://github.com/leanprover-community/mathlib4/pull/42312 "feat(Topology): huber (f-adic)
  rings", https://github.com/leanprover-community/mathlib4/pull/42315 "feat(Topology): the adic
  spectrum Spa(A, A+) as a topological space".
- Kedlaya arXiv:1410.5160's abstract does state strong noetherianity for extended Robba rings over a
  **perfect** nonarchimedean field of characteristic `p`, so you are right that algebraic closedness
  is not needed. `PowerSeries.IsRestricted` and `IsTopologicalAddGroup.rightUniformSpace` both exist
  in pinned Mathlib as cited.
- Power-comparability really does follow from continuity for the `(p, [ϖ])`-adic topology: taking
  `γ = v(p)^a` gives `n` with `v([ϖ])^n < v(p)^a`, and symmetrically. Good observation, and you are
  right that it is invisible at rank one.
- `κ ∘ φ_Spa = p·κ` is right for `φ_Spa(v) = v ∘ φ`; `𝒴 = D(p) ∩ D([ϖ])` against the analytic locus
  `D(p) ∪ D([ϖ])` is right; the windows `{v([ϖ])^b ≤ v(p)^a}` really are rational subsets of
  `Spa(A_inf, A_inf)`, since `(p^a, [ϖ]^b) ⊇ (p, [ϖ])^{a+b}` is open; the Tate characterisation of
  rational-subset openness is Remark 7.30(2), correctly cited; and the `{X, 1−X} ⊆ K⟨X⟩` example is
  correct, since `1/(1−X)` has coefficients not tending to zero.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  the specification here is its own document. Where existing Lean work proves a milestone,
+  that is provenance, not the standard it is judged against.
+
+## What Mathlib already has (consume)
+
+- **Valuative relations.** `ValuativeRel R`, the canonical valuation into
+  `ValueGroupWithZero R`, `Valuation.Compatible`, `ValuativeExtension`, the induced topology,
+  `DiscreteValuativeRel`.
+- **Nonarchimedean topological algebra.** `NonarchimedeanRing`, `OpenSubgroup`, adic
+  topologies and `IsAdic`, `IsTopologicallyNilpotent`, boundedness, topological
```
kim-em: **Blocking: Mathlib does not have this.**

`A°`, "rings of definition are exactly the open bounded adic subrings", and `IsUniform` all rest on
"a subset of a topological ring is bounded". `Bornology.IsVonNBounded` requires
`[SeminormedRing 𝕜] [SMul 𝕜 E]` and does not apply to a Huber ring; there is nothing else in pinned
Mathlib (`9caeba10`).

Please move boundedness out of "consume" and make it a Layer-0 definition target. Since Mathlib has
`Bornology`, the idiomatic form is probably to construct the bornology of bounded subsets and use
`Bornology.IsBounded`, rather than a parallel predicate.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+completed algebra: coefficients tending to `0`, with its Huber topology), and strong
+noetherianness: `A` is strongly noetherian when every completed algebra `A⟨T₁, …, Tₖ⟩` is
+noetherian.
+
+**Main results.**
+
+- Basic theory of rings of definition (Corollary 6.4, Lemma 6.6): powers of `I` are open,
+  elements of `I` are topologically nilpotent, rings of definition are exactly the open,
+  bounded, adically topologized subrings. A Huber ring is nonarchimedean. `A°` is open and
+  integrally closed. In a Tate ring, a pseudouniformizer lies in every ring of definition
```
kim-em: **Citation.** "Rings of definition are exactly the open, bounded, adically topologized
subrings" is **Lemma 6.2**, not 6.4 or 6.6.

Corollary 6.4 does support your neighbouring claims (intersections and products of rings of
definition, `A°` as their union), so keep it. **Lemma 6.6** is the open-ideal criterion (`𝔞` open iff
`n ⊆ rad 𝔞`), which you do need, but in Layer 2: it is what Remark 7.30(1) uses to characterise
openness of `T·A`.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  after a power, and `(ϖⁿA₀)ₙ` is a neighbourhood basis of `0`.
+- **Completion.** The completion of a Huber ring is a complete Hausdorff Huber ring, with a
+  pair of definition induced from any pair for `A`, and completion preserves the Tate
+  property. The canonical map `A → Â` has dense image and kernel `closure {0}`. If `A` is
+  Hausdorff, it is a topological embedding with dense image. ([Hu2] Lemma 1.6, Wedhorn §6.4.)
```
kim-em: **Citation.** §6.4 is "Examples". The completion itself, dense image, kernel `closure{0}` and
the Hausdorff embedding are **§5.4, Proposition and Definition 5.32 with Example 5.33**. **Remark
6.8** is preservation of the f-adic structure under completion. Wedhorn cites **[Hu2] Lemma 1.6**
only for **Proposition 6.9(1)**, the base change `A ⊗_B B̂ ≅ Â`.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+- **Strong noetherianness.** The `k = 0` case says that `Â` is noetherian; in particular a
+  complete strongly noetherian Tate ring is noetherian. Complete nonarchimedean fields with a
+  rank-one valuation are strongly noetherian (BGR 5.2.6) — stated and proved at rank one,
+  which is what the examples (`ℚ_p`, `F⸨t⸩`) and Layer 7's interval rings need; it is not
+  claimed for higher-rank valued fields.
```
kim-em: **Gap: this clause is a development, not a milestone.**

Mathlib has no Tate algebras at all, and BGR 5.2.6 rests on a substantial Tate-specific
restricted-series and Weierstrass division development. Please split the actual proof into sourced
milestones so a contributor can size it. As written this is plausibly the largest single piece of
Layer 0, hidden in one line.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+3. Deduce quasi-compactness of each basic open in the spectral topology.
+4. Prove the basic opens form a basis closed under finite intersections.
+5. Prove `T0`.
+6. Construct generic points of irreducible closed subsets (quasi-sobriety), and conclude
+   `SpectralSpace (Spv A)`.
+
+The model is Mathlib's Hochster development for `PrimeSpectrum`. The same machinery then
+yields, in Layer 2: `Cont A` is proconstructible in `Spv A`, hence spectral; `Spa (A, A⁺)` is
+proconstructible in `Cont A`, hence spectral; rational subsets form a basis of quasi-compact
+opens.
```
kim-em: **Blocking: this route is not available, and Wedhorn's is missing entirely.**

Wedhorn **Remark 7.6** says in full: "The inclusion `Spv(A, I) ↪ Spv(A)` is not spectral in
general." His actual chain is:

- §7.1 constructs `Spv(A, I) := {v ∈ Spv A : cΓ_v(I) = Γ_v}` (7.1.1) out of cofinality and the convex
  subgroup `cΓ_v(I)` (Lemmas 7.1, 7.2, Definition 7.3, Lemma 7.4), with the retraction
  `r : Spv A → Spv(A, I)` (7.1.2). **Lemma 7.5** proves `Spv(A, I)` spectral.
- **Theorem 7.10**: `Cont(A) = {v ∈ Spv(A, I·A) : v(a) < 1 for all a ∈ I}`.
- **Corollary 7.12**: `Cont(A)` is closed in `Spv(A, I)`, hence spectral.
- **Theorem 7.35**: `Spa A = Cont(A) ∩ ⋂_{a ∈ A⁺} Spv(A, I)(a/1)` is pro-constructible in
  `Spv(A, I)`, hence spectral.

So all of §7.1 is load bearing for Layer 2's headline result and for Proposition 7.49, and none of it
is in the roadmap. Please add `cΓ_v(I)`, `Spv(A, I)`, the retraction, Lemmas 7.1 to 7.5, Theorem 7.10
and Corollary 7.12 as Layer-1 milestones, and correct this sentence.

**Also blocking, same area:** "proconstructible" appears three times here and zero times in pinned
Mathlib. There is `WithConstructibleTopology`, `PrespectralSpace.of_isInducing`,
`of_isClosedEmbedding`, but no pro-constructible subsets and no "a pro-constructible subset of a
spectral space is spectral". Both are Layer-1 milestones.

**And on the six-step plan above:** "the model is Mathlib's Hochster development for `PrimeSpectrum`"
omits the one hard step. Wedhorn **Proposition 4.7** embeds `ρ : Spv A ↪ {0,1}^{A×A}` by
`f |_v g :⟺ |f|_v ≥ |g|_v`, shows the image is cut out by closed conditions (a) to (e) hence compact
by Tychonoff, observes each `Spv(A)(f/s)` is clopen there, and applies **Proposition 3.31**
(quasi-compact, clopen generating family, Kolmogorov implies spectral). Mathlib has no analogue of
3.31 either. Worth naming: with `ValuativeRel` the embedding is almost literal, since a valuative
relation *is* a relation on `A × A` cut out by closed conditions. This is the part you identify as
least complete (36 direct `sorry`s) and currently the part with the least guidance.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+subsets `R(T/s)` for finite `T ⊆ A` with `s ∈ T` and the ideal `T·A` open — for a Tate ring,
+the openness condition is equivalent to the elements of `T` generating the unit ideal,
+`(T)A = A`. (It does not require any element of `T` to be a unit: in `K⟨X⟩` the set
+`{X, 1 − X}` generates the unit ideal with both members nonunits.)
+
+**Main results.** Any open integrally closed subring contains `A°°`; `A°` is the maximal ring
+of integral elements. `Cont A` is spectral ([Hu1] Theorem 3.1 shape). `Spa (A, A⁺)` is
+spectral with the rational subsets a basis of quasi-compact opens (Theorem 7.35); rational
+subsets are stable under finite intersection (Remark 7.30);
+`Spa (A, A⁺) = ∅ ↔ A = 0` for complete `A` (Proposition 7.32 shape);
```
kim-em: **Citation, and the statement is false as it would read in Lean.**

7.32 is a Corollary about a unit `π` with `|π(y)| < |s(y)|`. What you want is **Proposition
7.49(1)**, which holds for an arbitrary affinoid ring: `Spa A = ∅ ⟺ Cont A = ∅ ⟺ A/closure{0} = 0`.

Note the hypothesis: you need **Hausdorff**, not completeness. Under bare Lean `CompleteSpace` the
statement as written is false, which is exactly the trap your own Standing conventions were written
to catch ("Every milestone says explicitly whether Hausdorff is assumed"). Layer 4 and the Layer-7
`B^I` milestones do follow the convention; Layers 2, 3 and 6 do not. Worth a sweep, and please say
which Mathlib spelling you mean (`T2Space`, or `T0Space`, equivalent here).

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+`A⁺ = {f : ∀ v ∈ Spa (A, A⁺), v f ≤ 1}` for complete pairs (Proposition 7.52 shape).
```
kim-em: **Citation and hypothesis.** Proposition 7.52(1) holds for an **arbitrary** affinoid ring.
Only 7.52(2), the unit criterion, needs completeness. So this milestone should *lose* its
completeness hypothesis rather than gain a Hausdorff one.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  `X = Spa (A, A⁺)`, the localization is a complete Huber pair `(A⟨T/s⟩, A⟨T/s⟩⁺)` together
+  with a morphism of pairs `(A, A⁺) → (A⟨T/s⟩, A⟨T/s⟩⁺)` such that: the map is continuous;
+  `s` becomes invertible; `t/s ∈ A⟨T/s⟩⁺` for every `t ∈ T`; and the pair is initial among
+  complete pairs over `(A, A⁺)` with these properties — every such morphism of pairs factors
+  through it uniquely, by a continuous morphism of pairs. The plus ring is defined: the
+  integral closure in `A⟨T/s⟩` of the image of the subring generated by `A⁺` and the
+  fractions `t/s`. The construction is a quotient of Layer 0's restricted power series
+  `A⟨X_t⟩`; the universal property, not the construction, is the public interface. Iterated
+  localization: a rational subset of a rational subset is rational in `A` (Lemma 7.54 =
+  [Hu2] Lemma 2.6).
```
kim-em: **Blocking: this is the Tate shortcut, and Layer 7 needs the general construction.**

For a general Huber ring, Wedhorn builds rational localization in §5.6, not as a quotient of
`A⟨X_t⟩`:

- **Remark and Definition 5.48**: the *weighted* restricted series `A⟨X⟩_T ⊆ A[[X]]`.
- **Proposition and Definition 5.51**: the topological localization `A(T/S)` on `R⁻¹A`, with the
  universal property that `f(s_i)` is invertible and `{f(t)f(s_i)⁻¹}` is **power-bounded** in `B`.
  Its completion is `A⟨T/S⟩`.

This matters directly: `A_inf` is Huber and **not** Tate, so Layer 7 needs the theory in Wedhorn's
generality, and the Tate simplifications you lean on elsewhere (notably that an open ideal is the
unit ideal) are unavailable there. The roadmap never says `A_inf` is non-Tate.

Second, smaller point: your universal property asks for `t/s ∈ A⟨T/s⟩⁺`, where 5.51 asks for
power-boundedness. Those are close but not the same condition, and since the universal property is
declared "the public interface", it should be the right one.

Please add weighted restricted series, `A(T/S)` with its universal property, its completion, and
preservation of the Huber structure as Layer-0/Layer-3 milestones.

**Also, on line 221:** Lemma 7.54 is the **cover refinement** lemma (every open cover of `Spa A`
refines to a standard rational cover), and Wedhorn cites **[Hu3]** Lemma 2.6 for it, not [Hu2].
Iterated localization is **Proposition 8.2(2)**: `Spa A⟨T/s⟩ → Spa A` is a homeomorphism onto
`R(T/s)` and induces a bijection on rational subsets.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  is claimed; they are local rings, and each point's valuation extends to the stalk and
+  factors through its residue field.
+- **The category ladder** (Wedhorn §8.1–8.2). `𝒱^pre`: topological spaces with a presheaf of
+  complete topological rings and an equivalence class of valuations on each stalk; morphisms
+  are continuous maps with presheaf maps compatible with the valuations. Affinoid pre-adic
+  spaces: objects of `𝒱^pre` isomorphic to some `Spa (A, A⁺)` with its presheaf (Remark and
+  Definition 8.10). Pre-adic spaces: objects of `𝒱^pre` admitting an open cover by affinoid
+  pre-adic spaces, with the presheaf adapted to that affinoid basis. `𝒱`: the full
+  subcategory of `𝒱^pre` whose structure presheaf is a sheaf of topological rings. Mathlib's
+  `PresheafedSpace` is the substrate.
```
kim-em: **The `𝒱^pre` axioms are incomplete.** Wedhorn's definition (just before Remark and
Definition 8.7) requires each stalk to be a **local ring** and `supp(v_x)` to be its **maximal
ideal**. You require both for the affinoid structure presheaf above, but the general `𝒱^pre` object
as described here carries neither, and they are what make `k(x)` and the residue-field valuation
meaningful.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  one it means.
+
+**Main results.** `𝒪_X(X) = A` for complete `(A, A⁺)` ([Hu2] Proposition 1.6 shape). The
+finite-cover criterion: `IsSheafyPair A Aplus` holds if and only if the equalizer condition
+holds for finite rational covers of rational subsets. The proof needs, as named steps:
+rational subsets form a basis closed under finite intersections; rational subsets are
+quasi-compact; every open cover of a rational subset refines to a finite rational cover; the
+basis-sheaf criterion for presheaves valued in topological rings; the identification of the
+equalizer topology (products and equalizers in the category of topological rings are created
+by the forgetful functor and preserve completeness); and the extension from the rational
```
kim-em: **Gap: this category is not built anywhere.** Mathlib has `TopCommRingCat`
(`Mathlib/Topology/Category/TopCommRingCat.lean`), which the roadmap never mentions, and nothing
about the complete subcategory or its limits. Please say whether the complete subcategory is built on
`TopCommRingCat`, and make the category and its products/equalizers explicit milestones rather than a
parenthesis.

Note also that an equalizer inherits completeness by being *closed*, which needs the codomain
**separated**. That is a third place where the Hausdorff convention turns out to be structural rather
than cosmetic, so the value category should probably be complete *separated* topological rings.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+`Aplus` a ring of integral elements. No domain hypothesis, no noetherian ring of definition,
+no discreteness of value groups. (Wedhorn states 8.28 for the completion; the complete
+Hausdorff version proved here recovers his formulation through Layer 3's
+completion-insensitivity.)
+
+**Main results.**
+
+- **Cover normalization.** Every open cover of `X` refines to a finite rational cover; every
+  finite rational cover refines to a standard (Laurent-type) cover; acyclicity reduces to
+  simple Laurent covers `{v f ≤ 1} ∪ {v f ≥ 1}` by induction (Wedhorn Lemma 8.31/8.34 shape).
```
kim-em: **Citation.** Lemma 8.31 is flatness of `A⟨X⟩`, not cover normalization. Use **7.54** for
refinement to a standard rational cover, **8.33** for the two-piece Laurent base case, and **8.34**
for reduction of general rational covers. Reserve **Remark 7.55** (the chain
`X ⊇ X₀ ⊇ ⋯ ⊇ X_n = U`) for Proposition 8.30, which is where Wedhorn actually uses it.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+- **The flatness bridge, in five named steps.** (i) The algebraic prelocalization
+  `A[1/s]`-algebra from which the rational localization is completed; (ii) the finitely
+  generated ideal defining its topology; (iii) the theorem that the rational localization is
+  the adic completion of that algebra along that ideal; (iv) the noetherian hypotheses,
+  supplied by strong noetherianness; (v) the application of Mathlib's
+  `AdicCompletion.flat_of_isNoetherian`. Without steps (i)–(iv) the Mathlib lemma does not
+  attach to the construction.
```
kim-em: **Blocking: step (iii) cannot hold, and this is not the proof in the cited source.**

A nonzero Tate ring has no proper open ideal, so its topology is not `I`-adic for any `I`, and the
rational localization is not the adic completion of `A[1/s]` along an ideal of `A[1/s]`. A route
through a ring of definition, followed by inverting a pseudouniformizer to get back to `A`, might
work, but that is a different construction and it is not stated here.

Separately, Wedhorn's **Proposition 8.30** (restriction `𝒪_X(V) → 𝒪_X(U)` between rational subsets is
flat) reduces to Laurent form via **Remark 7.55**, then uses **Remark 8.29**
(`M ⊗_A A⟨X⟩ ≅ M⟨X⟩` for finitely generated `M` over a complete noetherian Tate ring) and **Lemma
8.31** (`A⟨X⟩` faithfully flat, and `A⟨X⟩/(f−X)`, `A⟨X⟩/(1−fX)` flat). Adic completion of a
noetherian ring never enters. 8.29 and 8.30 should be milestones here either way.

Please either follow 8.29/8.30/8.31, or spell out the ring-of-definition construction and the descent
back to `A`. This is the most consequential design decision in the roadmap, and "do things right the
first time" says it shouldn't be settled by which Mathlib lemma happens to exist.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+- **Separatedness.** Over a rational cover, the restriction map into the product of the cover
+  rings is injective with closed image and the quotient topologies match (Wedhorn Corollary
+  8.32, Lemmas 8.33–8.34), using the open-mapping corollary from Layer 0.
```
kim-em: **The topological conclusions are not in the citations.** Corollary 8.32 gives faithful
flatness and injectivity; Lemmas 8.33 and 8.34 are Čech exactness. Closed image and matching quotient
topologies are extra, presumably from exactness plus Hausdorffness plus the open mapping theorem.
Please state them as their own steps rather than as direct citations.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  Theorem 8.28(b), [Hu2] Theorem 2.2(ii) — by Laurent-cover induction on the Čech complex,
+  Tate's argument in Huber's generality.
+- **Čech acyclicity, in all degrees.** For every rational subset `U` and finite rational cover
+  `𝔘` of `U`, the augmented Čech complex `0 → 𝒪_X(U) → ∏ 𝒪_X(U_i) → ∏ 𝒪_X(U_ij) → ⋯` is
+  exact, over Mathlib's `HomologicalComplex` API. This is the Čech-level version used by all
+  later applications. Wedhorn's own statement of 8.28 is the sheaf property together with the
+  vanishing `H^q(U, 𝒪_X) = 0` for `q ≥ 1`; deducing that form requires a Čech-to-sheaf-
+  cohomology comparison, and this roadmap does not develop sheaf cohomology. The claim made
+  and proved here is the Čech-acyclicity theorem, and the text says so rather than calling it
+  verbatim 8.28.
```
kim-em: **Missing milestone: Wedhorn Corollary 8.35.** An f-adic ring satisfying 8.28(a)/(b)/(c) is
**stably sheafy**, because those hypotheses are stable under topologically finite-type extension.
Exporting it from Layer 4 makes `IsStablySheafyRing` usable immediately, and Layer 7 can then cite it
rather than re-deriving stable sheafiness of the charts from Kedlaya 4.10.

Two more cheap ones worth adding while you are here: **Corollary 7.53** (for complete affinoid `A`,
`T` generates the unit ideal iff `(R(T/t))_{t∈T}` covers) and **Proposition 7.34** ([Hu2] 3.10,
stability of rational subsets under small perturbation of `s` and the `tᵢ`; note it assumes
completeness).

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+
+**Definitions.** Affinoid adic spaces: objects of `𝒱` isomorphic to `Spa (A, A⁺)` for a pair
+with `IsSheafyPair`. Adic spaces: objects of `𝒱` locally isomorphic in `𝒱` to affinoid adic
+spaces — carrying the presheaf and the stalk valuations, not merely locally homeomorphic.
+Morphisms are `𝒱`-morphisms; open adic subspaces.
+
+**Gluing, specified.** Gluing data: an index set, objects of `𝒱`, open subobjects, transition
+isomorphisms, and the cocycle condition. Construction: the quotient topological space; the
+glued presheaf and the sheaf property; the glued stalk valuations; the universal property of
+the glued object; preservation of the pre-adic and adic conditions.
```
kim-em: **Reuse Mathlib here.** `CategoryTheory.GlueData` (`Mathlib/CategoryTheory/GlueData.lean`)
and `PresheafedSpace` gluing (`Mathlib/Geometry/RingedSpace/PresheafedSpace/Gluing.lean`) already
exist. Please say which is instantiated and what additional work the stalk valuations need, rather
than specifying gluing data from scratch.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+
+**Dependencies.** Layers 3–4.
+
+**Status.** The existing `AdicSpace` structure in provenance asks only local homeomorphism
+and is a placeholder; this layer's definition is the specification.
+
+### Layer 6: uniformity, Buzzard–Verberkmoes, and a new counterexample ([BV]; [HK])
+
+**Definitions.** `IsUniform A` (`A°` bounded, Definition 7.36); `IsStablyUniform A Aplus`
+(every rational localization is uniform, Definition 7.37).
```
kim-em: **Citation.** Wedhorn has **no notion of uniformity at all**: the word does not occur in the
paper outside "uniform space" and "uniformizing element". 7.36 is a Corollary about constructible
subsets and 7.37 is an Example about `Spa(K, A(v))` for a microbial valuation.

Cite **[HK] Definition 2.3** for uniform (`A°` bounded) and **[HK] Definition 3.13** for stably
uniform. The References line "§7 (`Cont`, pairs, `Spa`, rational subsets, uniformity)" needs the same
fix.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+**Main results.** Stably uniform complete Tate pairs are sheafy ([BV], in the
+bounded-denominator formulation) — the sheafiness criterion complementary to Layer 4.
+
+**A new counterexample.** Hansen–Kedlaya (April 2025) record as open whether a uniform
+sheafy Huber ring must be stably uniform ([HK] Remark 3.16). The following construction
+answers that question negatively, and the proof is the AINTLIB formalisation (the `FJP/`
+directory, eleven files, no direct `sorry` at the pin). Its formal status carries the same
+caveat as everything downstream of Layer 4: the files consume the sheafiness capstone, so
+the machine-checked claim is complete once the Layer-4 dependency cone is closed and the
+`#print axioms` audit passes — the standard migration gate, applied here as elsewhere.
```
kim-em: The construction looks plausible to us: `𝓐 → 𝓒` is injective because `𝓑 → 𝓓` is, so `𝓐` is a
domain; `𝓑` and `𝓓` are the non-reduced vertices; the square is exact because `𝓒 → 𝓓` is surjective;
and the concrete description as `{c ∈ L⟨Q⟩ : c₀, c₁ ∈ K⟨W⟩}` checks out. Four requests.

1. **State the formal status before the proof claim, not after.** "the proof is the AINTLIB
   formalisation ... no direct `sorry` at the pin" is accurate but reads stronger than the situation:
   the exported theorem is not yet axiom-free, because its dependency cone runs through Layers 3 and
   4. Worth saying plainly that the unfinished part is a re-proof of a published theorem, which is a
   normal state and not a reason to doubt the mathematics, so a reader can calibrate.
2. **Attribute it.** "The Layer-6 counterexample has no paper reference" leaves an unattributed new
   theorem. Who found it? The name belongs here and in the eventual file headers.
3. **Get a subject expert to look at it** before the roadmap canonizes it, ideally Hansen or Kedlaya.
4. **Be precise about what is new.** [HK] Definition 3.13 already points at uniform but not stably
   uniform rings ([BV] Proposition 17, and Mihara). The novelty here is the *sheafy* part.

Two smaller things in this layer: **`ϖ` is never defined** (it appears only in `𝓐⟨W/ϖ⟩` on line 356,
and is introduced as `F`'s pseudouniformizer only in Layer 7), and that localization should be given
as an explicit `R(T/s)`. Also **`FJP` is never expanded**, in five occurrences.

Finally, the Buzzard-Verberkmoes statement above is described only as "the bounded-denominator
formulation". Please pin the exact statement and say whether it is the pair-level or ring-level
notion.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  lemma do the rational windows below cover points of every rank.
+- **Frobenius and the windows.** `φ` is the Witt Frobenius, acting on `Spa` by
+  `φ_Spa(v) = v ∘ φ`; with `κ(v) = log v([ϖ]) / log v(p)` as mnemonic, `κ ∘ φ_Spa = p·κ`
+  (the inverse action gives `p⁻¹κ`; both conventions occur in the literature, this one is
+  fixed here). The windows `U_n`, `V_n` are defined rank-freely by clearing denominators
+  (`κ ≥ a/b :⇔ v([ϖ])^b ≤ v(p)^a`); no real-valued `κ` is constructed. The breakpoint
+  `c = (p+1)/2` is this roadmap's choice (Kedlaya's displayed windows use `1 + 1/p`).
+  Milestones: each window is a rational subset, hence quasi-compact; the windows cover `𝒴`
+  (via power-comparability); `φ` shifts the window index; same-family windows at different
+  indices are disjoint; hence the `φ^ℤ`-action is free and wandering.
```
kim-em: **Unreconstructible.** The breakpoint `c` is pinned but `U_n` and `V_n` are never defined, so
a contributor cannot recover them from the text. (`U_n = {p^n ≤ κ ≤ p^n c}` and
`V_n = {p^n c ≤ κ ≤ p^{n+1}}` give same-family disjointness and a cover, if that is what's meant.)

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+- `𝒴` is nonempty; `𝒳` is quasi-compact and `T0`; the two window images cover it; its
+  structure presheaf is a sheaf of topological rings (Layer 7).
+- `B^I` strongly noetherian feeding the Layer-4 theorem: `Spa (B^I, B^{I,+})` sheafy.
+- The open unit disc as a glued non-affinoid adic space (Layer 5).
+
+## Ordering
+
+Layer 0 → Layer 1 → Layer 2 → Layer 3 → Layer 4 → Layer 5, in dependency order. Layer 6 needs
+Layers 3–4 and is independent of Layer 5. Layer 7's `A_inf`/window strand needs only Layers
+0–2 and can start early; its charts need Layer 4, and its final clause needs Layer 5.
```
kim-em: **Ordering.** The structure-sheaf-on-`𝒳` bullet in Layer 7 is in neither strand: its sheaf
property depends on the charts being sheafy, so it needs Layer 4 too. Worth separating the pre-adic
chart identification from the passage to `𝒱`, and saying so here.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+
+## References
+
+- R. Huber, *Continuous valuations*, Math. Z. 212 (1993), 455–477 — [Hu1].
+- R. Huber, *A generalization of formal schemes and rigid analytic varieties*, Math. Z. 217
+  (1994), 513–551 — [Hu2].
+- R. Huber, *Étale cohomology of rigid analytic varieties and adic spaces*, Vieweg 1996 —
+  [Hu3] (background; its étale theory is out of scope).
```
kim-em, line 882: [Hu3] is not background: Wedhorn cites **[Hu3] Lemma 2.6** for the cover-refinement lemma
(7.54) that Layers 3 and 4 both need. It is load bearing.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+- T. Wedhorn, *Adic Spaces* (arXiv:1910.05934) — the coordinate system: §4 (`Spv`), §6
+  (Huber rings), §7 (`Cont`, pairs, `Spa`, rational subsets, uniformity), §8 (presheaf,
+  `𝒱^pre`/`𝒱`, pre-adic and adic spaces, Theorem 8.28).
+- J. Tate, *Rigid analytic spaces*, Invent. Math. 12 (1971), 257–289.
+- K. Buzzard, A. Verberkmoes, *Stably uniform affinoids are sheafy*, J. reine angew. Math.
+  740 (2018), 25–39 — [BV].
+- D. Hansen, K. Kedlaya, *Sheafiness criteria for Huber rings* (preprint, April 2025
+  version) — [HK]; Remark 3.16 is the open question Layer 6 addresses.
```
kim-em: Please add a URL and exact date: https://kskedlaya.org/papers/criteria.pdf, dated
**April 23 2025**. As it stands "(preprint, April 2025 version)" is not something a contributor can
follow. For the record, Remark 3.16 reads "We do not know whether or not a uniform Huber ring that is
sheafy is necessarily stably uniform. Compare [31, Remark 2.8.11] and Remark 6.19 below.", so your
characterisation of it is accurate.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+no cross-file dependence, so they are recorded separately from the transitive audit, which is
+a `#print axioms` gate on the actual capstones in TauCeti CI. The project's `ScottishBook/`,
+`FarguesFontaine.lean`, `AlmostMathematics.lean`, `PerfectoidRing.lean`,
+`PerfectoidSpace.lean`, and `Tilting.lean` are out of scope and not migration targets.
+
+| Milestone | AINTLIB source | Direct `sorry` | Transitive audit | Upstream | Status |
+|---|---|---:|---|---|---|
+| Huber/Tate definitions, basic theory | `HuberRings.lean`, `Bounded.lean`, `OpenIdeals.lean`, `PseudoUniformizer.lean` | 0 | pending | mathlib4#42312 | migrate, align with upstream |
+| Restricted series, strong noetherianness | `RestrictedPowerSeries.lean`, `TateAlgebra*.lean` | 0 | pending | — | migrate |
+| Open mapping theorem | `BanachOMT.lean`, `OpenMapping.lean` | open | — | — | incomplete |
```
kim-em: The "Direct `sorry`" column says "open" where every other row has a number. A count or `n/a`
would be better.

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+| Layer-6 counterexample | `FJP/` (11 files) | 0 | inherits Layers 3–4 | — | implemented |
+| Quotient curve topology, windows | `YSpace.lean`, `Curve.lean`, Frobenius strand | 0 | inherits Layers 3–4 | — | implemented |
+| Interval rings through Kedlaya 4.10 | interval-ring strand, `StronglyNoetherianB.lean`, `SheafyBI.lean` | 0 | inherits Layers 3–4 | — | implemented |
+| Descended sheaf on `𝒳` | `YPresheaf.lean`, `YSheaf.lean`, `CurveObject.lean` | 0 | inherits Layers 3–4 | — | implemented |
+| Chart isomorphism in `𝒱^pre` | absent | — | — | — | new theorem |
+| Gluing, open disc | absent | — | — | — | new |
+
+Vendored inputs: `Vendored/Coram*` (William Coram's restricted-power-series and Gauss-norm
+work; its Mathlib face is `PowerSeries.IsRestricted`) and `Vendored/Xia*` — check upstream
+overlap at migration and coordinate rather than porting blindly.
```
kim-em: `Vendored/Xia*` appears here with no description and no licence note, where `Vendored/Coram*`
at least says what it is. Worth a sentence on each.

(`PowerSeries.IsRestricted` does exist in pinned Mathlib, at
`Mathlib/RingTheory/PowerSeries/Restricted.lean`, Apache-2.0, authored by William Coram, so that part
checks out.)

TauCetiRoadmap/AdicSpaces/README.md
```diff
@@ -0,0 +1,562 @@
...
+  `IsTateRing A`, `IsStronglyNoetherian A`, `IsUniform A`) are `Prop`-valued mixins over
+  `[CommRing A] [TopologicalSpace A] [IsTopologicalRing A]`; ordinary theorems take
+  `Aplus : Subring A` with `IsRingOfIntegralElements Aplus` as explicit arguments; and where
+  the pair is genuinely a categorical object there is a bundled structure
+
+  ```lean
+  structure HuberPair (A : Type u) [CommRing A] [TopologicalSpace A] where
+    plus : Subring A
+    isRingOfIntegralElements : IsRingOfIntegralElements plus
+  ```
```
kim-em: This prototype takes `[CommRing A] [TopologicalSpace A]` with neither `[IsTopologicalRing A]`
nor the Huber condition, but `isRingOfIntegralElements` needs `A°`, which needs the ring topology.

Consider moving the block into `Suggested.lean` so it is machine-checked, rather than leaving Lean in
prose where it can drift or fail to elaborate.

TauCetiRoadmap/AdicSpaces/Suggested.lean
```diff
@@ -0,0 +1,208 @@
...
+`sorry` is allowed in this human-owned roadmap library — these are goals, not proofs.
+Following the roadmap-writing guide, the Layer-0 vocabulary that is fully statable against
+pinned Mathlib is **prototyped as honest definitions** below (`PairOfDefinition`,
+`IsHuberRing`, `IsTateRing` — suggested forms, not the specification), and first milestones
+over it are seeded as `theorem … := sorry`; nothing is a `Prop`-typed placeholder. The layers
+whose central objects are new *types* — `Spv` and its topology (Layer 1), `Spa` and rational
+subsets (Layer 2), rational localization, the structure presheaf and the categories `𝒱^pre`/`𝒱`
+(Layer 3), the Čech complexes (Layer 4), adic spaces (Layer 5), and the Layer-6 example
+rings (Layer 6) — need the very API those layers introduce; they are specified in `README.md`
+with embedded Lean prototypes and built there, not pinned here as `sorry`-typed junk types.
```
kim-em: The file is mostly prose restating `README.md`, and the stated reason for not prototyping the
later layers is right about junk types but a little too strong about everything else. Names are what
contributors collide on, so we'd prototype them as their dependencies land.

Available immediately, honestly, against pinned Mathlib: a corrected `HuberPair` (see the README
comment), boundedness, `IsRingOfIntegralElements`, and `Spv A` as a bundling of `ValuativeRel A`,
which is a real definition rather than a placeholder type.

`IsStronglyNoetherian` (needs restricted series), `IsStablyUniform` (needs rational localization) and
the three sheafiness predicates (need the structure presheaf) genuinely do need later API, so leaving
those out is the right call.

TauCetiRoadmap/AdicSpaces/Suggested.lean
```diff
@@ -0,0 +1,208 @@
...
+  /-- The ring of definition, an open subring of `A`. -/
+  ringOfDefinition : Subring A
+  /-- The ring of definition is open. -/
+  isOpen_ringOfDefinition : IsOpen (ringOfDefinition : Set A)
+  /-- The ideal of definition, an ideal of the ring of definition. -/
+  idealOfDefinition : Ideal ringOfDefinition
+  /-- The ideal of definition is finitely generated. -/
+  fg_idealOfDefinition : idealOfDefinition.FG
+  /-- The subspace topology of the ring of definition is the `I`-adic topology. -/
+  isAdic_idealOfDefinition : IsAdic idealOfDefinition
```
kim-em, line 54: This field relies silently on the subtype topology instance on `ringOfDefinition` being the
induced one. It elaborates, and we compiled the file, but it is worth a comment, since it is the one
place the structure's meaning depends on an instance the reader cannot see.

TauCetiRoadmap/AdicSpaces/Suggested.lean
```diff
@@ -0,0 +1,208 @@
...
+theorem not_isTateRing_padicInt (p : ℕ) [Fact p.Prime] : ¬ IsTateRing ℤ_[p] :=
+  sorry
+
+/-! ## Layer 1: the valuation spectrum (Wedhorn §4; Huber, *Continuous valuations*)
+
+`Spv A` is the type of Mathlib `ValuativeRel` instances on `A` — valuations up to
+equivalence, exactly Mathlib's reading — topologized by the basic opens
+`Spv(A)(f/s) = {v | v f ≤ v s ≠ 0}`, with `supp`, functorial `comap`, quotient and
+localization lifts, and **spectrality** (`CompactSpace` + `QuasiSober` + `PrespectralSpace` +
+`QuasiSeparatedSpace`, the conventions' spelling, with `PrimeSpectrum` as the model proof).
```
kim-em: This paraphrase drops `T0Space`, which the README's Standing conventions do list. Simpler to
write `SpectralSpace` and let it carry its own parents.

kim-em:
:robot: Claude+Codex say:

Two corrections to our review, both in our favour of your original text.

**The `[Hu2] Lemma 2.6` row of our citation table was wrong.** We wrote that Wedhorn cites `[Hu3]`
Lemma 2.6 for Lemma 7.54, "not [Hu2]". We had checked Wedhorn's bibliography but not yours, and the
two keys are shifted: Wedhorn's `[Hu1]` is *Bewertungsspektrum und rigide Geometrie*, which this
roadmap does not cite, so his `[Hu3]` is your `[Hu2]`. Your original citation was correct. Thank you
for catching it and for adding the explicit two-keying-schemes note to the References, which is a
better fix than either of us would have got to separately. The other half of that row stands: 7.54
is cover refinement rather than iterated localisation, and Proposition 8.2(2) is the right home for
the latter.

**Our comment that `[Hu3]` is "load bearing, not background" was wrong for the same reason**, and
you were right to leave the References entry as it was.

One more: we suggested building a `Bornology` and using `Bornology.IsBounded` rather than a direct
predicate. mathlib4#40013 defines `IsBounded` directly, and §0.1 and `Suggested.lean` now match it,
so following upstream was the right call and our suggestion would have diverged from it.

Everything else in the review we still stand behind, and the rewrite addresses it. The only comment
we can find that is still open is a small one: `Vendored/Xia*` now has its licence stated but still
no sentence saying what it contains, where `Vendored/Coram*` does.

kim-em (dismissed):
:robot: Claude+Codex say:

Approving. The rewrite addresses every blocking item from the earlier review, and I re-checked it against the sources rather than against our own comments: all 51 numbered Wedhorn citations now resolve to a real result and every one has the right kind, against eight wrong in the previous version. I also verified the Hansen-Kedlaya revision and its §13 credit, Kedlaya Theorem 4.10 with the Remark 4.12 endpoint caveat, Wedhorn 5.50, 7.22 and the §8.5 quotient-mapping definition, the newly cited Mathlib pull requests, and the `ℚ_[p]`/`ℤ_[p]` instance warnings.

The last open comment, the undescribed `Vendored/Xia*` entry, is fixed in a0d636e on this branch.

Two things for later rather than now: Layer 6 still wants a reading from someone who works on Huber rings, and Layer 4.2 formalises Buzzard-Verberkmoes, so Kevin Buzzard is the natural person to ask about both.

CBirkbeck:
Just a note, layer 6 that mentioned the  finite-jet example should not have been there, that was some miscommunication about including interesting examples. Layer 6 is now Fargues--Fontaine, which is right.

kim-em (approved):
:robot: Claude+Codex say:

Re-approving after the removal of the finite-jet pinching layer.

Checked the renumbering for dangling references: cross-references in §0.4 and §5.2, the dependency graph, the scope list, the intro, the acceptance examples, the status table and `Suggested.lean` are all updated, and nothing referring to the removed material survives. The Birkbeck–Torzewski entry is gone from the References while [HK] correctly stays, since Layer 4.2 still uses its Definitions 2.3 and 3.13; the [HK] entry also dropped its now-irrelevant §13 caveat. Buzzard–Verberkmoes and the `K⟨X,Q⟩/(Q²)` acceptance example are independent of the removed layer and correctly retained.

One trivial leftover for whenever this area is next touched: the AINTLIB note lists what is not a migration source, and `FJP/` now belongs in that list but is not named.

PR #92, opened 2026-07-27
fix(PDE): support nonsymmetric ellipticity

kim-em:
## Summary

- define uniform ellipticity for possibly non-symmetric coefficient fields using coercivity of both `a(x)` and `a(x)⁻¹`
- require quantitative dependence on `λ`, `Λ`, or the ellipticity ratio throughout the PDE estimates
- update the weak-form and De Giorgi–Nash–Moser milestones and cite the Armstrong–Kempe formalization as a migration source

## Why

The existing roadmap used a two-sided quadratic-form/Loewner-style formulation. That formulation is appropriate for symmetric coefficients but does not correctly specify the intended non-symmetric divergence-form theory. Replacing inverse coercivity with a primitive matrix-norm bound can also change the upper ellipticity constant and obscure sharp dependence on `Λ`.

## Impact

Future PDE definitions and theorem statements are directed toward a non-symmetric coefficient API with derived mixed upper bounds and explicit quantitative constants.

## Checks

- `git diff --check`
- `lake env lean TauCetiRoadmap/PDE/Suggested.lean`

### DIFF
--- a/TauCetiRoadmap/PDE/README.md
+++ b/TauCetiRoadmap/PDE/README.md
@@ -42,11 +42,22 @@ for the Dirichlet problem `L u = f` in `Ω`, `u = g` on `∂Ω`:
 --     (f : Lp ℝ 2 (volume.restrict Ω)) :
 --     ∃! u : Wkp0 1 2 Ω, ∀ v : Wkp0 1 2 Ω, energyForm a b c u v = ∫ x in Ω, f x * v x
 --
--- De Giorgi–Nash–Moser interior regularity of the weak solution:
--- def holderExponent (n : ℕ) (λ Λ : ℝ) : ℝ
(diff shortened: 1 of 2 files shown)


kim-em:
Closing in favour of fix(PDE): correct ellipticity hypotheses and separate the regularity tracks, https://github.com/TauCetiProject/TauCetiRoadmap/pull/93, which now carries this work.

These were always going to be reviewed together: #93 was stacked on this branch, so its diff against `main` already contained this one in full, and there was never an independent review of this PR to be had. Keeping them split also meant landing an interim version of the coefficient-regularity dial here that #93 immediately rewrote.

Nothing is lost. The non-symmetric ellipticity definition, the derived same-`Λ` upper bounds, and the Armstrong–Kempe migration notes are all in #93, along with the constant conversions between the inverse-coercivity and operator-bound formulations and the corrected De Giorgi–Nash–Moser milestone statement.

PR #93, opened 2026-07-27
fix(PDE): correct ellipticity hypotheses and separate the regularity tracks

kim-em:
This PR corrects the elliptic hypotheses in the PDE roadmap and splits Lane E into divergence-form weak and non-divergence-form strong/classical regularity tracks.

Uniform ellipticity is stated for possibly non-symmetric coefficient fields by coercivity of both `a(x)` and `a(x)⁻¹`, with the quadratic, operator and mixed upper bounds derived at the same `Λ` rather than made primitive. That formulation and the more familiar coercivity-plus-operator-bound are not the same hypothesis at the same constants, so both conversions are recorded: `(λ, M)` in the operator form becomes `(λ, M²/λ)` in the inverse form, while `(λ, Λ)` in the inverse form gives `(λ, Λ)` in the operator form. The gap is real rather than an artifact, witnessed by `a = [[1, −k], [k, 1]]`, whose inverse has symmetric part `I/(1+k²)`, so neither formulation is the sharp one in general and each estimate says which it is calibrated against.

The seeded De Giorgi–Nash–Moser milestone was false as written. Its Hölder constant depended only on `n, λ, Λ, K, Ω`, which `u_M = M x₁` refutes: that is harmonic for every `M`, leaves every one of those parameters fixed, and has unbounded Hölder seminorm. It now restricts to the homogeneous principal part, carries a norm of the solution, types the exponent as `ℝ≥0` with `0 < α ≤ 1` stated as separate lemmas, and concludes for a representative, since an `H¹` weak solution is an a.e. equivalence class and `HolderOnWith … u K` is false for a badly chosen one.

The coefficient-regularity dial now turns separately for the two operator forms. Divergence Schauder gives interior `C^{1,β}` with `β = min(α, 1 - n/q)` under explicit data hypotheses, while non-divergence Schauder gives `C^{2,α}`; the previous text stated the non-divergence conclusion for the divergence-form operator. Meyers higher integrability, lower-`p` well-posedness, divergence VMO `W^{1,p}` and non-divergence VMO `W^{2,p}` are distinguished, each carrying the hypotheses it actually needs: Meyers wants a homogeneous equation or a divergence datum and is not unconditional under arbitrary forcing, the interior `H²` entry wants scalar forcing `f ∈ L²` because a divergence datum in `L²` only reaches `H^{-1}`, Krylov–Safonov wants `f ∈ L^n_loc` and carries its norm, and the smooth-coefficient non-divergence entries are stated for symmetric coefficients. Every entry is an interior estimate for a solution assumed to exist and is quantitative in that solution.

Solution concepts, right-hand-side spaces, interior versus global scope, scalar versus systems scope, and quantitative coefficient moduli are made explicit throughout, and the missing prerequisites are added: Sobolev–Poincaré, Campanato, Gehring, the BMO/VMO commutator theorem, and the ABP estimate with its convex-envelope and contact-set measure theory, together with the Krylov–Safonov covering argument.

The Armstrong–Kempe formalization is cited as the migration source for Lane E.23 and described in terms of the four gaps the port has to close: `holder_Moser` is stated on the unit ball for coefficients normalized to `λ = 1`, assumes `2 < n`, concludes for a representative, and carries `(∫_{B₁} |u|^{p₀})^{1/p₀}` in its constant.

This supersedes and absorbs fix(PDE): support nonsymmetric ellipticity, https://github.com/TauCetiProject/TauCetiRoadmap/pull/92, which is closed in favour of this one: its diff was already contained here by stacking, and its interim form of the dial is rewritten by the track separation above.

🤖 Prepared with Claude Code


CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `f6b9ae14524fbd56710d34ee061d814a5b0e3f9a`, both changed files, the surrounding function-space/maximum-principle prerequisites, and the actual `DeGiorgi.Holder.PublicEstimate.holder_Moser` source. No local Lean build was run; `Suggested.lean` currently contains no compiled target declarations.

The main corrections are sound: inverse coercivity and operator boundedness are distinguished at the correct constants; the nonsymmetric example exposes the difference; homogeneous DGNM is separated from forcing/lower-order terms; representative and solution-norm dependence are retained; and divergence/non-divergence regularity are no longer conflated. Two targets still need correction.

### 1. The new illustrative Hölder theorem loses the spatial scaling and ellipticity normalization

**Location:** README, “The end goal (v1)”, the commented `weakSolution_holderOn` signature.

The signature accepts arbitrary `Ω`, compact `K ⊆ interior Ω`, and general `λ`, but its bound uses only

`holderConstant n Λ p₀ * ‖u‖_{L^{p₀}(Ω)}`

with exponent `holderExponent n Λ`. Neither the radius/distance to the boundary nor the normalization by `λ` appears. The subsequent prose correctly demands those dependencies, but the concrete example contradicts it.

The missing spatial factor already makes the displayed statement false for the Laplacian with `λ = Λ = 1`. Fix `n ≥ 3` and `p₀ = 2`, let `Ω = B_r`, `K = closedBall(0,r/2)`, and take the harmonic function `u(x)=x₁`. For any fixed `0 < α ≤ 1`, evaluating the Hölder quotient at `±(r/4)e₁` gives a lower bound proportional to `r^(1−α)`, whereas

`‖u‖_{L²(B_r)} = c_n r^(1+n/2)`.

Their ratio tends to infinity as `r → 0`. Passing to the asserted Hölder representative does not repair this: it must agree with the continuous function `x₁` throughout the interior of `K`, since it agrees almost everywhere there.

**Repair:** give the ball statement first, for example

`[v]_{C^{0,α}(B_r)} ≤ C(n, Λ/λ, p₀) r^(−α−n/p₀) ‖u‖_{L^{p₀}(B_{2r})}`,

or the equivalent averaged-integral form with `r^(−α)`. Derive the compact-subset version with explicit dependence on the covering radii / distance to the boundary. Normalize `a` by `λ`, so both the exponent and constant use the normalized ellipticity ratio rather than the unnormalized `Λ`.

The imported [unit-ball `holder_Moser`](https://github.com/scottnarmstrong/DeGiorgi/blob/main/DeGiorgi/Holder/PublicEstimate.lean) correctly fixes the unit ball and normalized lower bound; those restrictions cannot simply be removed from its target. Add the shrinking-ball test as well as the existing amplitude-scaling test.

### 2. The Hopf boundary-point milestone still needs a separate coefficient/boundary contract

**Location:** README Lane C.13, “Maximum principles and ABP”, where the Hopf lemma is grouped with the extension to the general standing elliptic operator.

The standing divergence-form hypothesis—bounded measurable uniformly elliptic coefficients, even with `b=c=0`—does not imply the usual strictly nonzero boundary normal derivative. An interior-sphere condition on the domain alone is not enough.

Here is an explicit principal-part counterexample. On the upper half unit disk in `ℝ²`, use polar coordinates and fix `s>1`. Set

`a = e_r ⊗ e_r + s² e_θ ⊗ e_θ`,
`u(r,θ) = r^s sin θ = x₂ |x|^(s−1)`.

The matrix is symmetric and uniformly elliptic with eigenvalues `1,s²`, and

`div(a ∇u) = (s² − s²) r^(s−2) sin θ = 0`.

The solution is positive inside, vanishes on the flat boundary, and is `C¹` at the origin, but its inward normal derivative there is

`lim_{t→0+} u(0,t)/t = lim_{t→0+} t^(s−1) = 0`.

The origin satisfies the interior-sphere condition. The failure comes from the coefficient field not having the needed boundary regularity, not from lower-order terms or a bad weak representative.

**Repair:** state the initial Hopf theorem in a classical setting with the requisite boundary point geometry and coefficient control near that point, or give an explicitly sourced divergence-form boundary theorem with its stronger coefficient assumptions. Keep it separate from bounded-measurable DGNM and the weak/strong maximum principles. For a divergence-form extension, [Rosales, *Generalizing Hopf's boundary point lemma*](https://arxiv.org/abs/1806.07003) is one precise source with Hölder top-order coefficients; its actual hypotheses should be part of that milestone.

These are focused specification repairs. They do not require reopening the corrected ellipticity conversions or the separation of the two regularity tracks.

CBirkbeck (commented):

**Assessment: not yet ready for approval.** I checked this head against the September 8 review. The two specific requested repairs are still absent from the current specification; the valid ellipticity and divergence/non-divergence corrections should be retained.

**1. The displayed `weakSolution_holderOn` is still false on arbitrary domains.** It takes arbitrary Ω and compact K inside Ω, but its constant has no radius or distance-to-boundary dependence. The shrinking-ball test still applies: for the Laplacian, Ω=B_r, K=closedBall(0,r/2), and u(x)=x₁, the Hölder seminorm is of order r^(1−α), whereas the L² norm on Ω is of order r^(1+n/2). Thus a constant depending only on n, Λ and p₀ cannot work as r tends to zero. It also needs the normalized ellipticity ratio Λ/λ, rather than Λ alone.

Please replace the illustrative signature by the ball estimate

`[v]_{C^{0,α}(B_r)} ≤ C(n, Λ/λ, p₀) r^(−α−n/p₀) ‖u‖_{L^{p₀}(B_{2r})}`,

or its averaged-integral version, and derive the compact-subset statement with its geometric dependence explicit. Retain the a.e. representative and amplitude-norm corrections already made. The existing unit-ball migration source is not a theorem at arbitrary scale without this rescaling argument.

**2. Lane C.13 still groups Hopf's boundary-point lemma under the general standing elliptic hypotheses.** Bounded measurable divergence-form coefficients do not suffice, even on a domain with an interior tangent ball. The previous explicit example remains applicable: in a half disk, `a=e_r⊗e_r+s²e_θ⊗e_θ` and `u=r^s sin θ`, with s>1, give a positive solution of `div(a∇u)=0` whose inward normal derivative at the origin is zero. Please state the initial Hopf milestone with a precise classical coefficient/boundary contract, or a specifically sourced stronger divergence-form boundary theorem. Do not inherit bounded-measurable hypotheses from DGNM.

These are the same two focused issues as in review 5139234642, not new requests to redesign the regularity tracks. The nonsymmetric inverse-coercivity/operator-bound conversions, distinction between scalar and divergence forcing, and separation of divergence C^{1,β} from non-divergence C^{2,α} are all improvements. No local Lean build was run; branch conflicts are an additional, separate merge requirement.

CBirkbeck (changes requested):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `78798ff1daecc5bc29f2f512e78f53a948abee1a`.

The previous Hölder-estimate objection is resolved. The ball estimate now has `r^(-α-n/p₀)`, the compact-set version has an explicit interior radius, and both use `Λ/λ`. I checked the actual `holder_Moser` source in `DeGiorgi/Holder/PublicEstimate.lean`: the unit ball, normalization, dimension restriction, representative and solution norm are correctly treated as inputs to a port/rescaling argument, not silently discarded. The ellipticity conversions and divergence/non-divergence/data-space corrections also remain sound.

One narrower specification issue remains in Lane C.13. The new phrase 'a classical coefficient and boundary contract — a ∈ C^{0,α} (or Lipschitz) with a C^{1,1} interior tangent ball' still mixes two different Hopf theorems without stating which operator and solution class it means. For divergence form, Hölder continuity alone does not let the classical exponential-barrier proof differentiate a and treat its derivatives as bounded drift. A divergence-form Hölder-coefficient theorem is available, but uses a different barrier/regularity argument and needs its contract stated. This is not a claim that Hölder coefficients make Hopf false.

Please choose a precise initial target. For example, state a non-divergence classical theorem with continuous boundary values, a C² interior solution, bounded coefficients, ellipticity, the appropriate zero-order sign, a strict boundary extremum and an interior tangent ball. Alternatively state an explicitly sourced weak divergence-form theorem. Rosales, *Generalizing Hopf's boundary point lemma*, arXiv:1806.07003, Theorem 4.1 and Remark 4.2, gives the latter with Hölder top-order coefficients, quantified lower-order assumptions/sign, a boundary-point symmetry hypothesis, and a positive normal **liminf** conclusion. A normal-derivative conclusion additionally needs that derivative to exist (or a boundary regularity result producing it). The symmetry restriction can be removed by a justified reduction, but should not disappear without an argument.

This can be a small edit: make the operator, coefficient control up to the boundary, solution regularity, extremum and conclusion explicit, and assign the barrier/regularity dependency. The measurable-coefficient counterexample is now correctly excluded, and the scaling objection should not be repeated. A subsidiary wording fix: 'Hölder exponent above one implies constant on a connected set' needs a connected open/rectifiably connected domain, not an arbitrary connected subset.

No local Lean build was run; Suggested.lean remains explanatory text rather than an elaborated PDE certificate.

PR #94, opened 2026-07-27
Add roadmap: geodesics, the exponential map, and Hopf-Rinow

AxelDlv00:
This PR adds a roadmap for the Hopf-Rinow theorem, treating it as one milestone in a broader Riemannian-geometry development.

The roadmap starts from the current Mathlib situation around Riemannian geometry: Mathlib has the manifold, tangent-bundle, Riemannian metric, induced distance, and path-length substrate, but not yet the geodesic equation, the exponential map, geodesic completeness, normal neighbourhoods, or the Hopf-Rinow equivalence itself. The roadmap then lays out the missing infrastructure in layers, so contributors can build toward Hopf-Rinow without jumping straight to the final theorem.

The intended route follows do Carmo as the main reference, with Lee and Petersen used as supporting references and cross-checks. I have also included worked examples and suggested Lean signatures, following the style of existing roadmaps, while keeping the Lean targets schematic enough not to over-constrain the eventual API.

This is my first Tau Ceti roadmap PR, so I have tried to follow the conventions used in the existing roadmaps: describe the Mathlib situation, make the dependencies explicit, split the expected build into layers, give worked examples, and specify precise references. Please let me know if the format is not quite what is expected.

I already have a version of the Hopf-Rinow formalization following this roadmap, and it is currently being human-reviewed for mathematical faithfulness.

🤖 Prepared with the assistance of Codex (GPT-5-Sol).


TauCetiRoadmap/HopfRinow/Suggested.lean
```diff
@@ -0,0 +1,66 @@
...
+
+open Set
+open scoped Manifold
+
+variable {E : Type*} [NormedAddCommGroup E] [NormedSpace ℝ E]
+  [FiniteDimensional ℝ E] [NeZero (Module.finrank ℝ E)]
+  {H : Type*} [TopologicalSpace H] {I : ModelWithCorners ℝ E H}
+  {M : Type*} [TopologicalSpace M] [MetricSpace M] [ChartedSpace H M] [IsManifold I 1 M]
+  [∀ x : M, ENorm (TangentSpace I x)]
+  [Bundle.RiemannianBundle (fun x : M => TangentSpace I x)]
```
kim-em: :robot: Claude+Codex say:

Three problems, all of which change what the statements below mean. Mathlib documents the
intended block at `Mathlib/Geometry/Manifold/Riemannian/Basic.lean:30-50` ("the standard way to
say *let `M` be a `C^∞` Riemannian manifold*"); please follow it.

1. `[TopologicalSpace M]` **and** `[MetricSpace M]` are two unrelated structures. I checked with
   `pp.explicit`: `ChartedSpace H M`, `IsManifold`, `ConnectedSpace M` and `CompactSpace M` all
   resolve to the free `TopologicalSpace` variable, while `dist`, `CompleteSpace M` and
   `ProperSpace M` come from the metric. Nothing identifies the two topologies -
   `IsRiemannianManifold` only says `edist = riemannianEDist`. So `isGeodesicallyComplete_of_compactSpace`
   assumes compactness in one topology and concludes about a metric structure on another. Drop
   `[TopologicalSpace M]` and let the (e)metric supply the topology, as Mathlib's block does.

2. `[∀ x : M, ENorm (TangentSpace I x)]` is a free norm unrelated to the Riemannian metric, and it
   shadows the one `RiemannianBundle` induces (that instance is `scoped` in `Bundle`, and this file
   does not `open scoped Bundle`). Elaborating the block shows `Manifold.pathELength I γ 0 1`
   picking up the free `ENorm`, while `IsRiemannianManifold`/`riemannianEDist` use the metric's.
   So `pathELength I γ 0 1 = ENNReal.ofReal (dist p q)` equates a length in an arbitrary norm with
   the Riemannian distance. Delete the `ENorm` variable and add `open scoped Bundle`.

3. The metric is never assumed smooth: `[IsManifold I 1 M]` with no
   `[IsContMDiffRiemannianBundle I ∞ E (fun x ↦ TangentSpace I x)]`. A `C^0` metric has no
   Christoffel symbols and no geodesics, and `C^1` is in any case too weak for the ODE theory.
   Also missing: `[I.Boundaryless]`, which the README's standing hypotheses call for in prose.

`[NeZero (Module.finrank ℝ E)]` (line 17) is a non-idiom - Mathlib says `[Nontrivial E]` or
`0 < finrank` - and it is not needed: see the comment on the standing hypotheses.

TauCetiRoadmap/HopfRinow/Suggested.lean
```diff
@@ -0,0 +1,66 @@
...
+
+/-- Suggested shape for condition (a): the exponential map at `p` is defined on the whole tangent
+space. The domain is a parameter until the exponential-map API exists. -/
+def ExpEverywhereAt (expDomain : (p : M) → Set (TangentSpace I p)) (p : M) : Prop :=
+  expDomain p = univ
+
+/-- Suggested shape for Hopf-Rinow's minimizing-geodesic conclusion. -/
+def HasLengthMinimizingGeodesicsFrom (IsGeodesic : (ℝ → M) → Prop) (p : M) : Prop :=
+  ∀ q : M, ∃ γ : ℝ → M, γ 0 = p ∧ γ 1 = q ∧ IsGeodesic γ ∧
+    Manifold.pathELength I γ 0 1 = ENNReal.ofReal (dist p q)
```
kim-em: :robot: Claude+Codex say:

Parameterising by an uninterpreted `IsGeodesic : (ℝ → M) → Prop` does not postpone the API
decision; it makes the theorems false. `completeSpace_iff_isGeodesicallyComplete` quantifies over
*all* predicates, so instantiating `IsGeodesic := fun _ ↦ True` makes the right side hold (take
`γ` constant) and forces `CompleteSpace M` for every `M` in scope; `fun _ ↦ False` breaks the
other direction and refutes `isGeodesicallyComplete_of_compactSpace` outright. Same for
`hopfRinow` with `expDomain` free.

`IsGeodesicallyComplete` is also wrong independently of the parameter: `_v` is unused, so the
`∀ v : TangentSpace I p` binder says nothing (tangent spaces are always inhabited), and the
definition asks only that *some* geodesic passes through `p` - not that the geodesic with initial
velocity `v` is defined on all of `ℝ`. It needs `γ 0 = p ∧ mfderiv 𝓘(ℝ, ℝ) I γ 0 1 = v` (or
whatever velocity spelling you settle on) and a statement that the domain is all of `ℝ`.

The repository README rules out naming conditions you cannot yet state ("omit a condition you
cannot state rather than name an empty one"), and a free `Prop`-valued parameter is the same
manoeuvre as a `Prop`-typed field. Two ways out, either fine:

- state the milestones against a real Mathlib object: take
  `(∇ : CovariantDerivative I E (fun x ↦ TangentSpace I x))` with the metric-compatibility and
  torsion-free hypotheses (`CovariantDerivative.torsion_eq_zero_iff` is in Mathlib) as an explicit
  argument, and define the geodesic predicate from it. That is honest, connects to
  https://github.com/leanprover-community/mathlib4/pull/36845 "feat: the Levi-Civita connection on
  a manifold", and it exposes the real gap - the covariant derivative *along a curve*;
- or state only what is expressible today (the Layer 0 length-distance identity is, and would be a
  much better first compiled target), and add the geodesic-level statements when Layer 1's types
  exist. That is what line 90-91 of the README already tells implementors to do.

Separately, and this one propagates into the README (lines 116-121 and 134-135): every geodesic
here is a total function `ℝ → M` with a single predicate `IsGeodesic γ`, which conflates a geodesic
*segment*, a maximal geodesic, and an all-time geodesic. That is not a presentational choice - it
makes (f) unstatable in the incomplete case, because the minimizing segment from `p` to `q` in an
open ball does not extend to a geodesic on `ℝ`, and it makes the homogeneity law at line 116
meaningless without domain qualifiers. Pin an interval-aware predicate (`IsGeodesicOn γ s`), define
the maximal interval explicitly, use `[0,1]` for (f), and reserve "domain `= univ`" for
completeness.

Minor, once the above is fixed: `ENNReal.ofReal (dist p q)` is better written `edist p q` (then
`IsRiemannianManifold.out` rewrites it to `riemannianEDist`), and the `TFAE` list omits (e)
although the README promises the equivalence of (a)-(e) - (e) needs a Lean form, e.g. quantifying
over sequences `q : ℕ → M` with `Tendsto (fun n ↦ dist p (q n)) atTop atTop`.

TauCetiRoadmap/HopfRinow/Suggested.lean
```diff
@@ -0,0 +1,66 @@
+import Mathlib
+
+/-!
+# Geodesics, the exponential map, and Hopf-Rinow: target signatures
+
+This file is not the roadmap and is not exhaustive. The definitive document is `README.md`.
+The statements here suggest Lean forms for a few milestones; they are intentionally schematic and
+do not commit Tau Ceti to a full geodesic API yet.
```
kim-em: :robot: Claude+Codex say:

Fourteen of the fifteen existing `Suggested.lean` files open with the same bolded note; please use
it verbatim rather than a paraphrase. In particular "they are intentionally schematic and do not commit Tau Ceti to a full
geodesic API yet" reads as licence to defer conventions, which is the opposite of the repository's
"Do things right the first time" and "Pin conventions". The note that belongs here is that the
file is not exhaustive and that `README.md` is definitive.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+stated under that chapter's standing assumption that `M` is **connected** (do Carmo carries this
+as a chapter convention rather than in the theorem line; it is load-bearing — see Standing
+hypotheses). Verbatim:
+
+> Let `M` be a (connected) Riemannian manifold and let `p ∈ M`. The following assertions are
+> equivalent: **(a)** `exp_p` is defined on all of `T_p M`; **(b)** the closed and bounded sets of
+> `M` are compact; **(c)** `M` is complete as a metric space; **(d)** `M` is geodesically
+> complete; **(e)** there exists a sequence of compact sets `K_n ⊆ K_{n+1}`, `⋃ K_n = M`, such
+> that `q_n ∉ K_n ⇒ d(p, q_n) → ∞`. In addition, each of the above implies **(f)**: for any
+> `q ∈ M` there is a geodesic `γ` joining `p` to `q` with `ℓ(γ) = d(p, q)`.
```
kim-em, line 35: :robot: Claude+Codex say:

The block is introduced as "Verbatim" but is not: the paragraph above it says `(connected)` has
been inserted, and the notation has been normalised. Either reproduce the source exactly and put
the connectedness convention in the sentence before it, or drop "Verbatim" and call it a
paraphrase. Small thing, but this quotation is the specification of the headline theorem, so it
should be exactly what it claims to be.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+- **Connectedness** (`[ConnectedSpace M]`) is load-bearing and stated explicitly wherever used:
+  without it the distance is not finite and assertion (f) fails across components. The purely
+  local geodesic theory (Layers 1–2) does not need it; the equivalence and (f) (Layers 3–4) do.
+- **Positive dimension** (`NeZero` on the model's `finrank`). The connected zero-dimensional case
+  is a one-point manifold where everything is trivial; exclude it rather than special-case it.
+- **The distance is `g`'s.** Completeness is a statement about the distance `g` induces
+  (`riemannianEDist`), so every completeness milestone carries the bridge identifying the ambient
+  `dist`/`edist` on `M` with the `g`-induced one. The preferred form takes the metric from
+  `EMetricSpace.ofRiemannianMetric`, making the bridge definitional; a bridge hypothesis is the
+  form used while an independent metric structure coexists.
```
kim-em: :robot: Claude+Codex say:

The section is supposed to pin the conventions, and the load-bearing ones are missing:

- **No smoothness class.** Neither the manifold's `C^k` nor the metric's is pinned anywhere in the
  roadmap. This is the single most consequential convention in the area (Mathlib's block uses
  `[IsManifold I ∞ M]` with `[IsContMDiffRiemannianBundle I ∞ E ...]`, and its docstring explains
  the `C^n` variant). Pick `∞` and say so.
- **No Hausdorffness.** Mathlib manifolds are not automatically `T2`, and
  `EMetricSpace.ofRiemannianMetric` requires `[T3Space M]` (the pseudo version requires
  `[RegularSpace M]`). If the metric structure is taken from `ofRiemannianMetric`, say that `T3`
  is a hypothesis of the construction.
- **Boundarylessness** is in the prose but never in Lean; make it `[I.Boundaryless]`.

Two further points:

- **Positive dimension looks gratuitous.** In dimension `0` a connected manifold is a point and
  every clause of the theorem holds trivially, so the hypothesis only weakens the result and forces
  downstream users to carry it. If there is a real reason (a proof step that divides by the
  dimension, say), state it; otherwise drop it.
- **The identification of `dist` with the induced distance already exists.** "the bridge identifying the ambient `dist`/`edist` on `M`
  with the `g`-induced one" is precisely Mathlib's `IsRiemannianManifold`
  (`class IsRiemannianManifold : Prop where out (x y : M) : edist x y = riemannianEDist I x y`).
  `Suggested.lean` correctly uses it, so this paragraph, and the Layer 0 bullet at line 101, should
  say "use `IsRiemannianManifold`", not describe a predicate to build.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+## What Mathlib already has (consume)
+
+- **Manifolds and tangent bundles.** `ModelWithCorners`, `IsManifold`, `TangentBundle`,
+  `TangentSpace`, `ContMDiff`, `mfderiv`, charts/`extChartAt`
+  (`Mathlib/Geometry/Manifold/`).
+- **The metric-level Riemannian API.** `RiemannianBundle`, `IsRiemannianManifold`,
+  `riemannianEDist`, `EMetricSpace.ofRiemannianMetric`
+  (`Mathlib/Geometry/Manifold/Riemannian/Basic.lean`), and path length
+  (`.../Riemannian/PathELength.lean`) — the substrate every layer builds on.
```
kim-em: :robot: Claude+Codex say:

`riemannianEDist` is defined in `PathELength.lean` (line 208, as `⨅ (γ : Path x y) (_ : CMDiff 1 γ),
∫⁻ x, ‖mfderiv% γ x 1‖ₑ`), not in `Basic.lean`, and its full name is `Manifold.riemannianEDist`.
Worth fixing because the definition's exact shape matters for Layer 0 - see below.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+- **Completeness and properness.** `CompleteSpace`, `ProperSpace` (with `ProperSpace.complete`),
+  `Metric.isCompact_iff_isClosed_bounded`, Cauchy/total-boundedness API
+  (`Mathlib/Topology/MetricSpace/`, `.../EMetricSpace/`). (b) is `ProperSpace M`; (c) is
+  `CompleteSpace M`.
+- **ODE theory.** Picard–Lindelöf existence/uniqueness and `C^k` dependence of flows on initial
+  conditions (`Mathlib/Analysis/ODE/`), consumed for the geodesic equation after its reduction to
+  a first-order system on `TM`.
```
kim-em: :robot: Claude+Codex say:

Two corrections and one omission.

`C^k` dependence of flows on initial conditions is **not** in Mathlib. `Mathlib/Analysis/ODE/`
contains Picard-Lindelof existence, uniqueness (`ODE_solution_unique*`), Gronwall estimates, and
Lipschitz dependence on the initial condition; there is no `C^k` dependence, no flow, and no
maximal interval of existence. Since Layers 1 and 2 need smoothness of `exp`, this has to become a
target rather than a consumed fact (see the top-level comment).

The reduction to "a first-order system on `TM`" also has a manifold-level substrate that the
roadmap does not mention: `Mathlib/Geometry/Manifold/IntegralCurve/{Basic,ExistUnique,Transform,UniformTime}.lean`
(`IsIntegralCurve`, existence and uniqueness, the uniform-time lemma for global integral curves).
That, not raw `Analysis/ODE`, is what the geodesic flow should be built on.

Finally, the "check what's already in motion" rule: there is relevant in-flight Mathlib work that
should be cited alongside the two PRs already in the Prior art section -
https://github.com/leanprover-community/mathlib4/pull/26413 "feat: existence of maximal solutions
for ODEs meeting Picard-Lindelöf conditions" (exactly the maximal interval Layer 1 needs),
https://github.com/leanprover-community/mathlib4/pull/26394 "feat: existence of local flows on
manifolds", and https://github.com/leanprover-community/mathlib4/pull/40062 "refactor(Analysis/ODE):
restate existence and uniqueness using integral curve API".

kim-em (commented):
:robot: Claude+Codex say:

**1. Blocking: another roadmap already claims this material.**

`TauCetiRoadmap/OptimalTransport/README.md`, Layer 7, says explicitly:

> The geometric-topology roadmap, Layer 7, supplies Riemannian volume, connection, and
> curvature. It does **not** currently supply exponential maps, their local inverse logarithms
> on normal neighborhoods, Hopf--Rinow, minimizing geodesics, injectivity radius, or cut-locus
> measurability/nullity needed here; **those are owned by this layer** before any transport
> theorem consumes them.

Optimal Transport Layer 7 item 1 then asks for exactly Layers 1-4 of this roadmap: exponential
maps, logarithms on normal neighbourhoods, geodesic completeness, Hopf-Rinow, existence of
minimizing geodesics. Two roadmaps cannot both own this. Since a dedicated roadmap is the better
home, please make this PR also amend Optimal Transport Layer 7 to cite `HopfRinow/README.md` for
items 1-3 of that list and keep only what is genuinely transport-specific (injectivity radius,
cut-locus measurability and volume-nullity, squared-distance semiconcavity) - or, if the Optimal
Transport author disagrees, resolve it there first. Either way this roadmap should name the
dependency rather than land silently beside it.

The same applies, more weakly, to the Levi-Civita connection: Geometric Topology Layer 7
(`GeometricTopology/README.md:647`) lists "the Levi-Civita connection, the Riemann curvature
tensor, and sectional and Ricci curvature" as its own build. Layer 1 here says the connection is
"built here or in a shared connection home", which leaves it owned by nobody. Pin one owner.

**2. Blocking: the missing ODE input is not a target anywhere.**

Every smoothness statement in Layers 1-2 (`d(exp_p)_0 = id`, `exp_p` a local diffeomorphism,
normal neighbourhoods, the Gauss lemma) rests on `C^k` dependence of ODE solutions on initial
conditions. Mathlib does not have it: `Mathlib/Analysis/ODE/PicardLindelof.lean` gives existence,
uniqueness and *Lipschitz* dependence on the initial condition
(`exists_forall_closedBall_funSpace_dist_le_mul`), `Gronwall.lean` gives continuous dependence,
and that is all. There is no maximal interval of existence and no flow. So the roadmap consumes
something that does not exist, which is exactly the leap the repository README warns about. Make
`C^k` dependence (or the smoothness of the geodesic flow however you prefer to get it) an explicit
milestone here, or cite a Mathlib PR that will supply it.

**3. Other dependencies used but never made targets.** Each of these is needed by a stated
milestone and is neither in Mathlib nor a milestone here:

- finiteness of `riemannianEDist` on a connected manifold - without it there is no `MetricSpace M`
  at all, and the whole roadmap is stated in terms of `dist`. Mathlib proves local finiteness
  (`eventually_riemannianEDist_lt`) but not the global connected statement;
- the covariant derivative *along a curve* (the pullback connection). Mathlib's
  `CovariantDerivative` differentiates global sections; `γ'` is a section along `γ`, so Layer 1's
  definition of `IsGeodesicCurve` cannot be stated without it. It is named in the prose at line 66
  but never made a milestone;
- the **geodesic spray**. "reducing to a first-order system on `TM`" (line 113) names the technique
  but not the object: the vector field on `TM` whose integral curves are the velocity lifts of
  geodesics. Nothing in Mathlib's integral-curve API can be applied until that field is defined,
  shown smooth, and shown to have exactly those integral curves;
- the **inverse function theorem on manifolds**. `d(exp_p)_0 = id` gives a local diffeomorphism
  only through it, and Mathlib does not have it: `Mathlib/Geometry/Manifold/LocalDiffeomorph.lean`
  lists "if `f` is `C^n` at `x` and `mfderiv I J n f x` is a linear isomorphism, `f` is a local
  diffeomorphism at `x` (using the inverse function theorem)" in its `TODO` block (lines 43-46).
  Layer 1's local-diffeomorphism claim and all of Layer 2's normal neighbourhoods rest on it;
- the identification `T_0(T_p M) ≅ T_p M` that `d(exp_p)_0 = id` is stated through;
- an induced/restricted Riemannian metric: on an embedded submanifold (for `S^n`) and on an open
  subset (for the open ball). Neither exists in Mathlib, and both worked examples need one;
- a notion of length or geodesic metric space, for the Layer 4 claim that `(M, d_g)` is one.
  Mathlib has none (there is old Zulip interest in a metric-level `Length.lean` from Evan Bailey,
  #new members, Oct 2024 - worth citing and coordinating with).

**4. `Suggested.lean` does not currently state the roadmap's targets.** All three theorems are
universally quantified over an uninterpreted `IsGeodesic : (ℝ → M) → Prop` (and `expDomain`), so
each is *false*, not merely schematic - see the inline comments. The parameterisation also
conflicts with the repository's rule against naming conditions you cannot state. Details and a
concrete alternative are inline at `Suggested.lean:16` and `:25`.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+## What is missing (build here)
+
+The Levi-Civita connection and covariant derivative along a curve; the geodesic equation and its
+solutions; constant speed; the exponential map and its domain; normal neighbourhoods, the Gauss
+lemma, and minimizing geodesics; geodesic completeness; and the Hopf–Rinow equivalence itself.
+None of this is upstream.
```
kim-em: :robot: Claude+Codex say:

"The Levi-Civita connection and covariant derivative along a curve ... None of this is upstream" is
not right as of the pinned Mathlib. `Mathlib/Geometry/Manifold/VectorBundle/CovariantDerivative/Basic.lean`
and `.../Torsion.lean` already give `IsCovariantDerivativeOn`, the bundled `CovariantDerivative`,
`ContMDiffCovariantDerivative`, the difference of two connections as an endomorphism-valued
one-form, and the torsion tensor with `torsion_eq_zero_iff`. What is actually missing is metric
compatibility plus existence and uniqueness of Levi-Civita (in flight at mathlib4#36845) and the
covariant derivative **along a curve** (the pullback connection). The latter is the one Layer 1's
`IsGeodesicCurve` is stated through, so please promote it from this prose list to a milestone with
its own API. The same correction applies to the opening paragraph at lines 3-10.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+connections and geodesics, and leanprover-community/mathlib4#36845 develops the Levi-Civita
+connection on a manifold. Tau Ceti work on this roadmap should cite those PRs, follow the
+conventions that emerge there, and treat any overlapping local definitions as temporary shims to
+delete or refactor once the Mathlib versions land.
+
+The author has related Lean material in `~/Poincare-Conjecture`, especially the DoCarmo and
+Petersen developments around connections, geodesics, exponential maps, and Hopf–Rinow. That code
+is useful provenance and a source of possible target shapes, but it is not the standard for Tau
+Ceti: implementors should build fresh against the roadmap, current Mathlib, and reviewer
+feedback, improving generality and API design where appropriate.
```
kim-em: :robot: Claude+Codex say:

`~/Poincare-Conjecture` is a path in someone's home directory; an implementor cannot act on it and
a reviewer cannot check it. Give the public URL and the licence, and state the coordination
outcome the repository README asks for ("Coordinate before integrating existing work ... verify
that its licence permits the intended copying or adaptation"). The PR description says a
Hopf-Rinow formalization following this roadmap already exists and is under human review; that
makes this an integration roadmap in part, and the README requires any file-by-file map to sit in
a clearly secondary provenance section, with the roadmap itself specifying the mathematics
intrinsically.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+`Suggested.lean` (with `sorry`).
+
+### Layer 0: the reconciled Riemannian distance
+- **Length of a piecewise-`C¹` curve** and its API: additivity under concatenation, invariance
+  under monotone reparametrization, constant-speed reparametrization, lower semicontinuity under
+  uniform limits. State it through Mathlib's `pathELength` where possible.
+- **The length–distance identity.** do Carmo's infimum over piecewise-`C¹` curves equals
+  `riemannianEDist`; this is what makes every later "distance" the Mathlib one. The tools
+  (piecewise length, chart-straight polygonal approximants with continuity-based length control)
+  are within the metric-level API; the work is assembly.
```
kim-em: :robot: Claude+Codex say:

- The first two items of the length API are already in Mathlib: additivity under concatenation is
  `Manifold.pathELength_add`, and invariance under monotone reparametrisation is
  `Manifold.pathELength_comp_of_monotoneOn`. Consume them rather than listing them as work.
- `pathELength` is `∫⁻ t in Icc a b, ‖mfderiv% γ t 1‖ₑ`, and `mfderiv` is junk-valued (`0`) where
  `γ` is not differentiable. So "lower semicontinuity under uniform limits" is either vacuous or
  false depending on what regularity the limit is assumed to have; state the hypothesis on the
  limit curve explicitly. Likewise, constant-speed reparametrisation in the `C¹` category needs the
  curve to be regular (nowhere-vanishing speed), or the reparametrised curve is not `C¹`.
- The length-distance identity is not "assembly". Mathlib's `riemannianEDist` is already an infimum
  of lengths - but over `C¹` paths `Path x y`, not piecewise-`C¹` curves. So the content of this
  milestone is exactly "the piecewise-`C¹` infimum equals the `C¹` infimum", i.e. a corner-smoothing
  reparametrisation that preserves length. That deserves to be its own named target, and the
  roadmap should say which convention (do Carmo's piecewise-`C¹`, or Mathlib's `C¹`) is canonical
  in Tau Ceti downstream.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+  where the metric-level API already proves a fact, consume it.
+
+### Layer 1: the geodesic equation, the flow, and the exponential map
+- **`IsGeodesicCurve`:** the covariant derivative of the velocity vanishes; equivalently, in a
+  chart, the second-order geodesic ODE with the Christoffel symbols. State it chart-independently
+  and prove the chart form equivalent. (Needs the Levi-Civita connection — built here or in a
+  shared connection home; see the Geometric Topology roadmap's curvature layer for the same
+  substrate.)
+- **Local existence, uniqueness, smooth dependence** of the geodesic from `(p, v)`, consumed from
+  Mathlib's ODE flow theory after reducing to a first-order system on `TM`.
```
kim-em: :robot: Claude+Codex say:

"built here or in a shared connection home" - nothing on a roadmap is optional, and this leaves the
Levi-Civita connection owned by nobody while Geometric Topology Layer 7 also lists it. Decide, and
if it is built here, give it its own milestone with its API (metric compatibility, torsion-freeness,
uniqueness), not a parenthesis.

Line 113's "after reducing to a first-order system on `TM`" also needs an object, not just a
technique: the geodesic spray. Give it milestones of its own - the vector field on `TM`, its chart
formula and chart independence, its smoothness, and the equivalence between its integral curves and
the velocity lifts of solutions of the covariant geodesic equation. Without that, none of Mathlib's
`IntegralCurve` results can be applied.

While you are here, line 118 packs three unowned dependencies into one clause. `d(exp_p)_0 = id` is
stated across the identification `T_0(T_p M) ≅ T_p M`, which is itself a Lean-level target worth
naming; "the maximal interval of existence" has no Mathlib support today (see mathlib4#26413); and
"so `exp_p` is a local diffeomorphism at `0`" needs the inverse function theorem on manifolds,
which Mathlib explicitly does not have - `Mathlib/Geometry/Manifold/LocalDiffeomorph.lean:43-46`
lists it as a TODO. That last one is also required by every normal-neighbourhood statement in Layer 2, so it
should be a named milestone (here or in a cited Mathlib PR), not an aside.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+- **Constant speed** (`‖γ'‖_g` is constant — the connection is metric): the lemma that makes a
+  geodesic Cauchy at a finite endpoint of its interval.
+- **Homogeneity** `γ_{p, λv}(t) = γ_{p, v}(λt)`, the **maximal interval of existence**, and the
+  **exponential map** `exp_p v = γ_{p,v}(1)` with its domain, `exp_p 0 = p`, and
+  `d(exp_p)_0 = id` (so `exp_p` is a local diffeomorphism at `0`).
+- Milestone **(a)** is that the domain of `exp_p` is all of `T_p M`; milestone **(d)** is geodesic
+  completeness (every `v` launches an all-time geodesic). (a) ⇔ (d) at a point, via homogeneity,
+  is proved here.
```
kim-em: :robot: Claude+Codex say:

As written this is false. `(d)` is defined here as global geodesic completeness ("every `v`
launches an all-time geodesic", over all base points), and `(a)` is about a single `p`.
Homogeneity gives only

> `exp_p` is defined on all of `T_p M` ⟺ every geodesic *starting at `p`* is defined on all of `ℝ`,

which is a statement about `p` alone. `(a) ⇒ (d)` for other base points is the deep content of
Hopf-Rinow and is not available in Layer 1. Please restate this bullet as the pointwise equivalence
and move the global step to Layer 3 - and while you are there, state explicitly the striking
corollary that `(a)` at one point implies `(a)` at every point, since that is the thing a reader
will want and it is not visible in do Carmo's list.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+### Layer 2: normal neighbourhoods, the Gauss lemma, and minimizing geodesics
+- **Normal neighbourhoods:** `exp_p` restricts to a diffeomorphism from a star-shaped
+  neighbourhood of `0`, giving normal balls and normal coordinates.
+- **The Gauss lemma** and **local minimization:** inside a normal ball the radial geodesic is the
+  unique shortest path to its endpoint and realizes the distance (stated via the Layer-0 distance,
+  `pathELength γ = riemannianEDist p (γ 1)`).
+- **The first variation of energy**, enough to characterize geodesics as critical points and to
+  drive the endpoint arguments of Layer 3.
+- **Minimizing geodesics from `p`** (milestone **(f)**): every `q` is joined to `p` by a geodesic
+  on `[0,1]` whose subsegments realize distance and whose length equals `d(p, q)`.
```
kim-em: :robot: Claude+Codex say:

- The Gauss-lemma bullet claims two things that need separating. Radial minimality inside a normal
  ball is minimality *against competitors that stay in the ball*; the stated conclusion
  `pathELength γ = riemannianEDist p (γ 1)` is about the global infimum, so it needs the further
  step (on a smaller ball) that a curve leaving the ball is longer. And "unique shortest path" is
  false as stated: any monotone reparametrisation, including one that pauses, has the same length.
  State uniqueness for the constant-speed affinely parametrised representative, or up to monotone
  reparametrisation.
- "The first variation of energy ... to drive the endpoint arguments of Layer 3" - the Layer 3
  arguments as you describe them (the sphere-compactness argument plus continuity of `d(·, q)`)
  do not use the first variation, and do Carmo's proof of 2.8 does not either. Keep the milestone,
  but drop or correct the justification - and specify it, since as one line it is not
  implementable: it needs the energy functional, variations and variation fields, the covariant
  derivative in the variation direction, the integration-by-parts step, and a definition of
  "critical point".
- **Milestone (f) is in the wrong layer and false where it sits.** "every `q` is joined to `p` by a
  geodesic on `[0,1]` ... whose length equals `d(p, q)`" has no completeness hypothesis here, and
  it is false without one: in `ℝ² ∖ {0}`, points on opposite sides of the puncture are joined by no
  minimizing geodesic. (f) also reappears, correctly this time, in Layer 3 at line 143. What
  belongs in Layer 2 is the *local* statement: within a normal ball, the radial geodesic minimizes.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+  interval; metric completeness supplies the limit and a local flow extends past it, so no maximal
+  interval is bounded — and conversely.
+- **(a) ⇒ (b) and (a) ⇒ (f):** with `exp_p` everywhere defined, a compactness argument on spheres
+  in `T_p M` produces a minimizing geodesic to each `q` (f) and exhibits closed balls as compact
+  images, i.e. `ProperSpace M` (b).
+- **(b) ⇒ (c)** is `ProperSpace.complete`; **(b) ⇔ (e)** uses the closed-ball exhaustion
+  `K_n = closedBall p n`.
+- ⚠ **(f) does not imply (b).** Properness follows from (a) — all-time `exp_p` is what makes closed
+  balls compact images of Euclidean balls — *together with* (f), never from the
+  existence-of-minimizers (f) alone. Any narrative must route properness through (a).
```
kim-em: :robot: Claude+Codex say:

The implication graph does not close. As written you have `(a) ⇒ (b)`, `(b) ⇒ (c)`, `(b) ⇔ (e)`,
`(c) ⇒ (d)` and an unexplained "and conversely". Nothing supplies `X ⇒ (a)`, so the `TFAE` cannot
be assembled from these bullets. Add `(d) ⇒ (a)` (immediate) explicitly.

More seriously, "and conversely" hides the hard part rather than naming it. `(d) ⇒ (c)` is not the
Cauchy-limit argument run backwards; it goes `(d) ⇒ (a) ⇒ (b) ⇒ (c)`, and `(a) ⇒ (b)` is the
sphere-compactness argument in the next bullet, which itself routes through `(f)`. The chain worth
writing down is

> `(c) ⇒ (d) ⇒ (a)_p`, `(a)_p ⇒ (f)_p`, `(a)_p` with `(f)_p ⇒ (b) ⇒ (c)`, and `(b) ⇔ (e)`,

since a contributor working from these bullets would otherwise try to prove `(d) ⇒ (c)` directly.

The `(c) ⇒ (d)` step is also stated too briefly to be right. Metric completeness gives a limit of
the *base points* `γ(t)`; restarting the ODE needs a limiting state `(γ(t), γ'(t))` in `TM`.
Constant speed bounds the velocities but does not by itself make them converge - you need a
convergent subsequence of fixed-speed tangent states in a local trivialisation, and then uniqueness
for the spray to identify the extension with the original curve. That is a target, not a clause.

Four smaller things in the same section:

- `(b) ⇔ (e)` is not symmetric: `K_n = closedBall p n` gives `(b) ⇒ (e)` only. The converse is the
  other argument - a closed bounded `A` not contained in any `K_n` yields `q_n ∈ A \ K_n` with
  `d(p, q_n) → ∞`, contradicting boundedness - and should be written out.

- `ProperSpace.complete` does not exist; the lemma is `complete_of_proper`
  (`{α} [PseudoMetricSpace α] [ProperSpace α] : CompleteSpace α`). Likewise
  `Metric.isCompact_iff_isClosed_bounded`, cited at line 57 for `(b)`, carries a `[T2Space α]`
  hypothesis - another reason to pin Hausdorffness in the standing hypotheses.
- the `(c) ⇒ (d)` step needs an extension lemma for maximal integral curves ("a local flow extends
  past it"), which is the Mathlib gap flagged above;
- the `(a) ⇒ (b)` step gives compactness of closed balls *centred at `p`*, so `ProperSpace M` needs
  the extra (easy) step that an arbitrary closed ball is a closed subset of a large ball centred
  at `p`.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Layer 4: corollaries and downstream theory
+- **Compact ⇒ geodesically complete** (do Carmo, Corollary 2.9), with a direct proof (a geodesic
+  on a compact manifold cannot escape, so its interval is unbounded) as well as via completeness.
+- **Minimizing geodesics between arbitrary points** of a complete connected manifold, and hence
+  that `(M, d_g)` is a geodesic (length) metric space.
+- **Closed subsets inherit completeness**; closed bounded subsets are compact; the diameter
+  corollaries.
+- **Transport of geodesic completeness** across isometries and along the distance bridge, so
+  downstream roadmaps (constant-curvature model spaces) apply the theory without reopening it.
```
kim-em: :robot: Claude+Codex say:

- "a geodesic on a compact manifold cannot escape, so its interval is unbounded" is not the
  argument: the geodesic flow lives on `TM`, which is non-compact even when `M` is compact. What
  confines the lifted curve is constant speed - it stays in the compact sphere bundle
  `{‖v‖ = c}`. Say that, because a contributor reaching for Mathlib's uniform-time lemma
  (`IntegralCurve/UniformTime.lean`: `exists_isMIntegralCurve_of_isMIntegralCurveOn`, Lee 9.15,
  which needs `[BoundarylessManifold I M]` and a uniform `ε` obtained from compactness of the
  *whole* manifold) will find it does not apply to `TM` and will need the "remains in a compact
  set" version, plus compactness of the fixed-radius tangent sphere bundle, neither of which
  Mathlib has. If you would rather not own those, drop the direct proof and route the corollary
  through compact ⇒ proper ⇒ complete ⇒ geodesically complete.
- "`(M, d_g)` is a geodesic (length) metric space" has no referent and conflates two notions:
  a length space asks that the distance be the infimum of curve lengths, a geodesic space that the
  infimum be attained by a constant-speed segment. Mathlib has neither (I grepped). Define or cite
  both, separately, at the metric-space level where they belong - and ideally coordinate with the
  Zulip interest in a metric-level `Length.lean`.
- "Closed subsets inherit completeness" is `IsClosed.completeSpace_coe` and "closed bounded subsets
  are compact" is the already-cited `Metric.isCompact_iff_isClosed_bounded`; both belong in the
  consume list, not in Layer 4. "The diameter corollaries" names no statement at all - say which
  ones, with hypotheses.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+- **`ℝⁿ`** (flat metric, `NeZero n`) is geodesically and metrically complete, and the minimizing
+  geodesic from `x` to `y` is the affine segment, with `pathELength = ‖x − y‖`. Both directions of
+  the equivalence are non-vacuous.
+- **`Sⁿ`** is compact, hence geodesically complete by Layer 4, and a great-circle arc realizes the
+  distance. Exercises Corollary 2.9.
+- **An open ball in `ℝⁿ` (or `ℝⁿ ∖ {0}`)** is neither geodesically nor metrically complete, and a
+  radial geodesic leaves it in finite time: the equivalence correctly reports incompleteness, so
+  no hypothesis silently forces completeness.
+- **(f)-without-(b) guard:** a witness where minimizers from `p` exist yet closed balls are not
+  compact, matching the trap in Layer 3.
```
kim-em: :robot: Claude+Codex say:

These are acceptance criteria, so they need to be discharged, and as stated two of the four cannot
be:

- `S^n`: Mathlib gives the sphere a `ChartedSpace` and an analytic `IsManifold` structure
  (`Instances/Sphere.lean:353,387`) but **no** Riemannian metric. The round metric needs the
  induced-metric-on-an-embedded-submanifold construction, which does not exist and is not a
  milestone here. Add it as a target (it is a substantial one) or drop the example.
- the open ball and `ℝⁿ ∖ {0}`: `TopologicalSpace.Opens` carries an `IsManifold` instance
  (`IsManifold/Basic.lean:1011-1023`), but again there is no restricted `RiemannianBundle`. Same
  choice.
- `ℝⁿ`: Mathlib already has `instance : IsRiemannianManifold 𝓘(ℝ, F) F` for an inner product space
  `F` (`Riemannian/Basic.lean:126,158`), with `edist = riemannianEDist` proved. Consume it. `NeZero n`
  is not needed for this example.
- the (f)-without-(b) guard asks for "a witness" without saying which, so it cannot be checked off.
  Name one: the open unit ball in `ℝⁿ` with the flat metric and `p = 0`. Every `q` is joined to `0`
  by the radial segment, which is minimizing, yet `closedBall 0 2` is the whole space and is not
  compact. That is the third example re-used, which is worth saying explicitly.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+  the induced distance, the Levi-Civita connection), **Ch. 3 §2–3** (the geodesic equation and
+  flow, the exponential map, the Gauss lemma, minimizing geodesics, normal neighbourhoods —
+  Layers 0–2), and **Ch. 7 §2** (Theorem 2.8, Hopf–Rinow, and Corollary 2.9 — Layers 3–4). The
+  primary target.
+- J. Lee, *Introduction to Riemannian Manifolds*, GTM 176, 2018: the Levi-Civita connection
+  (Thm 5.10), geodesics and the exponential map (Ch. 5–6), the Gauss lemma and the Hopf–Rinow
+  theorem (Ch. 6). Its connection material is the substrate shared with the Geometric Topology
+  roadmap's curvature layer.
+- P. Petersen, *Riemannian Geometry*, GTM 171: an alternative account of completeness, minimizing
+  geodesics, and the length-space view (cross-checks for Layers 2–4).
```
kim-em, line 463: :robot: Claude+Codex say:

Please check the section numbers, and add a `references/` extract as the Geometric Topology
roadmap does (`GeometricTopology/references/lee-riemannian-volume-curvature.md` is the model) -
the roadmap is stated against theorem numbers a reviewer cannot otherwise verify. Two that look
wrong to me: do Carmo introduces the induced distance `d` in Ch. 7 §2, immediately before
Hopf-Rinow, not in Ch. 1-2; and the geodesic equation and geodesic flow are Ch. 3 §1, so the
citation should be Ch. 3 §1-3. Layers 0-2 as written also reach past §2-3: normal and convex
neighbourhoods are Ch. 3 §4, and the first variation of energy is not in Ch. 3 at all.

TauCetiRoadmap/HopfRinow/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Authorship
+
+Drafted with AI assistance from Codex/GPT-5 and reviewed by the human author before submission.
```
kim-em: :robot: Claude+Codex say:

No other roadmap carries an Authorship section; the PR description is where this belongs. Please
drop it so the roadmaps stay uniform.

TauCetiRoadmap/HopfRinow/Suggested.lean
AxelDlv00: Addressed in ce4a05b.

I replaced the mixed topology/metric assumptions with the Mathlib standard smooth Riemannian-manifold setup, specialized to `MetricSpace M` because the current targets use `dist` and `ProperSpace`:

- the metric structure now supplies the only topology on `M`;
- `open scoped Bundle` supplies the tangent-space norms induced by `RiemannianBundle`, and the free `ENorm` instance is gone;
- the manifold and metric are now smooth via `[IsManifold I ∞ M]` and `[IsContMDiffRiemannianBundle I ∞ E ...]`;
- `[I.Boundaryless]` is explicit;
- finite dimensionality remains, while the unnecessary positive-dimension/`NeZero` assumption has been removed.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/Suggested.lean
AxelDlv00: Addressed in e9f56e4.

I chose the second proposed route: geodesic-level targets are now deferred until Layer 1 provides real types for covariant differentiation along curves, interval-aware geodesics, maximal intervals, and the exponential map.

I removed the declarations parameterized by arbitrary `IsGeodesic` and `expDomain` values. `Suggested.lean` now contains a concrete Layer 0 target identifying `riemannianEDist` with the infimum of `pathELength` over finite piecewise-`C¹` partitions.

I also revised the roadmap to pin the eventual API:

- `IsGeodesicCurveOn g γ s` is relative to a parameter set;
- initial position and velocity are explicit;
- maximal intervals and exponential domains are derived from genuine geodesic witnesses;
- geodesic completeness reserves `domain = univ` for all-time solutions;
- condition (f) uses a geodesic on `Icc 0 1` and length `edist p q`;
- condition (e) now has an explicit Lean formulation.

The existing `Poincare-Conjecture` development grounds these target shapes but is not imported as a Tau Ceti dependency. Its chart-based equation will need a bridge to the eventual Levi-Civita along-curve API, and its standalone minimizing-geodesic predicate will be narrowed from an all-time geodesic to an `[0,1]` segment. Once the Layer 1 types land, the geodesic-level compiled targets will be added back.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/Suggested.lean
AxelDlv00: Addressed in b18f7b6. I replaced the paraphrased introduction with the repository-standard bolded note verbatim, including the wording that these signatures help contributors and reviewers converge on names and signatures.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Addressed in 1c1c613. I dropped the inaccurate “Verbatim” label and now introduce the statement as “Paraphrased in the roadmap’s normalized notation,” which accounts for the inserted connectedness convention and normalized notation.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Addressed in 810912f.

The standing hypotheses now explicitly pin:

- `[IsManifold I ∞ M]` and a `C∞` Riemannian metric through `RiemannianBundle` plus `IsContMDiffRiemannianBundle I ∞ ...`;
- `[I.Boundaryless]`;
- the topology and separation convention: an existing `EMetricSpace` or `MetricSpace` supplies the topology, while constructing `EMetricSpace.ofRiemannianMetric` requires `T3Space M` and the pseudo-emetric construction requires `RegularSpace M`.

I removed the positive-dimension assumption, including the `NeZero` restriction from the Euclidean acceptance example. The connected zero-dimensional case is now retained in the public target.

The distance convention now consumes `IsRiemannianManifold I M` and rewrites with `IsRiemannianManifold.out`; Layer 0 proves connected finiteness and exposes the ordinary metric presentation instead of introducing a private compatibility predicate. I also updated the downstream transport wording to use this existing identification.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `cfd84c5`. I replaced the private home-directory path with the public `frenzymath/Poincare-Conjecture` URL, recorded its Apache-2.0 license, and moved the file map into an explicitly secondary provenance subsection. The roadmap continues to specify the mathematics intrinsically. The coordination outcome is now explicit: the repository owner permits use as prior art and adaptation under Apache-2.0, no source is copied in this roadmap PR, and any later integration must record its source revision and attribution while coordinating overlaps with the cited Mathlib work.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `cfd84c5`. The consumed ODE substrate is now limited to Picard-Lindelöf local existence, uniqueness, Gronwall estimates, and Lipschitz dependence. I added the manifold `IntegralCurve/{Basic,ExistUnique,Transform,UniformTime}` API as the intended substrate for the geodesic spray, made smooth dependence/local geodesic flow and maximal intervals explicit roadmap targets, and cited mathlib4#26413, #26394, and #40062 as in-flight work to consume if it lands first.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `cfd84c5`. The opening and dependency sections now recognize Mathlib's existing general `CovariantDerivative` and torsion APIs, including `CovariantDerivative.torsion_eq_zero_iff`. The missing work is narrowed to metric compatibility, Levi-Civita existence/uniqueness, and the pullback connection/covariant derivative along a curve. Layer 1 now gives both Levi-Civita and along-curve differentiation explicit milestones and API obligations before defining the connection-based geodesic predicate.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `cfd84c5`. The metric API inventory now assigns `RiemannianBundle`, `IsRiemannianManifold`, and `EMetricSpace.ofRiemannianMetric` to `Riemannian/Basic.lean`, and assigns the fully qualified `Manifold.pathELength` and `Manifold.riemannianEDist` to `Riemannian/PathELength.lean`. Layer 0 now uses the qualified names and records the infimum-over-`C^1`-paths role of `Manifold.riemannianEDist`.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed across `cfd84c5` and `fbe4aae`. Layer 1 now defines `(a_p)` and pointwise completeness `(d_p)` and proves only `(a_p) ↔ (d_p)` by homogeneity. Global (d) is explicitly `∀ p, (d_p)` and is not inferred there. Layer 3 now states the global step separately and records the base-point propagation corollary: `(a_p)` at one point implies global (d), hence `(d_q)` and `(a_q)` for every `q`.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed across `cfd84c5` and `fbe4aae`. Levi-Civita now has its own owned milestone with metric compatibility, torsion-freeness, existence, and uniqueness; this roadmap delivers it in the shared manifold namespace unless mathlib4#36845 lands first. Layer 1 now also has an explicit geodesic-spray milestone: the vector field on `TM`, chart formula and chart independence, smoothness, and equivalence of its integral curves with velocity lifts of covariant geodesics. Maximal intervals remain a separate target tied to mathlib4#26413. Finally, the roadmap names the `NormedSpace.fromTangentSpace` identification for `T_0(T_p M)`, the derivative/strict-derivative theorem for `exp_p`, and an owned manifold inverse-function theorem before concluding local diffeomorphism.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `fbe4aae`. Layer 0 now consumes `Manifold.pathELength_add` and `Manifold.pathELength_comp_of_monotoneOn` directly. Constant-speed reparametrization is restricted to regular `C¹` curves, and lower semicontinuity explicitly assumes a uniformly convergent sequence with a `C¹` limit on a fixed compact interval. The substantive target is now a named corner-smoothing theorem proving equality of the piecewise-`C¹` and `C¹` infima. Mathlib's `C¹` `Manifold.pathELength`/`Manifold.riemannianEDist` convention is explicitly canonical downstream.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `fbe4aae`. Layer 2 now separates the Gauss identity, minimization against competitors staying in the normal ball, and the escape estimate needed to obtain the ambient `edist` identity on a smaller ball. Uniqueness is stated for the constant-speed affine representative and, more generally, only up to nondecreasing surjective reparametrization. The first-variation milestone now names the energy functional, variations and variation fields, both covariant-derivative directions, integration by parts, the variation formula, criticality, and the geodesic characterization, without claiming it drives Hopf-Rinow. Global `(f_p)` has been removed from Layer 2 and appears only in Layer 3 under `(a_p)`.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `e1a03b1`. Layer 3 now gives the complete directed graph explicitly: `(c) ⇒ (d) ⇒ (a_p) ⇒ (f_p)`, `(a_p) ∧ (f_p) ⇒ (b) ⇒ (c)`, and `(b) ⇔ (e_p)`. The `(c) ⇒ (d)` milestone now requires local norm equivalence, a convergent subsequence of lifted tangent states in a trivialization, a named finite-endpoint extension criterion for integral curves, and spray uniqueness. `(d) ⇒ (a_p)` is stated separately. I also replaced the nonexistent `ProperSpace.complete` with `complete_of_proper`, pinned `[T2Space M]`, wrote both directions of `(b) ⇔ (e_p)`, and added the arbitrary-center closed-ball step needed for `ProperSpace M`.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `e1a03b1`. The direct compactness argument on `TM` has been dropped; Corollary 2.9 now routes through `proper_of_compact`, `complete_of_proper`, and `(c) ⇒ (d)`. `IsClosed.completeSpace_coe` and `Metric.isCompact_iff_isClosed_bounded` were moved to the consume list, and the unspecified diameter item was removed. Layer 4 now owns separate metric-level APIs for curve length/length spaces and constant-speed geodesic spaces in `TauCeti/Topology/MetricSpace/Length.lean`, proves their relationship explicitly, and records coordination with the existing metric-geometry discussion before upstreaming.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `e1a03b1`. The flat example now consumes Mathlib's existing `IsRiemannianManifold 𝓘(ℝ, F) F` instance and carries no `NeZero` assumption. I dropped the sphere example because the required induced round metric is not in Mathlib or this roadmap. I also dropped the punctured-space alternative. The roadmap now owns restriction of a smooth Riemannian metric to open submanifolds, including identification of the induced distance on convex open subsets, and uses the open unit ball in `ℝ` at `p = 0` as one fully specified example: radial minimizers give `(f_p)`, `closedBall 0 2` is the noncompact whole space, and a radial geodesic reaches the missing boundary in finite time.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `e1a03b1`. I checked the 1992 do Carmo source numbering and replaced the broad ranges with exact loci: Ch. 1 Def. 2.9, Ch. 2 §§2–3, Ch. 3 §§2–4, Ch. 7 §2 Def. 2.4 and Thm. 2.8/Cor. 2.9, and Ch. 9 §2 Props. 2.4–2.5. In this edition Ch. 3 §1 is introductory material; the geodesic flow begins at Def. 2.1 in §2, minimizing properties are §3, and convex neighborhoods are §4. I added `references/do-carmo-hopf-rinow.md`, modeled on the Geometric Topology extracts, with a concise source map, theorem summary, and proof-dependency graph.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

TauCetiRoadmap/HopfRinow/README.md
AxelDlv00: Fixed in `e1a03b1`. The `Authorship` section has been removed from the roadmap; the AI-assistance disclosure remains a PR-level concern.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

AxelDlv00:
Addressed on the current branch, with the cross-roadmap ownership changes completed in `3f2dadd`.

1. **Ownership:** Hopf--Rinow now explicitly owns the Levi-Civita connection, interval-aware geodesics and their flow, exponential maps and local inverse logarithms, and Hopf--Rinow. Geometric Topology Layer 7 now consumes that connection and owns curvature and Riemannian volume. Optimal Transport Layer 7 now consumes the Hopf--Rinow exponential, logarithm, completeness, and minimizing-geodesic APIs and retains only the transport-facing geometry: injectivity radius, squared-distance semiconcavity, cut-locus measurability/nullity, and measurable/a.e. selections. Its numbered items 1--3 and cross-roadmap dependency section say this explicitly. Optimal Transport Layer 8 likewise consumes the shared metric length/geodesic-space API instead of rebuilding it.

2. **ODE regularity:** smooth dependence is no longer consumed from Mathlib. Hopf--Rinow Layer 1 explicitly targets `C^∞` dependence on time and initial data and the local geodesic flow, after constructing the smooth geodesic spray. Maximal intervals and the finite-endpoint extension criterion are also explicit targets, with the in-flight Mathlib PRs recorded only as work to consume if they land.

3. **Other dependencies:** Layer 0 now orders the distance construction as `EMetricSpace.ofRiemannianMetric`, global finiteness of `Manifold.riemannianEDist` under connectedness, and only then the compatible `MetricSpace`. The roadmap separately targets the pullback connection/covariant derivative along curves, the smooth geodesic spray and its integral-curve characterization, the manifold inverse-function theorem, and the identification `T_0(T_p M) ≃L[ℝ] T_p M`. It owns open-submanifold metric restriction; the sphere example was removed, so no embedded-submanifold metric is assumed. Layer 4 owns the shared metric length/geodesic-space API and now links the prior Evan Bailey Zulip proposal for coordination. Layer 2 also names the domain-aware local `log_p` API needed by Optimal Transport.

4. **`Suggested.lean`:** the universally quantified `IsGeodesic` and `expDomain` parameters have been removed. The file currently states only the honest Layer 0 length-distance target and explicitly defers geodesic-level signatures until Layer 1 supplies the real interval-aware geodesic, initial-data, maximal-domain, and exponential-map types.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

kim-em:
:robot: Claude+Codex say:

**1. Levi-Civita regularity is still owned by nobody, and it is load-bearing.** Layer 1 (lines
201-205) asks for existence and uniqueness of the torsion-free metric-compatible connection and
says "consume mathlib4#36845 if it lands first". But Mathlib separates a bare `CovariantDerivative`
from a `ContMDiffCovariantDerivative`, and
https://github.com/leanprover-community/mathlib4/pull/36845 "feat: the Levi-Civita connection on a
manifold" says in its own description: "Future PRs will prove that if `M` is `C^{k+2}` and `g` is
`C^{n+1}`, the Levi-Civita connection is a `C^n` connection." So even when #36845 lands you will
have the connection without its regularity, while the geodesic spray milestone at line 221 asks for
`S_g` to be `C^∞`, which needs smooth Christoffel coefficients. Add an explicit milestone: for a
`C^∞` metric on a `C^∞` manifold, the Levi-Civita connection is `ContMDiffCovariantDerivative ... ∞`,
with the chart-level smoothness of the Christoffel symbols, and order it before the spray.

**2. The exponential map has no basic API milestone.** Lines 243-257 define
`expDomain g p = {v | 1 ∈ J_g(p,v)}` and then immediately ask for `mfderiv exp_p 0`, a strict
derivative, and a local diffeomorphism, and Layer 2 asks for normal balls on top of that. Nowhere
does the roadmap ask for

- `IsOpen (expDomain g p)`,
- smoothness of `exp_p` on that domain,
- smoothness of the time-one evaluation extracted from the local flow.

Those are what carry you from "the flow depends smoothly on initial data" to any statement about
`mfderiv exp_p`, and none of them is a formality. Add an exponential-map basic API milestone with
domain openness, smoothness on the domain, the evaluation and homogeneity lemmas, and whatever
continuity a total junk-valued implementation needs where it is used.

**3. Layer 4's compact corollary: the statement is fine, the prescribed route is not.** Line 335
says "connectedness is unnecessary for this corollary". The corollary itself is indeed true without
connectedness. But the route it prescribes runs `proper_of_compact` then `complete_of_proper`, both
`PseudoMetricSpace` results, and Layer 0 (lines 189-195) only exposes an ordinary `MetricSpace`
after proving `Manifold.riemannianEDist I x y ≠ ∞` under `[ConnectedSpace M]`. Two disjoint round
spheres are compact with Riemannian distance `∞` between components, so there is no ordinary metric
inducing `riemannianEDist` to be proper in. Say "the corollary holds without connectedness, but
this proof route does not", and then pick one: assume `[ConnectedSpace M]` as do Carmo does, argue
componentwise, or state `(c) ⇒ (d)` at the extended-metric level, where `complete_of_compact`
(`Mathlib/Topology/UniformSpace/Cauchy.lean`) already gives completeness from compactness.

**4. `IsGeodesicCurveOn` is under-specified, and so is `g`.** Line 212 says "`γ` is continuous on
the parameter set `s`, and its velocity has vanishing covariant derivative there". Continuity does
not give a velocity, and differentiating the velocity through the along-curve connection needs the
velocity field itself to be differentiable, so the honest requirement is `ContMDiffOn ... 2` (or
smooth) on `s`, unless the predicate separately carries a differentiable section agreeing with `γ'`.
Pin it. Two riders: the initial-data bullet at line 217 needs `0 ∈ s`; and every geodesic-level
name here takes an explicit `g`, while the standing hypotheses supply the metric through the
`Bundle.RiemannianBundle` instance. Say which spelling is canonical.

**5. Line 91 understates what Mathlib's ODE library supplies, and that is my fault.** I told you in
round 1 that there is "no flow", and the line now reads "It does not yet supply `C^k` dependence,
flows, or maximal intervals." The middle item is wrong, and I retract it. Even the pinned revision
has local flows: `Mathlib/Analysis/ODE/ExistUnique.lean` documents
`IsPicardLindelof.exists_forall_mem_closedBall_eq_hasDerivWithinAt_lipschitzOnWith` ("the existence
of a local flow that is Lipschitz continuous in the initial point"), a continuous-on-its-domain
variant, and `ContDiffAt.exists_eventually_eq_hasDerivAt` ("a `C¹` vector field admits a local
flow"). Narrow the line to `C^k` dependence on initial data and maximal intervals, which are the
real gaps.

**6. Metric compatibility has landed upstream since the pin.** Lines 100-101 and 201-202 list
metric compatibility as work to define here. It is now `CovariantDerivative.IsMetricCompatible` in
`Mathlib/Geometry/Manifold/VectorBundle/CovariantDerivative/Metric.lean` on master, from merged
https://github.com/leanprover-community/mathlib4/pull/36299. It is genuinely absent from the
revision this repository pins (that tree has only `Basic.lean` and `Torsion.lean`), so the roadmap
is not wrong today, but it will be wrong at the next dependency bump. Say so explicitly: consume
`IsMetricCompatible` after the bump, and keep only existence, uniqueness and regularity of
Levi-Civita as owned work, rather than a lasting duplicate predicate.

**7. "Transport of geodesic completeness across isometries" (line 346) needs its adjective.**
Mathlib has no Riemannian-isometry API here, and a bare metric `IsometryEquiv` that preserves
connection-defined geodesics is Myers-Steenrod. Say "smooth Riemannian isometry", define that
notion, and get preservation of the Levi-Civita connection and of geodesics from uniqueness. If you
really do mean metric isometries, Myers-Steenrod has to become an owned milestone.

**8. Non-blocking, your call.** Dropping the sphere was one of the two options I offered, so this
is not a change I am asking for. But it is worth noticing what the acceptance criteria now cover: a
flat inner-product space and the open unit interval in `ℝ`. Those do exercise affine geodesics,
completeness and an incomplete maximal geodesic, but nothing curved and nothing compact, so most of
Layer 2 goes untested. I withdraw the conformal-metric replacement I was going to suggest here: it
is not as cheap as it looks (it needs the rescaled metric constructed, its connection computed, and
its induced distance identified), and hyperbolic balls in the standard models are Euclidean balls
as sets anyway, so it would not even give an interesting normal ball. The cheap version is to
extend the Euclidean example to compute `exp_p`, `expDomain`, `log_p`, a normal ball and the Gauss
identity. Still flat, but it exercises the API wiring rather than only the statements.

Minor: at line 314, "identify each closed ball centered at `p` with the image under `exp_p`" reads
as a claim of injectivity. The statement wanted is the set equality
`closedBall p r = exp_p '' closedBall 0 r`; `exp_p` is generally not injective there.

One more retraction: my proposed change of the geodesic-flow locator from Ch. 3 §2 to §1 was wrong.
§1 is introductory in this edition, and your §2 is correct.

AxelDlv00:
Thanks for the detailed review. I have pushed the fixes and synced the branch with current `main`.

Current head: [`15bbf793a08c334d304ef2a494feec49d43b39ad`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/15bbf793a08c334d304ef2a494feec49d43b39ad)

Point-by-point:

1. **Levi-Civita regularity.** [`e10215d001f5e9b3f1c9667c0cccbebd8501a3f2`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/e10215d001f5e9b3f1c9667c0cccbebd8501a3f2) adds the explicit `ContMDiffCovariantDerivative ... ∞` milestone, smooth chart-level Christoffel maps and scalar coefficients, and orders it before the spray.

2. **Exponential-map basic API.** The same commit adds openness of `expDomain`, smoothness on the domain, time-one flow evaluation, homogeneity, star-shapedness, and the continuity consequences needed for a total junk-valued implementation.

3. **Compact completeness route.** [`f63fe15a82508bf14c9abaff866947623c543fba`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/f63fe15a82508bf14c9abaff866947623c543fba) now says explicitly that the corollary is true without connectedness, while the ordinary-metric `proper_of_compact` then `complete_of_proper` route requires `[ConnectedSpace M]`.

4. **Geodesic predicate and metric spelling.** [`669499ff9a89b42e41b98d4bb7a26dcb25932d39`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/669499ff9a89b42e41b98d4bb7a26dcb25932d39) makes the Riemannian bundle instance the canonical implicit metric, requires `C²` regularity, and includes `0 ∈ s` in initial data. [`5f327f2fa51381b2a6b54550122fe9894a01b998`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/5f327f2fa51381b2a6b54550122fe9894a01b998) further pins `UniqueDiffOn ℝ s`, specifies the `mfderivWithin` velocity, and states constant speed on preconnected domains or componentwise.

5. **ODE inventory.** [`bb839bc589d212d164bffcac67b88561d9592617`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/bb839bc589d212d164bffcac67b88561d9592617) narrows the Mathlib gap to `C^k` dependence on initial data and maximal intervals; local flows and the existing Lipschitz, continuous, and `C¹` results are now listed as available.

6. **Metric compatibility.** [`eb7ebb8d8bcb196e474f6038c3e5740e931ea420`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/eb7ebb8d8bcb196e474f6038c3e5740e931ea420) records that `CovariantDerivative.IsMetricCompatible` is available after the dependency bump and leaves existence, uniqueness, and regularity as the owned work.

7. **Isometry transport.** [`ca56823f27493046da9a1bdf73ce8b477ce007a0`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/ca56823f27493046da9a1bdf73ce8b477ce007a0) now scopes this to smooth Riemannian isometries and derives connection and geodesic preservation by uniqueness, with Myers-Steenrod explicitly separate.

8. **Acceptance example.** [`1bdd9c6f214615f30d5d5404ea4bc592d2160480`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/1bdd9c6f214615f30d5d5404ea4bc592d2160480) extends the flat example through `exp`, `expDomain`, `log`, a normal ball, and the Gauss identity. The sphere remains intentionally omitted.

Minor ball-image wording: [`8bbcc3f2356b95eec078de3844ea5017d2066dfa`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/8bbcc3f2356b95eec078de3844ea5017d2066dfa) states the requested set equality without implying injectivity. The Chapter 3 §2 geodesic-flow locator was left unchanged, as requested.

Two final roadmap-policy fixes are in [`35fb2a546a45887b33d93107e1e47d777578d9b0`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/35fb2a546a45887b33d93107e1e47d777578d9b0): Mathlib proposals are now treated as API-shape references rather than dependencies to wait for, and every target is explicitly owned and delivered in Tau Ceti, with later adoption of equivalent Mathlib declarations where applicable.

The branch was merged with current `main` in the head commit above. Fresh CI is green: `build` and dependency resolution both passed.

kim-em:
:robot: Claude says:

All eight points check out, including the two I was most worried about. The
`ContMDiffCovariantDerivative ∇ ∞` spelling matches Mathlib's actual signature, the homogeneity
equivalence `t ∈ J(p,v) ↔ t • v ∈ expDomain p` is correct at `t = 0` as well, and the ball-image
set equality holds in both directions. Restricting constant speed to preconnected `s` was a real
correctness fix that I had not asked for, and pinning `UniqueDiffOn ℝ s` with `mfderivWithin` is a
better answer to the regularity question than the one I proposed.

One thing to put back. `35fb2a5` rewrote the prior-art paragraph to

> These proposals determine API shape but are not Tau Ceti dependencies. Implement every missing
> milestone in Tau Ceti without waiting [...]

and in doing so dropped "coordinate with their authors". That goes a step further than the rule it
is conforming to. The repository README, under *Porting existing work*, still says:

> **Coordinate first.** Work with the authors of the existing material and obtain their agreement
> before integrating it. [...] A roadmap that independently develops the same mathematics should
> still cite the existing work and coordinate where possible, to avoid needless duplication or
> incompatible design choices.

"Never push work to Mathlib" removes the destination, not the coordination. Layer 1 now duplicates
four in-flight Mathlib PRs written by other people (mathlib4#36845, #26413, #26394, #40062), which
is exactly the "needless duplication or incompatible design choices" case that rule is about. The
section is still headed "Prior art and coordination", and after this commit the only coordination
instruction left in the file is the Evan Bailey one in Layer 4.

Please restore a clause asking implementors to coordinate with those authors on API direction,
while keeping the new and correct point that nothing here waits on them.

AxelDlv00:
Fixed in [`4980b67`](https://github.com/AxelDlv00/TauCetiRoadmap/commit/4980b67): the paragraph now asks implementors to coordinate API direction with the authors of the cited Mathlib PRs where practical, while keeping the point that Tau Ceti should implement the missing milestones without waiting for those PRs to land.

🤖 Prepared with the assistance of Codex (GPT-5.6-Sol).

kim-em (approved):
:robot: Reviewed and repaired at `662c08f0671d4d1c14608d6d63660e265cae20c4`; the remaining PR-history prose is now timeless, the branch includes current `main`, and no blocker remains. Approving.

PR #102, opened 2026-07-29
feat(AdoIwasawa): add arbitrary-field roadmap

### LINKED ISSUE #101: [Intention]: add the Ado–Iwasawa roadmap
### Roadmap area

RepresentationTheory

### Items in scope

Author a new Ado–Iwasawa roadmap that:

- specifies the finite-dimensional faithful-representation theorem over an arbitrary field;
- builds the reusable universal-enveloping-algebra, nilrepresentation, derivation-stability, and positive-characteristic central-element machinery needed by the proof;
- reconciles the existing PBW and Lie’s-third-theorem dependencies in the representation-theory roadmap family; and
- provides typechecked suggested Lean target signatures.

### Notes

Roadmap authorship only; this claim does not claim implementation targets in the Tau Ceti code repository. The proposal will cite the existing Etingof formalization’s statement/reduction API and the active Mathlib PBW work, without assuming either supplies the missing proof.

kim-em:
This PR adds a roadmap for the Ado–Iwasawa theorem: every finite-dimensional Lie algebra over an
arbitrary field has a faithful finite-dimensional representation, together with Hochschild's
strengthening that the representation may be chosen so that every element acting nilpotently in the
adjoint representation acts nilpotently in it. The roadmap builds the theorem as a reusable theory of
finite-dimensional Lie representations rather than racing to the existential statement, and keeps the
two characteristic regimes on genuinely separate proof tracks: derivation-stable cofinite ideals of
enveloping algebras, split extensions and Levi decomposition in characteristic zero; central
`p`-polynomials, module-finiteness over the commutative algebra they generate, and the Krull
intersection theorem in characteristic `p`. The tracks meet only after each has produced the same
faithful finite-target interface, and neither the statement nor the summit carries a characteristic
or algebraic-closure hypothesis.

The convention that most affects contributors is the nilradical. Fulton–Harris and Hochschild both
define the nilrepresentation condition against the largest nilpotent ideal, and Mathlib has no such
declaration: `LieAlgebra.maxNilpotentIdeal` is the largest ideal on which `L` acts nilpotently, which
is a different and in general smaller ideal (for the two-dimensional nonabelian algebra it is `⊥`
while the nilradical is one-dimensional). Layer 0 therefore builds `nilradical` and its basic theory
as general Lie theory wanted independently of Ado, and the roadmap warns against the Mathlib
declaration by name so that nobody reaches for it and proves something vacuous.

Three further conventions are pinned up front, because getting them wrong is expensive later. Ideals
of `U(L)` are `Ideal` carrying `Ideal.IsTwoSided` rather than the separate `TwoSidedIdeal` type,
which is the side of the choice with the powers, `FG` and quotient-ring API the positive-characteristic
track uses. The split hypothesis of the extension layer is stated externally, through Mathlib's
`LieAlgebra.SemiDirectSum`, with a once-and-for-all lemma taking internal complementary data to that
form. And the passage from the commutative Krull intersection theorem to powers of a two-sided ideal
of `U(L)` goes through a single named identity, `(UC)^n = J^n • ⊤`, so that no coercion has to be
trusted at the point of use.

Where Mathlib already has what is needed, the roadmap says so and applies it:
`Ideal.mem_iInf_smul_pow_eq_bot_iff` is the generalized Krull intersection theorem Hochschild cites,
in exactly the generality wanted, so no Tau Ceti file restates it. Where a Mathlib hypothesis bites,
the roadmap names it: Lie's theorem carries `LieModule.IsTriangularizable`, which does not hold over
`ℝ` or `ℚ`, so scalar extension and descent for `nil`, `radical` and Lie derivations is an explicit
milestone rather than a silent step.

The PBW project stays owned by
[the highest-weight roadmap](https://github.com/TauCetiProject/TauCetiRoadmap/blob/main/TauCetiRoadmap/RepresentationTheory/LieHighestWeight/README.md),
whose PBW unit is now stated over an arbitrary field with no characteristic or closure hypothesis and
tracks the injectivity, domain and Noetherian consequences that Ado–Iwasawa consumes over `𝔽₂`. The
real specialization is exported in the form the
[Lie-groups roadmap](https://github.com/TauCetiProject/TauCetiRoadmap/blob/main/TauCetiRoadmap/RepresentationTheory/LieGroups/README.md)
consumes for Lie's third theorem, and that roadmap's previously unowned reference to "Ado's theorem"
now points here.

Closes https://github.com/TauCetiProject/TauCetiRoadmap/issues/101 ([Intention]: add the Ado–Iwasawa roadmap)

🤖 Prepared with Codex and Claude Code


CBirkbeck (changes requested):
:robot: **Disclosure:** This review was written by ChatGPT 5.6 Sol on behalf of Chris (CBirkbeck).

**Changes requested at `52aad658a880f93ab916ab64ec327795b533f963`.** I checked the earlier findings against the current head and have limited this review to the remaining mathematical blockers and major dependency gaps.

1. **The nilradical is still incorrectly declared derivation-stable over an arbitrary field.** Layer 0 says every `LieDerivation K L L` preserves `nil L`. Automorphism invariance is characteristic-free, but derivation invariance is not: in positive characteristic it needs additional hypotheses. Keep only automorphism invariance in Layer 0, and move a theorem such as `lieDerivation_nilradical_le` to the characteristic-zero track (or state the exact modular nilpotency-class bound). Layer 3's planned theorem `D (radical K L) ≤ nil L` can supply the characteristic-zero result. Avoid using “characteristic” ambiguously here. See [Maksimenko](https://admjournal.luguniv.edu.ua/index.php/adm/article/view/770).

2. **The characteristic-zero flag still does not satisfy Layer 4's hypotheses.** Layer 5 constructs an arbitrary codimension-one flag from `Z(L)` to `radical K L`, while Layer 4 preserves the nilrepresentation only when the larger algebra is nilpotent or the two terms have the same nilradical. Set `N = nil L` and `R = radical K L`, and build
   `Z(L) ⊂ ⋯ ⊂ N ⊂ ⋯ ⊂ R`.
   Below `N`, make every term nilpotent. Above `N`, prove `[L,R] ≤ N`, use this to make every intermediate subspace an `L`-ideal, and prove each intermediate algebra has nilradical exactly `N`. Those are the lemmas that justify every invocation of Layer 4.

3. **Levi decomposition remains a hidden major sub-roadmap.** The roadmap asks for Levi–Malcev in one bullet, but neither Mathlib nor another named Tau Ceti supplier provides it, and no dependency-closed proof route is specified. Choose and spell out a non-circular route—cohomology plus the required Whitehead lemmas, or a precise direct induction—and export the semidirect-product existence theorem Layer 4 consumes. The proof must not use Ado or linear structure theory derived from Ado. Conjugacy of complements is not needed for the existential representation and should be removed unless a named consumer requires it.

4. **The derivation-stable cofinite-ideal refinement is still under-specified.** The paragraph defining `B/I` and `J = B^m` needs separate load-bearing milestones proving: the lifted derivation has range in the nilradical-generated ideal; that ideal and its powers are stable; a sufficiently large power lies in `I`; the quotient by that power is finite-dimensional; and every nilradical element acts nilpotently in the quotient. Suggested targets such as `liftedDerivation_range_le_nilIdeal`, `nilIdeal_pow_stable`, `nilIdeal_pow_le`, `finiteDimensional_quotient_nilIdealPow`, and `nilradical_isNilpotent_mod_nilIdealPow` would make the construction reconstructible.

5. **The Hochschild assembly still makes an unjustified jump.** “Apply the extension lemma to `s` and `nil L` on `L` to get `r ∈ nil L`” does not follow from that lemma: it only controls the action of the Lie algebra spanned by `s` and the nilradical. The roadmap must add `[L,R] ≤ N`, the nested extension/sum argument proving `ad(r)` nilpotent from `ad(x)`, `ad(s)`, and `[x,s] ∈ N`, and the radical criterion
   `r ∈ R ∧ IsNilpotent (ad r) → r ∈ N`.
   Only then does the final module application show that `x=s+r` acts nilpotently. Also pin the scalar-extension/descent step for solving `[s,t]=s` over the original field.

6. **The PBW supplier contract still does not match what Ado consumes.** Ado says the highest-weight roadmap owns every item in Layer 1, but that supplier currently omits compatibility for general Lie maps and quotients and, most importantly, the monic-central-relation module-finiteness theorem used in Layers 3 and 6. Move the exact general-field targets to the supplier (or explicitly make Ado own them), including the filtration/graded equivalence, ordered basis, required functoriality, filtered domain/Noetherian transfer, and finite spanning from monic central relations. The consumer should then cite an exact contract rather than carry the only complete specification.

These points remain blocking under the repository's dependency-closed, no-gaps standard.

CBirkbeck:
Sorry I don't know much about this area, but ChatGPT pro said it found a few errors.

CBirkbeck (changes requested):
Reviewed at `7d27245a9c9dfc27ddeb4da8b161aeb5072e8645`. The six blockers from the previous round are
addressed: derivation invariance of the nilradical is confined to characteristic zero, the flag now
passes through `nil L` with nilpotent steps below it and common nilradical above it, Levi has a
noncircular induction with the complete-reducibility input named, the stable-ideal and Hochschild
layers are decomposed into checkable milestones, and `LieHighestWeight` now owns the general-field
PBW contract. One target is still stated more generally than the argument that backs it.

## `exists_derivationStable_cofiniteIdeal` is missing solvability

`AdoIwasawa/Suggested.lean:124` states the refinement for an arbitrary finite-dimensional `L`:

```lean
theorem exists_derivationStable_cofiniteIdeal [CharZero K] [FiniteDimensional K L]
    (I : Ideal (UniversalEnvelopingAlgebra K L)) [I.IsTwoSided] ...
```

Every milestone in the Layer 3 chain this packages (`README.md:308-327`) is stated for a solvable
`S`, and has to be. The load-bearing step is *Lifted derivations have range in `B`* (`README.md:316`),
which gets `Dᵁ(U(S)) ≤ B` from `D(S) ≤ nil S`. What the previous bullet actually proves in
characteristic zero is `D (radical K L) ≤ nil L`; `D L ≤ nil L` follows only when `L = radical K L`.
It genuinely fails otherwise — for `L` semisimple and `D = ad x` one has `nil L = ⊥` and `D L ≠ ⊥` —
so for general `L` the ideal `B` generated by the image of the nilradical need not contain the range
of `Dᵁ`, and `Dᵁ(B^m) ≤ B^m` does not follow. Reichstein's notes state the lemma for finite-dimensional
solvable `S` for exactly this reason.

I am not claiming the general statement is false; for semisimple `L` every derivation is inner and
every two-sided ideal is stable under `ad ι(x)`, so `J = I` works there. The point is that the type is
broader than the proof plan under it, and broader than what Layer 4 consumes.

Fix: add the hypothesis,

```lean
theorem exists_derivationStable_cofiniteIdeal
    [CharZero K] [FiniteDimensional K L] [LieAlgebra.IsSolvable L] ...
```

Note the spelling: there is no `LieRing.IsSolvable`; solvability is still `LieAlgebra.IsSolvable L` on
the pin, even though nilpotency has moved to `LieRing.IsNilpotent`. Since Layer 4 consumes this for a
solvable ideal `S` sitting inside a larger `L`, it would read better to give the section its own
variable `S` rather than reusing `L`. If the arbitrary-`L` statement is the intended endpoint, it needs
a different proof; the one in the README does not reach it.

This is the only statement I found whose type is broader than its argument.

## Three hardening changes for the same commit

**Pin the Levi interface that Layer 4 actually consumes.** `Suggested.lean:67` exports only the
vector-space complement, and its own docstring says the definitive form is the external
`SemiDirectSum` equivalence. Seed that instead of (or beside) the internal form, so the interface
typechecks rather than being described:

```lean
theorem exists_leviDecomposition [CharZero K] [FiniteDimensional K L] :
    ∃ (S : LieSubalgebra K L)
      (ψ : S →ₗ⁅K⁆ LieDerivation K (LieAlgebra.radical K L) (LieAlgebra.radical K L)),
      LieAlgebra.IsSemisimple K S ∧
        Nonempty (L ≃ₗ⁅K⁆ (LieAlgebra.radical K L) ⋊⁅ψ⁆ S)
```

The coercions on `radical` will need adjusting; the point is that Layer 3 → Layer 4 is where this
roadmap is most likely to lose a hypothesis, and that seam should be a Lean signature.

**Give `⁅s, t⁆ = s` an elementary proof.** `README.md:420` calls it the easy half of `sl₂`-triple
existence and routes it through `exists_isSl2Triple_of_weight_isNonZero` plus base-change descent,
which makes Jacobson–Morozov look like a prerequisite and burns one of the two uses of the Layer 3
range-membership descent. Over the original field: if `z ∈ ker (ad s)` then `ad z` commutes with the
nilpotent `ad s`, so `ad s ∘ ad z` is nilpotent and `κ(s, z) = 0`; hence `s ∈ (ker (ad s))ᗮ`, which is
`range (ad s)` by invariance, nondegeneracy of the Killing form and rank-nullity. That gives `t` with
`⁅s, t⁆ = s` directly, with no algebraic closure in sight.

**Pin the PBW supplier in Lean, not only in prose.** `LieHighestWeight/README.md` now names the right
contract, but `LieHighestWeight/Suggested.lean` has no declaration for the filtration and ordered
monomials, the filtered-to-graded equivalence, functoriality and quotient compatibility, domain and
Noetherian transfer, or the central-monic-relation `Module.Finite` theorem — its only enveloping-algebra
content is the Casimir and the `U(L)`-module structure. Meanwhile `AdoIwasawa/Suggested.lean` imports
`Mathlib` alone and restates `ι_injective` locally at line 101, so nothing checks that the two
roadmaps agree. Put the load-bearing PBW signatures in the supplier's `Suggested.lean` and import that
module here. This matters more than usual while the Mathlib PBW PR is an open draft rather than a
merged supplier.

Everything else in both characteristic tracks reads correctly to me, including the weighted
lower-central-series truncation for the nilpotent case, the `nil I = nil L ⊓ I` identity and its use in
the flag, and the `p`-polynomial / Krull-intersection route in positive characteristic. Add the
solvability hypothesis and I will approve.

CBirkbeck (commented):
## Review by ChatGPT Pro

Reviewed `88064edf9ccc6852ef9eb58883f613e812148a63`, both Ado–Iwasawa files, all four companion-file changes, and the previous review discussions. This is a mathematical and source-level interface review, not a local Lean build.

The previous mathematical/dependency requests are addressed. In particular, the cofinite refinement now explicitly assumes solvability; the flag passes through the nilradical; the Levi proof has an independent complete-reducibility supplier; the `[s,t]=s` argument is elementary and works over the original characteristic-zero field; and the positive-characteristic argument keeps the commutative central subalgebra distinct from the two-sided ideal in the enveloping algebra. I found no further mathematical blocker in those two proof tracks.

There is one focused improvement needed in the **new PBW supplier interfaces**, rather than another change to the Ado proof.

### Identify the advertised PBW basis with the actual ordered monomials

**Location:** `LieHighestWeight/Suggested.lean`, `nonempty_basis_pbwGraded` and `pbwBasis`.

The new signature

```
pbwBasis (b : Basis κ K L) : Basis (κ →₀ ℕ) K (UniversalEnvelopingAlgebra K L)
```

provides an indexed basis but no equation relating its entries to the ordered products of `ι (b i)`. Likewise, `nonempty_basis_pbwGraded` states only existence of a basis with the cardinality of degree-`n` monomials, not that their canonical images form that basis. Its docstring's claim that every other target follows from this statement is therefore too strong: the statement records graded vector-space dimensions, not the canonical graded multiplication or the symmetric-algebra comparison.

This distinction already appears for the one-dimensional abelian Lie algebra over `ℚ`. In `U(L) = ℚ[X]`, the vectors `2^m X^m` form a basis indexed by `m : ℕ`, just as `X^m` do. Both satisfy the current basis type, but only the latter is the prescribed ordered-monomial basis for the chosen generator. The consumer cannot recover the missing normalization from the type.

Please add a representative value theorem saying that `pbwBasis b m` is the ordered product of the powers `(ι (b i))^(m i)`, with the empty-product and singleton cases exposed. For the graded statement, specify that the basis consists of the classes of those same monomials, or expose the canonical graded-algebra isomorphism from the symmetric algebra with its generator equation. This is the equality needed by weighted truncation, monic-relation reduction, and the advertised functoriality—not merely an implementation convenience.

Acceptance tests should calculate the empty monomial as `1`, a singleton as `ι (b i)`, and recover the ordinary polynomial monomial basis in the one-dimensional abelian case. The prose already asks for the right PBW theorem; the requested change makes the newly added shared Lean interface express it too.

CBirkbeck (commented):

**Assessment: ready for human approval on mathematical and dependency content.** I found no remaining blocking issue in the revised Ado–Iwasawa proof plan or its shared PBW contract. This is a comment review, not an approval or a claim that the planned mathematics is already implemented.

I checked the current two characteristic tracks, the companion supplier changes, and the previous reviews. In particular:

* The nilradical is the largest nilpotent Lie ideal, not Mathlib's differently defined `maxNilpotentIdeal`; derivation invariance is confined to characteristic zero. The flag through the nilradical now supplies the actual hypotheses of the split-extension theorem at every step.
* The stable-cofinite refinement is stated for solvable S. Its construction distinguishes the ideal in U(S)/I from its preimage B, proves B^m⊆I, obtains cofinite quotients using ordered monic relations, and preserves nilpotence after refinement. This matches the scope of the characteristic-zero lemma in Asgarli's exposition following Fulton–Harris.
* Levi decomposition has an Ado-independent quotient induction and explicitly consumes the arbitrary-characteristic-zero complete-reducibility contract. The Hochschild assembly includes the radical criterion and nested nilpotent-extension argument. The Killing-form argument producing [s,t]=s avoids a hidden Jacobson–Morozov dependency.
* The characteristic-p track keeps the commutative central algebra R distinct from the two-sided ideal UC in U(L). The identity `(UC)^n = J^n • ⊤` is exactly what makes the existing commutative Krull theorem applicable. Augmentation gives 1−r≠0, PBW supplies the domain property, and finite dimensionality of L supplies separation at a finite stage. Using the full central augmentation ideal gives the strengthened nilpotence conclusion, not just bare faithfulness.

The most recent PBW review is now addressed in the actual signatures: `pbwMonomial` is an explicit ordered product; `pbwBasis_apply` and `basisPbwGraded_apply` identify the bases with those monomials and their canonical graded classes; the empty/singleton equations and one-dimensional polynomial test pin the normalization. The supplier owns the arbitrary-field monic-central-relation module-finiteness statement. I would not repeat the earlier basis-normalization objection against this head.

Two nonblocking wording improvements:

1. In the `pbwMonomial` docstring, qualify the alternative basis `2^m X^m` with `K=ℚ` (or characteristic different from 2). This is only an illustrative example, but the surrounding section deliberately works over arbitrary fields, including 𝔽₂.
2. In Layer 3, once `Dᵁ(U(S))⊆B` has been proved, `Dᵁ(B)⊆B` follows immediately. The warning should say that stability of B cannot be assumed **before** this range argument, rather than saying not to assert it at all. Also describe composition of derivations as composition in `Module.End`; derivations are closed under commutators, not arbitrary composition.

Reference checked for the characteristic-zero extension construction: S. Asgarli, *Ado's Theorem*, Lemma 1 and Proposition 2, https://personal.math.ubc.ca/~reichst/Ado%27s-Theorem.pdf. The positive-characteristic argument is explicitly laid out in the roadmap rather than treated as a consequence of the characteristic-zero theorem.

No local Lean build was run. CI and any remaining repository-process requirements should be checked independently before merging.

CBirkbeck (approved):

PR #106, opened 2026-07-30
feat: auto-merge and announce generated progress reports

kim-em:
This PR lets per-roadmap STATUS.md and PROGRESS.md reports land without a
human review, and announces each new PROGRESS.md section in the Tau Ceti >
Progress logs Zulip topic.

Both workflows delegate to SHA-pinned reusable workflows in
TauCetiProgress, the same discipline TauCeti uses for TauCetiReview: an
unpinned ref would let an upstream change alter what may merge here
without review. Each pin appears twice, once as the `uses:` ref and once
as `progress_ref`, and the two must match, since the first selects the
workflow and the second selects the validator code inside it.

The gate matters because this repository's ruleset cannot scope its
code-owner requirement to a path, so the only route to an unattended merge
is the roadmap App's bypass, which bypasses status checks too. The
validation is therefore the whole gate: provenance by immutable numeric
user id and a non-fork progress/* branch, exactly the two generated files
for one roadmap, a byte-exact append that continues the log, no symlinks,
a green build on that exact commit, and a merge bound to the validated
head. It never checks out pull request content.

Announcing is a separate workflow on push, so the App token and the Zulip
key never share a job, and a lost announcement can be retried without
re-merging; posts are idempotent on a stable per-window id.

The README records that these two files are machine-owned and that their
prose is not security-validated: the gate proves their shape, not their
truth.

🤖 Prepared with Claude Code

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/progress-announce.yml
+++ b/.github/workflows/progress-announce.yml
@@ -11,22 +11,22 @@ on:
   push:
     branches: [main]
     paths:
-      - '**/PROGRESS.md'
+      # Scoped to the directories a generated report can touch. A bare `**/PROGRESS.md` also matches
+      # a file at the repository root or in any unrelated directory, so a hand-written one would
+      # trigger an announcement attempt for something that is not a report.
+      - 'TauCetiRoadmap/**/PROGRESS.md'
+      - 'Completed/**/PROGRESS.md'
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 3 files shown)


CBirkbeck (dismissed):
LGTM!

PR #120, opened 2026-07-31
roadmap: weight-two forms on quaternionic inner forms of PGL2

kim-em:
## What changed

- replace the broad `GL_n` and arbitrary-cohomological programme with a roadmap for parallel-weight-two forms on quaternionic inner forms of `PGL₂`
- cover every quaternion algebra over a totally real field: split `M₂(F)`, mixed real ramification, and totally definite cases
- implement `PGL₁(D)` through central-invariant `Dˣ` forms, with a proved quotient comparison
- pin the two real infinity types: the full weight-two discrete-series representation of `PGL₂(ℝ)` at split places and the trivial representation of `PGL₁(ℍ)` at ramified places
- include cuspidal and one-dimensional eigenpackets, with the constant totally definite form and `1 ⊕ χ_cyc` as the normalization regression
- pin the good-place operator and polynomial to `T_v` and `X² - a_v X + N(v)`, using arithmetic Frobenius
- retain the quaternionic, adelic, rank-two cusp-integral, Hecke-choice-independence, and local-Galois prerequisites needed to make the endpoint an honest statement
- rename the directory and namespace to `WeightTwoAutomorphicForms`, and update the merged Modular Forms roadmap so its handoff supplies only the `n = 2` double-coset comparison consumed here

## Boundary

This PR stops at the attached-system declaration. It does not prove the declaration or develop general `GL_n`, arbitrary weights, general Harish--Chandra or relative-cohomology theory, general Satake theory, Jacquet--Langlands, Whittaker models, trace formulae, or Eisenstein series. A broader automorphic-representations roadmap can be written separately after this focused roadmap merges.

Tau Ceti does not import FLT. The roadmap instead gives an ordered, licence-aware migration map for the existing totally definite quaternionic and Galois material.

## Validation

- rebased onto current `origin/main` and resolved the roadmap-index conflict while retaining the merged Modular Forms entry
- `lake build TauCetiRoadmap.WeightTwoAutomorphicForms.Suggested`
- `lake build` (8,533 jobs)
- `git diff --check`
- repository audit finds no `TauCetiRoadmap.AutomorphicRepresentations` reference, no use of the private target name, and no Lean import of FLT


TauCetiRoadmap/AutomorphicRepresentations/README.md
```diff
@@ -0,0 +1,733 @@
...
+`SIGN-OFF` again at the layer where they arise.
+
+- **SIGN-OFF A — cuspidality and one-dimensional representations.** The Zulip formulation says
+  “cohomological automorphic representation,” without saying “cuspidal.” For split `D`, the usual
+  theorem takes cuspidal representations and excludes Eisenstein constituents. For division `D`,
+  there are no proper `F`-parabolics, so the standard constant-term definition of cuspidality is
+  vacuous and still includes the characters `χ ∘ Nrd`. The provisional summit therefore requires
+  both cuspidality and “not a reduced-norm character.” A reviewer should confirm that restriction or
+  specify the larger statement and its reducible systems for characters/Eisenstein constituents.
+- **SIGN-OFF B — cohomological normalization.** For even-rank groups, the passage between
```
CBirkbeck: "For even-rank groups" overstates: the C-algebraic/L-algebraic half-Tate shift is the `GL_n` statement — the twist `π ⊗ |det|^{(n−1)/2}`, half-integral exactly when `n` is even — not a fact about arbitrary even-rank reductive groups. Suggest: "For `GL_n` with `n` even (and its inner forms), the C-algebraic and L-algebraic normalizations differ by the half-integral twist `|det|^{(n−1)/2}` …" — Buzzard–Gee (already in the references) states it this way.

TauCetiRoadmap/AutomorphicRepresentations/README.md
```diff
@@ -0,0 +1,733 @@
...
+undefined mathematical nouns or vacuous predicates.
+
+> **Target statement.** Let `F` be a totally real number field, let `D` be a quaternion
+> algebra over `F`, and let `π` be a cuspidal cohomological automorphic representation of
+> `Dˣ(𝔸_F)` which is not a character factoring through reduced norm. There are a number field `E`,
+> Hecke polynomials `P_v ∈ E[X]` at all but finitely many
+> finite places `v` of `F`, and a semisimple good-place compatible family of continuous
+> two-dimensional representations
+> `ρ_{p,φ} : Gal(F̄/F) → GL₂(Q̄_p)`, indexed by primes `p` and embeddings
+> `φ : E → Q̄_p`, such that, whenever `v` is good and its residue characteristic differs from
```
CBirkbeck: "good" appears in the summit statement before it is defined (Layer 6 defines the bad set much later). Since the `N(v)` in `X² − a_vX + N(v)s_v` is convention-ambiguous precisely at this point, suggest the summit carry a named `IsGoodPlace` — `v ∉ S`, `D_v` split, `π_v` unramified, with `v ∤ p` at the use site — so the statement is self-contained.

TauCetiRoadmap/AutomorphicRepresentations/README.md
```diff
@@ -0,0 +1,733 @@
...
+
+At every split unramified finite place `v`, construct the spherical Hecke algebra of compactly
+supported, bi-`K_v`-invariant functions with convolution. Normalize Haar measure by `vol(K_v)=1`.
+Develop:
+
+- locally constant compactly supported functions, closure and integrability of convolution, the
+  algebra laws, and comparison of convolution with the finite double-coset sum acting on smooth
+  representations and global automorphic forms;
+- finiteness of `K_v\K_vgK_v`, characteristic functions of double cosets, the `GL₂` Cartan
+  decomposition, the spherical Hecke algebra calculation, and the normalized Satake transform;
```
CBirkbeck: Pin here that the **normalized** Satake transform requires a chosen square root of `q_v`, and that the characteristic functions of the double cosets of `diag(ϖ,…,ϖ,1,…,1)` correspond to scaled elementary symmetric polynomials, not the bare `eᵢ`. The unnormalized `T_v, S_v` route the summit uses is unaffected, but a contributor reaching for "the" Satake isomorphism will otherwise adopt a convention tacitly.

CBirkbeck (commented):
*(Review posted by Fable, Chris Birkbeck's AI assistant, at Chris's request; revised and expanded. Sources were verified against zbMATH and publisher records; the construction routes and the list of assumed theorems below were each put through independent adversarial model-review rounds, and the corrections from those rounds are incorporated.)*

The roadmap is exceptionally careful about logical dependencies: the dependency table, the three explicit SIGN-OFFs, the construction ensuring the summit statement is nonvacuous, the refusal to substitute proxy propositions, and the licence-aware account of migration would all be useful models for other roadmaps.

## Principal suggestion

This should become the roadmap for **automorphic representations of inner forms of GL_n over a number field** — the unit groups `G = Aˣ` of central simple algebras `A` of degree `n` over `F` — with split `GL_n` as the principal case and quaternion algebras (`n = 2`) as the worked example. The present summit statement, Galois representations attached to cuspidal cohomological automorphic representations of `Dˣ`, remains unchanged as the closing example. This is how the roadmap is in fact pitched: `Dˣ` *is* the inner form of `GL₂`, and nearly all of the general theory planned here is uniform in that generality. Three considerations decide the scope:

- The general framework is uniform. `A ≅ M_m(D)` splits at all but finitely many places, so almost every local factor is `GL_n(F_v)` with maximal compact `GL_n(𝒪_v)`; the `F`-parabolic subgroups correspond to flags of right `D`-spaces (so proper parabolics exist exactly when `m > 1`, and the `m = 1` case is anisotropic modulo centre — the phenomenon this roadmap already treats correctly); the archimedean factors are `GL_a(ℝ)`, `GL_a(ℂ)`, and `GL_{n/2}(ℍ)`.
- Nothing in this scope precludes a later treatment of the analytic theory. In a successor roadmap devoted to standard L-functions, one may use the uniform construction of Godement–Jacquet (*Zeta Functions of Simple Algebras*, LNM 260, 1972) for the multiplicative group of an arbitrary central simple algebra. That programme lies outside the present roadmap; see the exclusions note at the end.
- The comparison between an inner form and `GL_n` is the Jacquet–Langlands correspondence, which stays out of scope; but a later Jacquet–Langlands roadmap then finds both sides already living in one framework.

**Scope of the constructions.** The constructions considered here are foundational. The reorganization proposed in the thread — one roadmap, three parallel tracks (shared automorphic foundations; the split `GL_n` instance; a quaternionic track owning only the genuinely inner-form material), with Whittaker theory, multiplicity one, Rankin–Selberg, and Godement–Jacquet all excluded — is the right resolution. None of the excluded theory is needed even to state the summit. The constructions considered below fall naturally into the proposed three tracks: the adelic foundations, local smooth representation theory, and abstract Hecke algebras are shared; parabolics, Satake theory, and the archimedean theory belong to the `GL_n` instance; and reduced norms, orders, and independence of splittings belong to the quaternionic track. A closing note records the relevant facts about the excluded material, including points that a successor roadmap will need to address.

An honest costing, so "inner forms" is not mistaken for a free generalization: the central-simple-algebra material at general degree — reduced characteristic polynomial, trace, and norm via a splitting with independence of the splitting; orders; spreading out an order so that it is Azumaya away from finitely many primes (this, not global conjugacy of maximal orders, is the genuine global input behind the finiteness of the ramified set); `GL_{n/2}(ℍ)` as a real reductive group — is a substantial body of work. It contains no single hard hidden theorem, and it is the same shape of work this roadmap's Layer 0 already plans for quaternions, but at degree `n`. In the pinned Mathlib, `Algebra.IsCentral`, `IsSimpleRing`, `CSA`, `IsAzumaya`, and `BrauerGroup` exist; general reduced trace/norm and maximal-order theory do not, and `Algebra.discr` requires commutativity, so the reduced-trace discriminant is new work.

## The constructions, their sources, and their pitfalls

The point of this section is the roadmap-repository rule that nothing hard may be smuggled in: each construction below names its route, its source, and the theorems that must be either scheduled or taken as black boxes.

**Adelic foundations.** The required adelic foundations are modest once the L-function programme is excluded. They comprise Haar measure on the local groups, and the invariant quotient measures on the compact homogeneous spaces `U(F)\U(𝔸)` occurring in the constant-term integrals — for a nonabelian unipotent radical `U(F)` is not normal in `U(𝔸)`, so these are homogeneous spaces carrying invariant measures, not quotient groups carrying Haar measure — whose compactness rests on the cocompactness of `F` in `𝔸_F`, together with a height on `G(𝔸)` for the definition of moderate growth. One caveat remains: the entrywise height `‖g‖ = ∏_v max(|g_{ij}|_v, |det g|_v^{-1})` is only quasi-submultiplicative at the archimedean places, with a constant depending on `n`; this constant must either be retained or avoided by using operator norms. Mathlib's `NumberField.AdeleRing`, restricted products, and Haar measure provide what is needed here. The theory of additive characters, Schwartz–Bruhat spaces, and Poisson summation required for a roadmap on L-functions is deliberately omitted.

**What the FLT repository already provides.** The roadmap already plans to migrate the totally definite quaternionic material from the FLT project, and that repository (Apache-2.0, like the planned migration) contains more of the present scope than the totally definite corner. It has the discreteness and cocompactness of `F` in `𝔸_F` (`NumberField.AdeleRing.discrete`, `NumberField.AdeleRing.cocompact`) — exactly the input behind the compactness of the quotients `N(F)\N(𝔸)` above, and present in neither form in the pinned Mathlib — together with the base change `𝔸_F ⊗_F L ≅ 𝔸_L`. It has Fujisaki's lemma in the generality needed here (`NumberField.FiniteAdeleRing.DivisionAlgebra.finiteDoubleCoset`, for an arbitrary finite-dimensional division algebra over a number field): the finiteness of `Dˣ\(D ⊗ 𝔸_f)ˣ/U`, which is the class-set finiteness invoked in the inner-form section below. It has the space of `R`-valued automorphic forms on a totally definite quaternion algebra, with finite-dimensionality at finite level, an inner product, and Hecke operators — the latter built from an abstract double-coset operator on the `U`-invariants of a group action, which is a third formalism alongside Mathlib's `NumberTheory/HeckeRing` and the convolution algebras of the Hecke paragraph below; the comparison of the three should be a stated milestone rather than a discovery. It also has Haar-character machinery on adele rings and a body of restricted-product and module-topology lemmas already arranged for upstreaming into Mathlib. A prototype of automorphic forms on `GL_n` over `ℚ` exists there as well (`GlobalLanglandsConjectures/GLnDefs.lean`); it is incomplete, and it is prior art to consult rather than material to migrate.

**The group.** At split places: the Cartan decomposition of `GL_n(F_v)` is the elementary divisor theorem over the valuation ring; the archimedean Iwasawa decompositions are Gram–Schmidt (`GL_n(ℂ) = U(n)AN`); after fixing the standard lattice or inner product the maximal compacts are the standard ones. For general `A`: the local structure at ramified places, integral orders and their completions, and the ramified-set finiteness by spreading out, as above. The compact-open subgroups feed Mathlib's `RestrictedProduct` machinery, which supplies the abstract adelic topology once they are constructed.

**Strong approximation and the classical–adelic dictionary.** Strong approximation for the reduced-norm-one group holds exactly when the archimedean product is noncompact (Kneser's criterion; Platonov–Rapinchuk §7, Theorem 7.12) — so it holds for `SL_n` split, and *fails* for totally definite quaternion algebras; there the finite adelic double-coset model comes from compactness modulo centre at infinity, and the roadmap's totally definite example is exactly right to be a finite double-coset count rather than a classical dictionary. Two steps must be scheduled explicitly rather than absorbed: additive strong approximation for `F ⊂ 𝔸^S`, and the determinant/idele-class-group step — strong approximation for `SL₂` alone does not produce the classical↔adelic comparison at `n = 2`; the determinant fibration over the idele class group is the other half.

**Local smooth representation theory.** The definitions (smooth, admissible, contragredient, induction) are purely algebraic at the nonarchimedean places, exactly as this roadmap's Layer 2 has them. Principal series with the explicit modulus character `δ_B^{1/2}`; admissibility of parabolic induction via the Iwasawa decomposition. Scheduled with sources, not assumed: the Jacquet functor's exactness and preservation of admissibility (Jacquet; Casselman's notes — Bernstein–Zelevinsky I, Ann. Sci. ENS 1977, states but cites Jacquet for it); the theory of distributions on ℓ-spaces is the 1976 Bernstein–Zelevinsky *survey* (not BZ I), and the invariant-distribution theorem underlying Gelfand–Kazhdan arguments is its own milestone. Taken as a black box: the Bernstein–Zelevinsky/Zelevinsky classification of irreducibles. At ramified places only vocabulary is needed for the summit: `GL_m(D_v)` lives in the same locally profinite framework, and for `m = 1` the group is compact modulo centre, so its irreducible smooth representations are finite-dimensional — a valuable second test of the general definitions.

**The Satake isomorphism, at split places only** — which is all the good-place statement uses. Define the spherical Hecke algebra by the Cartan decomposition and prove the isomorphism onto the symmetric Laurent polynomials `ℂ[X₁^±, …, X_n^±]^{S_n}` following Gross (*On the Satake isomorphism*, in Galois Representations in Arithmetic Algebraic Geometry, CUP 1998): the transform is triangular for the dominance order with computable leading terms. Conventions to pin: the normalized transform `δ_B^{1/2}`-twisted lands in the ordinary `S_n`-invariants and requires a chosen `q_v^{1/2}`; the unnormalized transform carries a twisted Weyl action; the characteristic functions of the double cosets of `diag(ϖ,…,ϖ,1,…,1)` correspond to `q_v`-power-scaled elementary symmetric polynomials, not the bare ones.

**Hecke algebras, shared with the modular-forms roadmap.** The Hecke algebras needed here should extend the abstract double-coset Hecke ring already used in the modular-forms roadmap (Shimura, *Introduction to the Arithmetic Theory of Automorphic Functions*, Ch. 3; a first slice is in Mathlib's `NumberTheory/HeckeRing`), rather than taking convolution algebras as primitive. Shimura's Chapter 3 treats general `n`: for the pair `(GL_n(ℤ_p), GL_n(ℚ_p) ∩ M_n(ℤ_p))` the integral Hecke ring is the polynomial ring `ℤ[T_{p,1}, …, T_{p,n}]` — the Hecke–Tamagawa structure theorem, an integral precursor of the Satake isomorphism. (Tamagawa's proof of the general-`n` case is in his paper on the ζ-functions of division algebras, Ann. of Math. 77 (1963), so the integral theory is already adapted to orders in central simple algebras and fits naturally with the inner-form setting above.)

Only bi-invariant algebras are needed, with varying level `K` — with **one Haar measure `μ` fixed once** and the level idempotents `e_K = 1_K/μ(K)` identifying `H(G ⫽ K) = e_K C_c^∞(G) e_K`, so that `C_c^∞(G) = ⋃_K H(G ⫽ K)` is an honest union of subalgebras; per-level `vol(K) = 1` normalizations do not glue, since shrinking the level rescales convolution by the index. The Iwahori–Hecke algebras required in later roadmaps are the `K = Iwahori` instances of the same construction. The following milestones should be stated explicitly: (i) the convolution comparison `H_ℂ(G ⫽ K) ≅ ℂ ⊗_ℤ R(K, G)` at `vol(K) = 1` for the single level in question, with due attention to the `g ↦ g⁻¹`/opposite-ring convention — Mathlib's present treatment is purely algebraic, so the necessary topological finiteness results and this comparison remain to be formalized; (ii) the passage from Shimura's positive integral ring to the full local algebra by inverting the central double coset, `R(K, GL_n(F_v)) ≅ R⁺[T_{p,n}^{-1}] ≅ ℤ[T_{p,1}, …, T_{p,n−1}, T_{p,n}^{±1}]` (Rhodes–Shemanske); (iii) the normalized Satake isomorphism, including a fixed choice of `q_v^{1/2}`; (iv) the theory of modules at varying level, since smooth representation theory requires the idempotented algebra, its nondegenerate modules, and the transition maps between levels, not merely the individual bi-`K` rings; and (v) the comparison between classical and adelic Hecke operators, which should be proved rather than assumed, since the formula depends substantially on conventions for determinants, central characters, and unitary versus algebraic normalization (in one standard convention the characteristic function of `K_p·diag(p,1)·K_p` acts as `p^{1−k/2}·T_p`; Trotabas §5.3). At ramified places of the inner form, the same construction applies with `K_v = GL_m(𝒪_{D_v})`, although representations with no `K_v`-fixed vectors require smaller levels.

**Archimedean theory.** The `(𝔤,K)`-module definitions are algebraic (Borel–Wallach; Knapp, *Representation Theory of Semisimple Groups*), for the factors `GL_a(ℝ)`, `GL_a(ℂ)` treated honestly as a real group, and `GL_{n/2}(ℍ)`; the noncompact centre must be handled explicitly throughout. Taken as black boxes, with the warning that they are invoked the moment one touches smooth Fréchet models, continuous contragredients, or archimedean Whittaker functionals: Casselman–Wallach globalization, Harish-Chandra's admissibility theorem for irreducible unitaries, and the unitary dual (Vogan 1986 for the archimedean `GL_n`; Tadić's published classification is the nonarchimedean one).

**Automorphic forms.** The Borel–Jacquet definition (Corvallis, PSPM 33, 1979, 189–202): smooth, moderate growth in the height, `K`-finite, `Z(𝔤)`-finite, left `G(F)`-invariant; cusp forms by vanishing of the constant terms `∫_{N(F)\N(𝔸)} f(ng)\,dn` along the proper parabolics. The integral is over a compact quotient — compactness resting on the cocompactness of `F` in `𝔸_F` — so the *definition* needs no reduction theory. The roadmap's decisions here are exactly right and should be kept verbatim in the `GL_n` setting: the cuspidal spectrum defined algebraically (`K_∞`- and `Z(𝔤)`-finite cusp forms), and automorphic representations as bundled objects carrying their local data with an occurrence witness. One warning to record: those definitions avoid the theorems just taken as black boxes *only as long as* no claim is made that the bundled objects exhaust the irreducible constituents or that admissibility is inherited from the proposed local factors — either claim would silently reintroduce Flath's theorem and Harish-Chandra's finiteness theorem. Taken as black boxes, with sources, for whichever later roadmap needs them: reduction theory and Siegel sets (Borel, *Introduction aux groupes arithmétiques*); finiteness of the volume of `G(F)Z(𝔸)\G(𝔸)`; rapid decay of cusp forms and the inclusion of cusp forms in `L²` (Mœglin–Waldspurger I.2; Harish-Chandra); Harish-Chandra's finiteness theorem; and the `L²` spectral decomposition with Eisenstein series (Mœglin–Waldspurger, *Spectral Decomposition and Eisenstein Series*).

**The exclusions.** Whittaker theory, multiplicity one, Rankin–Selberg, and Godement–Jacquet are rightly excluded. None is needed to define automorphic forms, cuspidality, or the Hecke-eigensystem data appearing in the summit statement. For the benefit of a successor roadmap, the following points should nevertheless be recorded. Local Whittaker uniqueness rests on the Gelfand–Kazhdan invariant-distribution method, with the transpose anti-involution `g ↦ ᵗg`, not the contragredient `g ↦ ᵗg⁻¹`. The archimedean case is a genuinely separate analytic theorem, and nondegenerate Whittaker data do not exist for nonsplit inner forms. Their exclusion therefore also removes the most serious asymmetry between the split and inner-form cases from this roadmap.

The Godement–Jacquet programme (LNM 260) would require the full harmonic analysis of `A(𝔸)`: Schwartz–Bruhat spaces, self-dual Haar measures, adelic Poisson summation, and rapid decay of cusp forms. Precisely this material is no longer included among the adelic foundations above and should instead form the opening part of a successor roadmap on L-functions. (Getz–Hahn, GTM 300, remains the reference for the foundations retained here; its L-function chapters develop Rankin–Selberg, likewise excluded.)

## What genuinely changes from `GL₂` to `GL_n`, and from `Dˣ` to `GL_m(D)`

Since the reorganization makes `GL_n` a primary case, the additional requirements for `n ≥ 3` should be stated explicitly. Exactly two changes are unavoidable even to formulate the principal notions. The remaining higher-rank results may be taken as black boxes and need not enter the definitions.

The first change appears at `n = 3`: the definition of cuspidality requires the combinatorics of parabolic subgroups. For `GL₂`, cuspidality is the vanishing of a single constant term, since the Borel is the only proper standard parabolic. For `GL_n`, the standard parabolics are indexed by the compositions of `n` — `2^{n−1} − 1` proper ones — and the definition requires parabolic subgroups, Levi decompositions, unipotent radicals, the compact quotients `U_P(F)\U_P(𝔸)` with their probability Haar measures, and the constant term along each. An initial lemma should establish the maximal-parabolic criterion: cuspidality is equivalent to vanishing along the `n − 1` maximal standard parabolics, of types `(r, n−r)`, because the constant term along any parabolic factors through that of any larger one. From `n = 3`, the maximal unipotent subgroup is nonabelian. Since Fourier coefficients other than constant terms are excluded, this fact has no further effect on the present roadmap. Nevertheless, the formalism for compositions, dominant coweights, and the dominance order will be used in every successor treating Eisenstein series and should be formulated carefully here.

The second change appears at `m = 2`: proper parabolics and a noncompact quotient return for inner forms, although the break does not occur where one might first expect. The class sets `G(F)\G(𝔸_f)/K_f` remain finite for every `m` (Borel's finiteness theorem), but for `m > 1` this finite set no longer captures the entire theory. The quotient acquires a positive-dimensional archimedean direction, proper parabolics exist — they correspond to flags of right `D`-subspaces and are indexed by compositions of `m` — cuspidality ceases to be vacuous, and the continuous spectrum reappears. The reduction to functions on a finite set in the totally definite case is special to `m = 1`. Strong approximation applies to the simply connected group `SL_m(D) = ker(Nrd)`, which is isotropic with noncompact archimedean factors once `m ≥ 2`, and never to `GL_m(D)` itself. The reduced-norm direction is governed by the idele class group, as in the dictionary paragraph above.

**All other higher-rank issues may be assumed, for every `n`:**

- *Moderate growth requires no reduction theory.* Its definition in terms of an adelic height obtained from a faithful representation and its inverse is uniform in `n`. Neither the determinant nor the reduced norm is an adequate height, since both are insensitive to directions such as `diag(t, t⁻¹, 1, …, 1)`. Siegel sets, finite covolume, rapid decay, and the cone combinatorics in rank `> 1` are results rather than ingredients in the definition and lie outside the scope of this roadmap.

- *The archimedean definitions are uniform in `n`.* `(𝔤,K)`-modules, admissibility, and `Z(𝔤)`-finiteness do not change with `n`. The infinitesimal character of an irreducible representation is an algebra homomorphism `Z(U(𝔤)) → ℂ` and can be defined without the Harish-Chandra isomorphism. That isomorphism is needed only to *label* infinitesimal characters by Weyl orbits, while the Langlands classification and Casselman–Wallach theory are needed only for classification or globalization statements. Only the computation of `K`-types becomes more involved as `n` grows, and this concerns the subsequent theorems rather than the definitions.

- *Satake theory requires new proofs, not new definitions.* For `n = 2`, the theory can be verified by direct calculation. From `n = 3`, the Weyl group is nonabelian and the dominance order is genuinely partial, so the main difficulty lies in the combinatorial foundations already listed in the Hecke paragraph: the Cartan decomposition by dominant coweights, finiteness of convolution structure constants, and triangularity with respect to dominance. None of these changes the definitions.

- *A necessary qualification first appears at `n = 4`.* For `n = 2, 3` the residual spectrum consists of the characters `χ ∘ det`, so "discrete = cuspidal ⊕ characters" is true there — and **false** from `n = 4`, where `Speh(σ, 2)` with `σ` cuspidal on `GL₂` is residual and infinite-dimensional (Mœglin–Waldspurger, *Le spectre résiduel de `GL(n)`*, Ann. Sci. ÉNS 22 (1989): the discrete spectrum is parametrized by pairs `(σ, d)` with `σ` cuspidal on `GL_r`, `rd = n`). Any statement about the discrete spectrum must reflect this distinction from the outset. Even to state the classification, normalized parabolic induction and the Speh constituent must be defined, although their existence and the exhaustion of the discrete spectrum may be taken as black boxes.

- *Several apparently higher-rank features are already present in rank two.* The Iwahori–Hecke algebra of `GL₂` is already noncommutative, so noncommutativity at ramified level is not peculiar to `n ≥ 3`. The mirabolic subgroup, Bernstein–Zelevinsky derivatives, and cuspidal support are already nontrivial at `n = 2`; they serve as tools in the excluded programmes and need not be included here. Finally, there is no Hilbert-90-style obstruction for split `GL_n`, because `det` admits the section `a ↦ diag(a, 1, …, 1)`. Questions about the image of the reduced norm for inner forms (Hasse–Schilling–Maass) are already present at `m = 1`.

## Local points on the present text

Each has a line comment. First, at line 72, "for even-rank groups, the passage between cohomological/C-algebraic and L-algebraic normalizations involves a half-Tate shift" is imprecise: for `GL_n` the two normalizations differ by the twist `π ⊗ |det|^{(n−1)/2}`, half-integral exactly when `n` is even; here `n = 2` and its inner forms (Buzzard–Gee, already in the references, state it this way). Second, at line 15, "good" occurs in the summit statement before it is defined; the summit should carry a named definition — `v` outside a finite set `S`, `A_v` (here `D_v`) split, `π_v` unramified, `v ∤ p` — since the factor `N(v)` in `X² − a_vX + N(v)s_v` depends on conventions exactly there. Third, at line 508, the normalized Satake transform requires the chosen `q_v^{1/2}`, as above.

## Confirmations

Independently checked, since these are the roadmap's boldest choices: the algebraic definition of the cuspidal spectrum, without an `L²` decomposition, is correct provided "spectrum" never asserts a decomposition — the text already respects this; cuspidality is vacuous for division `D` and the separate exclusion of reduced-norm characters is the right treatment; and the relative pair — full real Lie algebra, genuine maximal compact, split centre not adjoined to `K` — agrees with Borel–Wallach, with cohomology in degrees `1, 2` for the cohomological `GL₂(ℝ)` discrete series and `0, 1` for `ℍˣ` when the infinitesimal central characters cancel.

## References (all verified)

Bump, *Automorphic Forms and Representations*, CSAM 55 (1997) · Getz–Hahn, *An Introduction to Automorphic Representations*, GTM 300 (2024) · Godement–Jacquet, *Zeta Functions of Simple Algebras*, LNM 260 (1972) · Flath, PSPM 33 (1979), 179–183 · Borel–Jacquet, PSPM 33 (1979), 189–202 · Jacquet–Langlands, LNM 114 (1970) · Gross, *On the Satake isomorphism* (CUP, 1998) · Bernstein–Zelevinsky, Ann. Sci. ENS (1977) and the 1976 survey · Casselman's notes · Borel, *Introduction aux groupes arithmétiques* · Mœglin–Waldspurger, *Spectral Decomposition and Eisenstein Series* · Borel–Wallach · Knapp · Platonov–Rapinchuk · Vogan (1986) · Shimura, *Introduction to the Arithmetic Theory of Automorphic Functions*, Ch. 3 · Rhodes–Shemanske · Trotabas.

kim-em:
:robot: Codex suggests:

I think a single roadmap is probably better—provided we include only the `GL_n` foundations relevant to automorphic representations, not Chris’s downstream Whittaker and Godement–Jacquet programme.

It would avoid both undesirable outcomes:

- making this roadmap depend on a sibling that does not yet exist;
- initially designing supposedly reusable APIs around `Dˣ`, then extracting them later.

I would reorganize the existing roadmap into three parallel tracks:

```text
shared automorphic foundations
          │
          ├── GL_n instance ────────┐
          │                         │
          └── quaternionic Dˣ ──────┼──▶ attached-system statement
                                    │
local Galois/compatible systems ────┘
```

The shared track would own smooth/admissible representations, Harish–Chandra modules, automorphic forms, relative cohomology, restricted tensor products, and abstract spherical Hecke algebras.

The `GL_n` track would provide the primary concrete instances:

- `GL_n(F_v)` and `GL_n(𝔸_F)`;
- standard maximal compacts and block parabolics;
- principal series and normalized induction;
- Cartan/Iwasawa decompositions;
- automorphic forms on `GL_n`;
- Satake theory, with exact `q_v` normalizations;
- `GL₁` characters and the adelized discriminant for `GL₂`.

The quaternionic track would then own only genuinely inner-form material:

- quaternion algebras, orders, ramification and restricted products;
- `D_vˣ`, `ℍˣ`, reduced norm and maximal compacts;
- independence of local splittings;
- comparison with the shared/`GL₂` interfaces at split places;
- the totally definite FLT migration.

I would explicitly exclude Whittaker theory, multiplicity one, Rankin–Selberg, and Godement–Jacquet. Those are valuable later roadmaps but are not prerequisites for stating the target.

The roadmap’s title could become:

> **Automorphic representations of `GL_n` and quaternionic inner forms**

with a prominently identified summit:

> **Galois representations attached to quaternionic automorphic representations**

The main cost is that the PR becomes larger. But structurally this is cleaner, and it prevents us from knowingly choosing an API ownership boundary that we already expect to reverse. I would favor this over the fast-follow approach.

TauCetiRoadmap/AutomorphicRepresentations/README.md
kim-em: Addressed in 1d4c6f9. The target now quantifies a finite bad set S containing all nonsplit and ramified places, defines IsGoodPlace π D S v through v ∉ S with the required split/spherical consequences, and keeps v ∤ p separate at the family-member use site. Layer 6 packages the same contract in RationalityModel.

kim-em:
:robot: Codex implemented the agreed single-roadmap structure in 1d4c6f9.

The revised roadmap now has shared automorphic foundations, split `GL_n` as the primary concrete family for every positive `n`, the quaternionic `Dˣ` material actually needed by the endpoint, and the independent local-Galois/compatible-system track. It explicitly excludes arbitrary `GL_m(D)` and the downstream Whittaker, multiplicity-one, Rankin--Selberg, Godement--Jacquet, Jacquet--Langlands, and spectral/Eisenstein programmes.

It also treats Modular Forms PR #47 Layer 2 as a planned input: that roadmap owns the integral general-`GL_n` double-coset calculation over `ℤ_p/ℚ_p`, while this roadmap owns the extension to arbitrary number-field places, the convolution comparison, normalized Satake, and adelic actions.

I went through the layers again for skipped prerequisites. The text now schedules the higher-rank parabolic/maximal-parabolic argument, nonabelian unipotent quotient measures and Fubini comparison, both the additive/`SL₂` and determinant parts of the classical--adelic dictionary, general-place lattice counting, height control through `g` and `g⁻¹`, the full real-group treatment of `GL_n(ℂ)`, and the exact Satake square-root and `q_v`-power conventions.

All three inline comments have been answered and resolved. The focused and full builds pass.

CBirkbeck (commented):
*(Review posted by Fable, Chris Birkbeck's AI assistant, at Chris's request, jointly with GPT 5.6 Sol Pro: the numbered findings are Sol Pro's, checked independently before posting — the verifications are recorded below.)*

**Verdict: request changes.** The reorganized roadmap is unusually careful — the track separation, the intrinsic quaternion-algebra definition, the refusal to hide missing mathematics behind empty `Prop`s, and the `ℚ̄_p`-valued compatible-system design are all right. But the current head contains genuine errors, one unresolved normalization that changes the meaning of the summit theorem, and several substantial theorems still hidden inside examples or "independence" claims.

**Independently verified before endorsement.** (a) The locally-constant-without-finite-level counterexample on `ℚ_p` is correct: for `B_n = p^{−n} + p^nℤ_p`, the union `⋃_{n≥1} B_n` is clopen (the balls sit in distinct valuation shells and escape to infinity), yet for any candidate period `p^mℤ_p` and `n > |m|` the points `p^{−n}` and `p^{−n} + p^m` take different values. (b) The Heisenberg computation `u(0,t,0)^{-1}·u(1,0,0)·u(0,t,0) = u(1,0,t)` is exact, so `U(F)` is not normal in `U(𝔸_F)` and `U(F)\U(𝔸_F)` is a homogeneous space, not a group. (c) The rescaling `f *_{μ′} g = m·(f *_μ g)` for `μ′ = mμ` is exact, so per-level `vol(K) = 1` normalizations are incompatible with a literal union of level algebras. (d) The normalized Satake formula `𝒮(c_i) = q^{i(n−i)/2}·e_i(z)` matches Gross, and the `n = 2` chain `T = q^{1/2}(z₁+z₂)`, `S = z₁z₂`, `(X−α₁)(X−α₂) = X² − TX + qS` with `α_i = q^{1/2}z_i` is right. (e) The dual-polynomial identity `P_{ρ^∨}(X) = X^d·P_ρ(X^{-1})/P_ρ(0)` and the restriction rule `Frob_w ↦ Frob_v^{f(w/v)}` are right; restriction is **not** `P_v` unchanged. The quoted passages are all present at the current head.

## Merge-blocking corrections

1. **Local constancy ≠ finite level.** Two predicates: `LocallyConstantFinitePart f` and `FiniteLevel f := ∃ K_f compact open, f(gk) = f(g)`. Finite level ⟹ locally constant; the converse only under a compactness hypothesis. Automorphic forms and smooth induction use the finite-level predicate. The current equivalence claim is false (counterexample above).
2. **`U(F)\U(𝔸_F)` carries an invariant quotient measure, not Haar.** Build the homogeneous space, the right action, the invariant measure via a Weil integration formula, finiteness/normalization of total mass, and Fubini/filtration compatibility. Prose may say "quotient Haar"; the Lean object must not be `HaarMeasure` on a quotient group.
3. **One fixed Haar measure across levels.** Fix `μ` once; define every `H(G ⫽ K; μ)` with it; set `e_K = 1_K/μ(K)`; identify `H(G ⫽ K) = e_K C_c^∞(G) e_K`; literal inclusions as `K` shrinks. (Or keep per-level normalizations with explicit rescaled transition maps — but then the level algebras are not literal subalgebras of one convolution algebra.)
4. **State the exact Satake formula in the README**, not "the precisely stated `q_v`-power": `𝒮(c_i) = q^{i(n−i)/2}·e_i(z₁,…,z_n)`, with four separately named objects — normalized Satake parameters `z_i`, arithmetic roots `q^{(n−1)/2}z_i`, normalized Hecke eigenvalues, and the classical unnormalized `T_v, S_v` — and an acceptance test comparing all four.
5. **Resolve SIGN-OFF B — until it is resolved, the final theorem has no fixed meaning.** It controls `W` vs `W^∨`, the split real center, the C-to-L-algebraic twist, the finite central character, normalized vs arithmetic parameters, the convolution inverse, arithmetic vs geometric reciprocity, `ρ` vs its dual, and the displayed polynomial. Add one normalization table `(π_∞, W, ω_π) → π^{L-alg} → (z_{v,i}) → (a_v, s_v) → ρ(Frob_v^arith)` with every twist, dual, inverse, and `q_v`-power on the arrows — and make the weight-12 regression test `X² − τ(p)X + p^{11}` in the declared convention.
6. **Prove cohomological ⟹ paritious ⟹ L-algebraic-after-pinned-twist**, or add the hypothesis to the summit. Non-paritious Hilbert forms occur on the definite quaternionic side with only projective/intermediate Galois representations; the implication is currently an intention, not a theorem. The proof should make the central argument explicit: relative cohomology matches infinitesimal central characters; the global central character on `F^×` plus the unit theorem forces equal central exponents at the real embeddings.
7. **Central characters and moderate growth.** `f(zg) = ω(z)f(g)` does not descend to the quotient modulo center unless `ω` is trivial. Define `A_G` and pick one contract: trivialize on `A_G` and define growth on `G(F)A_G\G(𝔸_F)`; or require `ω|_{A_G}` unitary/trivial; or work with sections of the line bundle attached to `ω`, with the correspondingly corrected norm.
8. **Uniform moderate growth.** Either define automorphic forms with uniform moderate growth (`∃N ∀D ∈ U(𝔤), |Df| ≪_D |g|^N`), or add the Harish-Chandra–Borel theorem deriving it; only then does the `U(𝔤)`-action preserve the space.
9. **Nonarchimedean Iwasawa first.** `G = KP` for every standard parabolic, the modulus character `δ_P(m) = |det(Ad m|_{𝔲_P})|_v` with its chosen square root, compactness of `P\G`, finiteness of `P\G/J`, and the deduction that induction preserves admissibility — before the principal-series tests, not at Layer 6.
10. **Restricted tensor products depend on the distinguished vectors.** Keep them as data; maps carry distinguished vectors to distinguished vectors almost everywhere; invariance under nonzero scalars only. Independence from the choice is claimable only after the one-dimensional spherical-line theorem lands (Layer 6).
11. **The `Δ` track needs a named factorization milestone**: archimedean discrete-series module; each unramified local factor with normalized spherical vector; the restricted tensor product; the equivariant map onto the automorphic orbit of `Δ`; the subquotient, irreducibility, admissibility, non-character proofs. Bundling local factors postpones Flath for *typing*; it does not construct this bundle from a classical form. Show the resulting Layer-6 edge in the dependency graph, or take a special factorization theorem as an explicit black box.
12. **Complete the archimedean side of the classical–adelic bridge**: the determinant-negative component and reflection giving the full `O(2)`-type (weight `k` alone gives the `SO(2)`-character), holomorphic/antiholomorphic components, the `A_G`-action, the right-translation convention, and cuspidality as vanishing of *every* adelic constant term via the quotient dictionary.

## Missing prerequisites that must become named milestones

13. **Splittings transport along a torsor, they are not "chosen-independent".** Skolem–Noether gives inner transport, with a cocycle law; integral statements need an *adapted* splitting carrying the chosen maximal order to `M₂(𝒪_v)` (existence from local conjugacy of maximal orders; ambiguity controlled by the normalizer). Literal independence only for conjugacy-invariant outputs; intertwiners are canonical only up to central scalar.
14. **Compactness mod center and class-set finiteness are named Layer-1 theorems** (with stabilizer finiteness, the fixed-level function-space equivalence, and level/Hecke compatibility), not facts left to the FLT migration.
15. **Supply the quaternion algebra of acceptance test 5**: either the prescribed-ramification classification (even sets of places, Hilbert reciprocity) or a pinned concrete example with proved local splittings.
16. **Fix one finiteness-of-ramification route**, e.g. order → Azumaya away from finitely many primes → Brauer vanishing over finite residue fields + henselian comparison → split at almost all `v`; name every step.
17. **Exactness of `K`-invariants is a theorem** (Reynolds averaging in characteristic zero, disconnected `K` via Haar), needed before any relative-cohomology long exact sequence; the finite-place analogue (exactness of compact-open fixed vectors) belongs in Layer 2.
18. **`IsCohomological` must carry its data**: a Type-valued `CohomologicalWitness` (coefficient field and system with all archimedean embeddings and conjugation data at complex places, degree, nonzero class, splitting-transport as coherent equivalences), consumed by the summit — not an opaque existential.
19. **One shared good-place datum**: a single `GoodPlaceHeckeDatum` (bad set + polynomials + sphericity + comparison) used by both the rationality model and `GoodPlaceCompatible`; state semisimplicity once, not twice.
20. **Compatible-system operations need their polynomial formulas**: `P_{ρ⊕σ} = P_ρP_σ`; `P_{ρ^∨}(X) = X^d P_ρ(X^{-1})/P_ρ(0)` with `P_ρ(0) ≠ 0`; Tate twists scaling roots by the pinned power of `N(v)`; tensor products via symmetric-function/resultant machinery without choosing roots; restriction via `Frob_w ↦ Frob_v^{f(w/v)}`; bad-set and coefficient-field bookkeeping for each.
21. **The finite-order Hecke-character test imports global class field theory** — remove it (keep the cyclotomic family as the rank-one test) or add a delimited global-CFT prerequisite (Artin map, continuity, local–global compatibility, Frobenius formula).
22. **Disambiguate "base change along a number-field embedding"** into `coeffScalarExtension`, `restrictToFieldExtension`, `transportSourceFieldEquiv`, `conjugateByClosureEmbedding`, with variance and residue-degree formulas.
23. **Tighten `Suggested.lean`**: `n : ℕ+`/`[NeZero n]`; the topological ring structure `FramedGaloisRepresentation` actually needs; `PadicGaloisFamily` as a structure if it is to carry operations; and resolve `QuaternionAutomorphicRepresentation` vs the separate `hπauto` hypothesis (bundle the occurrence witness or rename the type).

## Tests to add

Finite-level ⟹ locally constant plus the explicit `ℚ_p` counterexample; the `GL₃` unipotent homogeneous-space example; a level comparison `K′ ⊂ K` with one fixed `μ` and the idempotents; a Satake test separating `z_i` from `q^{(n−1)/2}z_i`; the full `O(2)` in the adelization of `Δ`; the exact `Δ`-polynomial `X² − τ(p)X + p^{11}`; a concrete totally definite quaternion algebra unramified at every finite place; and a test that the rationality model and the compatible system consume the identical good-place datum.

## Required before approval

(1) the local-constancy correction; (2) invariant homogeneous-space measures; (3) the single-measure or explicitly-rescaled Hecke design; (4) SIGN-OFF B resolved with the exact Satake formula and the `Δ` normalization test; (5) the cohomological-to-paritious/L-algebraic theorem or added hypothesis; (6) nonarchimedean Iwasawa and the `Δ`-factorization milestone in the dependency graph; (7) the `A_G`/uniform-growth contract; (8) the shared good-place datum; (9) the restricted-tensor-product choice dependence; (10) compactness, class-set, prescribed-ramification, and invariants-exactness as named milestones.

*A note on the earlier Fable review of this PR: findings 2 and 3 also supersede two of its phrasings — "Haar measure on the compact quotients `N(F)\N(𝔸)`" and the per-level `vol(K) = 1` union of Hecke algebras — and that review has been amended in place accordingly.*

kim-em:
:robot: Codex update: I pushed commit 5d04b68 (plus a merge of current `main`) implementing the agreed revision and the actionable points in the expanded Chris/Fable/Sol review.

The roadmap now:

- separates local constancy from finite level, and makes finite level the automorphic/smooth-induction condition;
- treats `U(F)\U(𝔸)` as a homogeneous space with an invariant measure from Weil integration, and adds the missing adelic discreteness/cocompactness, Iwasawa, class-set, and stabilizer prerequisites;
- fixes one Haar measure across levels and defines level Hecke algebras as the corners `e_J * C_c^∞(G) * e_J`;
- pins the Satake/arithmetic/Galois normalization in a table, including `Sat(c_i)=q^{i(n-i)/2}e_i(z)`, and makes `X²-τ(p)X+p¹¹` a required regression test;
- adds Type-valued automorphic and cohomological witnesses, the central-parity/L-algebraic normalization step, explicit real discrete-series and full `O(2)` data, and the complete local-factorization/occurrence track for the discriminant;
- replaces the duplicated rationality/compatibility records with one shared `GoodPlaceHeckeDatum`, and specifies exact root-free formulas for compatible-system operations;
- fills the quaternionic splitting-torsor, ramification, prescribed-ramification, compactness, and finite-class-set gaps, and turns all of the review's requested examples into acceptance tests.

`Suggested.lean` now also excludes rank/dimension zero, distinguishes underlying algebraic groups from their production topologies, requires a topological-ring coefficient in framed Galois representations, and makes `PadicGaloisFamily` a structure. The targeted file and the full `lake build` both pass.

kbuzzard:
Hells bells. AI generates such noise.

I thought about what should be being done here (before looking at this roadmap -- I was offline all yesterday) and I decided that sticking to weight 2 forms for inner forms of PGL_2 should be fine for FLT. It would certainly be enough to state the versions of Theorems A and B in the FLT thread [#FLT > Main blockers for a full proof of FLT. @ 💬](https://leanprover.zulipchat.com/#narrow/channel/416277-FLT/topic/Main.20blockers.20for.20a.20full.20proof.20of.20FLT.2E/near/613821903) . 

This project is far more ambitious, dealing with GL_n, which opens up far more cans of worms (for n>2 there are no discrete series representations, forcing you to have to develop the entire theory in full generality rather than just sticking to the one infinity type which is needed for FLT). 

Part of the roadmap is weirdly focussed on GL_n(D) for D a quaternion algebra. This is just a random inner form of GL_{2n}. If one is working in this generality then one should surely just work with units of central simple algebras, rather than restricting to the random subset "matrix algebras over the ground field, and matrices algebras over quaternion algebras", which I am not at all sure makes things easier and might make things artificially harder (I do not know the theory at all beyond n=2).

If we go down this route then one will have to develop all the theory at infinity, rather than just sticking to weight 2 discrete series reps for GL_2(R) and the trivial one dimensional representation of H^*. This will probably slow down the roadmap substantially as one will have to do all the analysis at infinity, (g,K)-modules, relating C^infinity and L^2 functions on non-compact manifolds. None of this is needed for FLT. One day we want this. But do we want this in this roadmap? Is now the time? Of course as well as being far too general for FLT, this roadmap is also far too specialised, because units of central simple algebras are just one special case of a reductive algebraic group, which is the correct generality for all of this. 

In short, I think I'm saying: this is both far too general for FLT, and not general enough for a general theory of automorphic representations. Should it be what we do?

CBirkbeck:
You can blame me for the GLn part. My thinking was that GL2 was too narrow and that general reductive groups was too much right now, so GLn feels like a nice middle ground (although another way to say it is that it fails in two directions lol). But Kim's original roadmap was for GL2, so I am happy to be outvoted here!

kim-em:
Kevin, let's split it into two roadmaps. Get this one down to exactly what you think you need, and then a "let's doing everything thoroughly" one that can follow on. I propose we just strip stuff out for now, and worry about writing the second one once this is merged and people have pointed some agents at it.

TauCetiRoadmap/ReductiveGroups/README.md
```diff
@@ -68,9 +68,10 @@ prove them equivalent** rather than committing.
 Comodules over a coalgebra (**none in Mathlib**) and the finite-dimensional subcoalgebra
 theory; the convolution group structure on the functor of points; the fppf-sheaf /
 faithfully-flat-descent machinery on `CommAlgCat k`; the explicit Hopf ⇆
-affine-group-scheme equivalence (the Toric work,
-[mathlib4#39281](https://github.com/leanprover-community/mathlib4/pull/39281), is not in
-master); the Lie algebra `Lie(G)` and the adjoint representation; Hopf ideals / closed
+affine-group-scheme equivalence (the convolution source from
+[mathlib4#39281](https://github.com/leanprover-community/mathlib4/pull/39281) is present at the
+pinned Mathlib revision but still uses `suppress_compilation`, so it is not yet a compiled
+foundation); the Lie algebra `Lie(G)` and the adjoint representation; Hopf ideals / closed
```
kbuzzard, line 74: Whatever does "so it is not yet a compiled foundation" mean?

TauCetiRoadmap/WeightTwoAutomorphicForms/Suggested.lean
```diff
@@ -0,0 +1,133 @@
...
+
+/-- The scalar extension of `D` to the full adèle ring. Layer 1 compares this algebra with
+the restricted product of the local scalar extensions. -/
+abbrev QuaternionAdeleAlgebra := NumberField.AdeleRing (𝓞 F) F ⊗[F] D
+
+/-- The finite-dimensional module topology on the adelic scalar extension. Layer 1 proves
+that the induced ring topology is coordinate-independent. -/
+@[reducible] noncomputable def quaternionAdeleTopology :
+    TopologicalSpace (QuaternionAdeleAlgebra F D) :=
+  moduleTopology (NumberField.AdeleRing (𝓞 F) F) (QuaternionAdeleAlgebra F D)
```
kbuzzard, line 71: I would not recommend making this definition, because it's not needed. My opinion is that if you want a topology on `QuaternionAdeleAlgebra F D` then you just assume that it has one, and that it's the module topology, i.e. `[TopologicalSpace (QuaternionAdeleAlgebra F D)] [IsModuleTopology (NumberField.AdeleRing (𝓞 F) F) (QuaternionAdeleAlgebra F D)`. You can use `moduleTopology` to fill in any holes later on if necessary.

TauCetiRoadmap/WeightTwoAutomorphicForms/Suggested.lean
```diff
@@ -0,0 +1,133 @@
...
+
+/-- With the module topology, the adelic scalar extension is a topological ring. -/
+example (hD : IsQuaternionAlgebra F D) :
+    letI : TopologicalSpace (QuaternionAdeleAlgebra F D) := quaternionAdeleTopology F D
+    IsTopologicalRing (QuaternionAdeleAlgebra F D) := by
+  sorry
```
kbuzzard, line 77: This is in FLT in some form. Again I think it's a bad way to do things -- the example should be assuming `[TopologicalSpace]` and `[IsModuleTopology]` and shouldn't have some random letI in it. This way it's more general. Furthermore you can make it an instance.

TauCetiRoadmap/WeightTwoAutomorphicForms/Suggested.lean
```diff
@@ -0,0 +1,133 @@
...
+abbrev QuaternionAdeleMultiplicativeGroup := (QuaternionAdeleAlgebra F D)ˣ
+
+/-- The scalar extension of `D` to the completion at a finite place. -/
+abbrev QuaternionFiniteLocalAlgebra (v : HeightOneSpectrum (𝓞 F)) :=
+  v.adicCompletion F ⊗[F] D
+
+/-- Its finite-dimensional module topology. -/
+@[reducible] noncomputable def quaternionFiniteLocalTopology
+    (v : HeightOneSpectrum (𝓞 F)) : TopologicalSpace (QuaternionFiniteLocalAlgebra F D v) :=
+  moduleTopology (v.adicCompletion F) (QuaternionFiniteLocalAlgebra F D v)
```
kbuzzard, line 90: Again this definition is completely unnecessary. Want an explicit topology for some reason? Use `moduleTopology` directly. But want to prove theorems? Just assume it's been given a topology.

TauCetiRoadmap/WeightTwoAutomorphicForms/Suggested.lean
```diff
@@ -0,0 +1,133 @@
...
+
+/-- Finite local quaternionic units are locally compact and totally disconnected. -/
+example (hD : IsQuaternionAlgebra F D) (v : HeightOneSpectrum (𝓞 F)) :
+    letI : TopologicalSpace (QuaternionFiniteLocalAlgebra F D v) :=
+      quaternionFiniteLocalTopology F D v
```
kbuzzard, line 95: again you just want to assume top space and mod top in this theorem, not this.

TauCetiRoadmap/WeightTwoAutomorphicForms/Suggested.lean
```diff
@@ -0,0 +1,133 @@
...
+classical `T_v` and arithmetic-Frobenius normalization. -/
+noncomputable def weightTwoGoodHeckePolynomial
+    {E : Type u} [Field E] (Nv : ℕ) (av : E) : E[X] :=
+  X ^ 2 - C av * X + C (Nv : E)
+
+/-- The constant eigenpacket regression: `a_v=N(v)+1` gives the roots `1` and `N(v)`. -/
+example {E : Type u} [Field E] (Nv : ℕ) :
+    weightTwoGoodHeckePolynomial (E := E) Nv (Nv + 1 : ℕ) =
+      (X - C 1) * (X - C (Nv : E)) := by
+  sorry
```
kbuzzard, line 114: I am not at all clear what the point of this is.

TauCetiRoadmap/WeightTwoAutomorphicForms/Suggested.lean
```diff
@@ -0,0 +1,133 @@
...
+
+/-! ## Layer 5: two-dimensional Galois families -/
+
+/-- A framed continuous two-dimensional Galois representation. Layer 5 also supplies the
+basis-free formulation and proves the comparison after choosing a basis. -/
+def TwoDimensionalGaloisRepresentation (K : Type u) [Field K]
+    (A : Type v) [CommRing A] [TopologicalSpace A] [IsTopologicalRing A] :=
+  Field.absoluteGaloisGroup K →ₜ* GL (Fin 2) A
```
kbuzzard, line 122: I think it's better to have `Ksep : Type u` and `[IsSepClosure Ksep K]` and then do the Galois group of Ksep over K, I am not sure we want to restrict to the given Kbar chosen via AC. For example you will have then trouble pulling back a global 2-dimensional Galois representation to a local one.

TauCetiRoadmap/WeightTwoAutomorphicForms/README.md
```diff
@@ -0,0 +1,470 @@
+# Weight-two automorphic forms on quaternionic inner forms of `PGL₂`
+
+## Target and boundary
+
+This roadmap builds exactly the language needed to state the following attachment problem.
+
+> Let `F` be a totally real number field, let `D` be a quaternion algebra over `F`, and let `π`
+> be a parallel-weight-two automorphic Hecke eigenpacket for `PGL₁(D)`. There are a number field
```
kbuzzard, line 8: I don't know what a Hecke eigenpacket is. Why does this not just say "parallel-weight-two cuspidal automorphic representation for PGL_1(D)"? I think that's what it's called nowadays.

At some point one is going to have to be more precise about what one means by "parallel-weight-two". This is laying down what the components of the automorphic representation are at the infinite places v of F. There are two choices: either D_v=M_2(\R) in which case we want pi_v to be weight 2 discrete series (have I nailed it down up to isomorphism by saying this and also demanding trivial central character?), or D_v=H the quaternions, in which case we want pi_v to be the trivial 1-dimensional representation. Edit: I see this is said later.

TauCetiRoadmap/WeightTwoAutomorphicForms/README.md
```diff
@@ -0,0 +1,470 @@
...
+> `E`, a finite set `S` of finite places, coefficients `a_v ∈ E` for `v ∉ S`, and a semisimple
+> family of continuous representations
+> `ρ_{p,φ} : Gal(F̄/F) → GL₂(Q̄_p)`, indexed by primes `p` and embeddings
+> `φ : E → Q̄_p`, such that for every `v ∉ S` with `v ∤ p`, the representation is
+> unramified at `v` and arithmetic Frobenius has characteristic polynomial
+> `X² - φ(a_v)X + N(v)`.
```
kbuzzard, line 14: Perfect. That'll keep the AIs happy. 

Historical remarks: Wiles' original proof only needed the case F=Q, and for D=M_2(Q) I could imagine that this was in scope nowadays, but for D a nonsplit quaternion algebra I suspect that one needs Jacquet-Langlands in some form. Demanding this for more F makes other parts of the argument easier because it gives us an "extra degree of freedom" when manipulating automorphic forms (for example we do not need Ribet's theorem, unlike Wiles).

TauCetiRoadmap/WeightTwoAutomorphicForms/README.md
```diff
@@ -0,0 +1,470 @@
...
+
+Here `D` is allowed to be `M₂(F)`, a division algebra split at some real places, or a totally
+definite quaternion algebra. The Lean implementation uses automorphic forms on `Dˣ` invariant
+under the adelic center, together with a proved equivalence to the corresponding `PGL₁(D)`
+formulation. This avoids making a quotient-group presentation the primitive analytic object while
```
kbuzzard, line 19: "together with a proved equivalence to the corresponding `PGL₁(D)` formulation."

I don't know what this means. My impression is that (a) we are a long way from being able to say "let G be a connected reductive group" and also "Oh look, D^*/F^* is an example", both of which one needs to be able to do before one can say "and now let's relate the concrete constructions of modular forms (in mathlib) and automorphic forms (here) to some fancy general theory".

TauCetiRoadmap/WeightTwoAutomorphicForms/README.md
```diff
@@ -0,0 +1,470 @@
...
+representations, Harish--Chandra modules, relative Lie-algebra cohomology, general Satake theory,
+Whittaker models, trace formulae, or Eisenstein series. Those subjects belong in a later roadmap.
+
+## Pinned conventions
+
+1. A quaternion algebra over `F` is intrinsically a central simple `F`-algebra of dimension four,
+   never a chosen Hilbert-symbol presentation.
+2. `PGL₁(D)` means `Dˣ/Fˣ` on rational points and the corresponding central quotient locally and
+   adelically. Automorphic functions are implemented on `Dˣ` with trivial adelic central character;
+   descent to and pullback from the projective quotient are proved inverse.
```
kbuzzard, line 41: "descent to and pullback from the projective quotient are proved inverse." I don't know what this means but I don't think we need it. Just like we don't need to rush to interpret classical modular forms as some functions on a projective adelic object.

What we need here is just a mathematically accurate definition of a weight 2 automorphic form, so we can start making profound statements about them.

TauCetiRoadmap/WeightTwoAutomorphicForms/README.md
kbuzzard: Actually, there is a big design decision which needs to be made here. What is the "capstone" that we want here? The main definition, that the main theorem is about?

I feel like before we have answered this, we're not even ready to write a roadmap, because the roadmap should be focussed on the definition. 

I currently think that the definition we want is *something like* this (but this is not a complete idea).

We are given a totally real field F, and a finite set S of finite places of F where things are allowed to be "bad".

We are given a quaternion algebra D over F unramified outside S. Except that maybe I want to assume D is unramified everywhere.

We are also given a compact open subgroup U of (D tensor_F (finite adeles of F))^* which is unramified outside S in a way which it is possible to make precise, but this needs some thought (there are different possibilities, because D is not canonically M_2(F_v) locally and one has to choose isomorphisms at different v which are compatible in a subtle sense). This needs some thought. In FLT I use `WithRigidification` but this assumes D is unramified at all finite places, which I'm not sure we want to do here.

I think that perhaps we should also be given a finite order character chi:(finite ideles of F) / F^* -> C^* (C the complexes), unramified outside S. Whether we include this or not is a design decision which I will have to think about. I think that we can get away without it in FLT but I am still torn about whether we should include it here.

 Associated to all of this, we need a finite-dimensional complex vector space of weight 2 forms for D^* of level U and character chi, equipped with an action of commuting linear Hecke operators T_v for all finite places v not in S.

I feel like there is a lot of stuff which needs to be thought about here and these are fundamental design questions. Thus I am not sure that I am ready to review this PR yet.

kim-em:
Okay, Kevin says we should wait for him: [#Tau Ceti > Getting started: roadmaps @ 💬](https://leanprover.zulipchat.com/#narrow/channel/610393-Tau-Ceti/topic/Getting.20started.3A.20roadmaps/near/615172136)

PR #151, opened 2026-08-04
Roadmap: Poincaré’s classical nonintegrability theorem for the planar CR3BP

gersh:
## Summary

- make Poincaré’s classical planar parameter-analytic theorem the single summit
- reduce the dependency graph to focused Hamiltonian mechanics, analytic parameter families, torus Fourier analysis, planar Kepler/Delaunay coordinates, and the effective-resonance obstruction
- remove the differential-Galois, higher-variational-equation, singular blow-up, contour, fixed-mass, and spatial milestones
- derive the standard Delaunay normalization intrinsically and require explicit angle periods and symplecticity
- make the effective-resonance nonvanishing/density calculation the principal source-sensitive bottleneck
- simplify Suggested.lean to honest planar CR3BP definitions and acceptance checks

## Response to review

The fixed-mass Yagasaki theorem is no longer a roadmap target. A short future-direction note records why the proposed singular-rescaling transport lemma is invalid for arbitrary meromorphic first integrals and lists possible research routes, but none is a milestone or dependency.

The classical route now explicitly requires:

- a uniform collision-free analytic domain
- standard Delaunay actions and 2π-periodic angles
- the submersion/constant-rank hypotheses in the functional-dependence step
- a rigorous source audit for the disturbing-function coefficient nonvanishing and effective-resonance density argument

## Validation

- python3 .github/scripts/check_roadmap_areas.py
- lake build TauCetiRoadmap.RestrictedThreeBody.Suggested
- full lake build before rebasing onto the latest main
- git diff --check


TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+27. Near `(-μ,0[,0])`, implement
+    `x+μ=ε²ξ`, `y=ε²η`, `z=ε²ζ`, the momentum shifts/scalings, and
+    `t ↦ ε³t`. Prove that this is symplectic with the stated Hamiltonian rescaling and
+    derive the exact transformed vector field before taking a series.
+28. Establish the Taylor expansions through the first nonzero order used by the proof:
+    Coriolis terms at order `ε³`, cancellation at order `ε⁴`, and the tidal perturbation
+    at order `ε⁶`. Give certified remainder statements on a named collision-free complex
+    neighbourhood; a symbolic `O(ε⁷)` with no uniform domain is not an acceptance
+    criterion.
+29. Prove the transport lemma: a meromorphic complete set for the original fixed-`μ`
```
kim-em: > 29. Prove the transport lemma: a meromorphic complete set for the original fixed-`μ` system pulls
>     back to one meromorphic jointly in the blown-up state and `ε` near `ε=0`. This is the logical
>     bridge that allows Theorem 2.1 to prove Theorem 1.1.

:robot: Claude+Codex say:

Identifying this as the bridge is exactly right, and it is the best thing in the roadmap: the paper
never isolates it. But as stated the lemma is false, and the roadmap currently asks a contributor
to prove it.

The pullback is `pₓ = p_ξ/ε`. Take `F = exp(pₓ)`, which is entire, hence meromorphic on `𝒮₂` in the
sense Theorem 1.1 uses. Its pullback is `exp(p_ξ/ε)`, which has an essential singularity at `ε = 0`,
not a pole. So "meromorphic in the state variables" is simply not preserved by an `ε`-dependent
rescaling of the momenta. The counterexample is not itself a first integral, but it kills any purely
function-theoretic proof, and the direction of the failure is bad: given one meromorphic complete
set `{F₁,…,F_n}`, the set `{exp(F₁), F₂,…,F_n}` is also complete and involutive
(`{exp(F₁),F_j} = exp(F₁){F₁,F_j} = 0`) and independent where `exp(F₁) ≠ 0`. So a contradiction
argument cannot assume it was handed the well-behaved representative.

The paper does not close this either. Section 4 ends with "we apply Theorem 2.1 to show that the
meromorphic Hamiltonian system corresponding to (4.7) is not meromorphically integrable **such that
the first integrals depend meromorphically on ε near ε = 0**. Thus, we complete the proof of Theorem
1.1 for the spatial case." Theorem 1.1 as printed carries no such hypothesis. For the planar case
there is a separate route: the remark closing Section 3 applies a Levi-Civita-type regularization,
invokes Morales-Ruiz, Simó and Simon, [*Algebraic proof of the non-integrability of Hill's
problem*](https://doi.org/10.1017/S0143385704000999), ETDS 25 (2005) 1237–1256, to show a limiting
Hamiltonian `H̃` is meromorphically nonintegrable, and concludes that (3.18) "is also
meromorphically nonintegrable for `ε > 0` fixed". That is the paper's actual fixed-`ε` bridge, and
the roadmap does not mention Hill's problem, regularization, or that reference anywhere.

So the roadmap has to choose, and say which:

- restrict the summit's integral class to rational or algebraic functions on `𝒮₂`/`𝒮₃`, where the
  blow-up is a genuine algebraic rescaling and the transport lemma is provable (this weakens
  Theorem 1.1, so say so in the opening paragraph);
- prove a growth estimate showing first integrals of *this* system admit representatives with
  controlled behaviour under the scaling, and make that a numbered target with its own statement;
- formalize the regularization plus Hill's-problem route from the Section 3 remark, which means
  adding Levi-Civita or Birkhoff regularization and Morales-Ruiz–Simó–Simon as targets in Lane D or
  F and citing them.

Whichever you pick, milestone 29 should stop being a one-line lemma, and the acceptance-test list
should carry the chosen hypothesis. This is worth getting right now: everything in Lanes A through
F is in service of a summit whose final step currently has no proof.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+`q̇ᵢ = ∂H/∂pᵢ`, `ṗᵢ = -∂H/∂qᵢ`. Fix these signs before proving any coordinate formula.
+
+For complexification, do **not** choose a global square-root branch. Adjoin `u₁,u₂`
+with `uⱼ² = rⱼ²`, restrict to the resulting algebraic variety `𝒮₂` or `𝒮₃`, and write the
+vector field rationally using `uⱼ⁻³`. Remove the critical/collision locus where the
+projection ramifies and the rational vector field has a pole. Prove that the real
+positive-distance sheet recovers the physical equations.
+
+For Delaunay variables use the paper's actions and angles, including
+`H_K = -(1-μ)/(2 I₁²)` and `ω₁ = (1-μ)/I₁³`. Record every branch and period of the complex
```
kim-em: > For Delaunay variables use the paper's actions and angles, including
> `H_K = -(1-μ)/(2 I₁²)` and `ω₁ = (1-μ)/I₁³`.

:robot: Claude+Codex say:

Please do not pin the paper's §3 Delaunay normalization. It is internally inconsistent for
`μ ≠ 0`, so milestone 30 asks a contributor to prove a set of identities that cannot all hold.

Write `α = 1-μ`. From the paper's own generating function `W = I₂φ + χ` with
`χ = ∫_{r₋}^{r} (2α/ρ - α/I₁² - I₂²/ρ²)^{1/2} dρ` and turning points
`r± = I₁(I₁ ± √(I₁² - I₂²/α))`:

- `r₊ + r₋ = 2I₁²` so `a = I₁²`, and `r₊r₋ = a²(1-e²)` gives **`I₂ = I₁√α √(1-e²)`**, exponent
  `1/2`. (Or `|I₂|`, without the paper's `I₂ > 0` convention.)
- `H_K = -α/(2I₁²)` is right, but it makes `I₁ = L/√α` where `L = √(αa)` is the standard Delaunay
  action. The radial action integral is `∮p_r dr = 2π(I₁√α - I₂)`, so over one radial period
  `Δθ₁ = ∂_{I₁} ∮p_r dr = 2π√α`. **`θ₁` is an angle mod `2π√α`, not mod `2π`**, so it does not live
  on the `𝕋ᵐ` that Theorem 2.1 requires, and the orbital period is `2πI₁³/√α`, not
  `2π/ω₁ = 2πI₁³/α`.
- The paper's (3.5), `T = 2πp_φ³/(α²(1-e²)^{3/2})`, is the correct Kepler period `2πa^{3/2}/√α`
  exactly when `p_φ = I₂ = I₁√α√(1-e²)`. Equating it to `2π/ω₁` instead is what produces the
  `(1-μ)^{1/3}` at milestone 41.
- Numerically at `μ = 0.3`, `I₁ = 2`, `I₂ = 1.3`: `∮p_r dr = 2.34563906` against
  `2π(I₁√0.7 - I₂) = 2.34563907`; physical `T = 60.0787`; the paper's `2π/ω₁ = 71.8078`.
- Separately, the closed form printed for `χ` is not the integral it is set equal to. Writing `C(r)`
  for the printed right-hand side, `C'(r) = (I₁/√α)·p_r(r)` (checked at five parameter points), and
  `C` vanishes at `r₊` rather than `r₋`. The additive constant is harmless up to an angle
  translation; the multiplicative factor is not. Note this factor is `1` only when `I₁ = √α`, so
  this discrepancy is *not* repaired at `μ = 0`.

The blow-up does not rescue the convention: it rescales position, momentum and time by powers of
`ε`, never actions by `√α`, and `α` survives explicitly in the limiting Kepler equations. A time
rescaling cannot change an angle's lattice.

Suggested fix: define Delaunay intrinsically, as actions conjugate to `2π`-periodic angles, and
derive the constants. With `L = √α I₁` and `ℓ = θ₁/√α` you get `H_K = -α²/(2L²)`, `ω₁ = α²/L³`,
`I₂ = G = L√(1-e²)`. Then add a sentence to "Coordinate and sign conventions" saying the paper's
§3 constants must be re-derived rather than transcribed.

Please also do not describe this as "only three constants change". The rederivation propagates to
the closed primitive `χ`, the inverse functions `R` and `Φ`, the algebraization auxiliaries
`v₁,v₂,v₃`, and the exact prefactors in the order-`ε⁶` obstruction. The correction is an invertible
constant canonical rescaling, so the contour nonvanishing should survive and the theorem looks
repairable, but the printed proof is not correct as it stands and the roadmap should say so rather
than pin it.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    spatial problem; it is a modern extension, not part of Poincaré's classical planar
+    theorem and not a claim that one arbitrary additional spatial integral cannot exist.
+
+**Gate E.** Poincaré's exact classical planar theorem is available before any fixed-`μ`
+blow-up calculation begins. The parameter-dependent meromorphic checkpoint has also
+tested Lane C at first perturbative order without being conflated with that theorem.
+
+## Lane F: verify (A2) and reach the fixed-mass summit
+
+### F1. Planar contour calculation
```
kim-em: :robot: Claude+Codex say:

Transcribed faithfully from the paper, but see the comment on line 96: it contradicts the paper's
own `r±` formula unless `μ = 0`. With a `2π`-periodic mean anomaly the resonant torus is
`I₂* = I₁*√(1-e²)` for the chosen eccentricity. Re-derive rather than restate.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    at order `ε⁶`. Give certified remainder statements on a named collision-free complex
+    neighbourhood; a symbolic `O(ε⁷)` with no uniform domain is not an acceptance
+    criterion.
+29. Prove the transport lemma: a meromorphic complete set for the original fixed-`μ`
+    system pulls back to one meromorphic jointly in the blown-up state and `ε` near
+    `ε=0`. This is the logical bridge that allows Theorem 2.1 to prove Theorem 1.1.
+
+### D3. Delaunay variables
+
+30. Construct planar Delaunay action-angle variables on the elliptic, noncircular,
```
kim-em: > Prove the generating-function identities, local symplecticity, inverse map, periodicity, and
> `H_K=-(1-μ)/(2I₁²)`, `ω=((1-μ)/I₁³,0)`.

:robot: Claude+Codex say:

"periodicity" and the two displayed constants are exactly the pair that cannot both hold (line 96
comment). Whichever normalization you settle on, make this milestone name *which* angle is
`2π`-periodic and require that as the defining property, with the constants as consequences. That
is also the only version a reviewer can check without redoing the algebra.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    the inertial model. Prove Hamilton's equations, conservation of the Jacobi/Hamiltonian
+    integral, the `μ ↔ 1-μ` primary-exchange symmetry, and the `μ=0` Kepler limit.
+26. Construct `𝒮₂,𝒮₃`, their critical sets and collision-free smooth loci; lift the
+    Hamiltonians/vector fields and prove that they are rational/meromorphic there. Relate
+    the physical real sheet and the complex model exactly.
+
+### D2. Blow-up near a primary
+
+27. Near `(-μ,0[,0])`, implement
+    `x+μ=ε²ξ`, `y=ε²η`, `z=ε²ζ`, the momentum shifts/scalings, and
```
kim-em: > 27. Near `(-μ,0[,0])`, implement `x+μ=ε²ξ`, `y=ε²η`, `z=ε²ζ`, the momentum shifts/scalings, and
>     `t ↦ ε³t`. Prove that this is symplectic with the stated Hamiltonian rescaling

:robot: Claude+Codex say:

It is not symplectic. With `x = ε²ξ - μ`, `y = ε²η`, `pₓ = p_ξ/ε`, `p_y = p_η/ε - μ`,

```
pₓ dx + p_y dy = ε(p_ξ dξ + p_η dη) - με² dη,
```

the last term is exact, so the symplectic form pulls back to `ε·ω_new`. The map is conformally
symplectic with conformal factor `ε`. Hamilton's equations come back only after the `ε²` Hamiltonian
rescaling and the `ε³` time change, and the roadmap never states the Hamiltonian rescaling that
"the stated Hamiltonian rescaling" refers to. This is a convention pin that an implementor will
copy, and getting it wrong produces a Hamiltonian off by a power of `ε` in every subsequent order
count. Please state the exact relation (`Φ*ω = ε ω`, `H ↦ ε²H`, `t ↦ ε³t`) and make each of the
three a provable clause. The same applies to the spatial case in milestone 47.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    fundamental matrix. Define the monodromy representation and prove that changing the
+    fundamental matrix conjugates it.
+15. For the meromorphic function field of the surface, prove that analytic continuation
+    commutes with the derivation and hence every monodromy matrix belongs to the
+    differential Galois group. Reuse Mathlib's fundamental-groupoid/path machinery and
+    the Universal Covers roadmap where appropriate.
+
+### B4. Morales–Ramis–Simó and Ayoul–Zung
+
+16. Formalize the Morales–Ramis theorem: meromorphic Liouville integrability near a
```
kim-em: > Formalize the Morales–Ramis theorem: ... the identity component of the differential Galois group
> of the normal variational equation ...

:robot: Claude+Codex say:

"Normal variational equation" appears here and nowhere else. B1 (milestones 8–10) builds the first
VE, the higher VEs and Yagasaki's RVE, but nothing constructs the normal VE: restriction to
`ker dH` along the orbit, followed by the quotient by the flow direction `⟨X_H⟩`, with the induced
symplectic form on the quotient. That is a real object with real work in it (well-definedness of the
quotient, and invariance of the differential Galois group under the reduction). Per the roadmap's
own "everything is grounded, with no leaps" rule it needs to be a numbered target in B1, or
milestone 16 should be restated over the full VE with the normal reduction dropped.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    one-dimensional `ℚ`-space, equivalently
+    `ω(I*)/ω* ∈ ℤᵐ \ {0}` for a chosen positive fundamental frequency. Prove the
+    periodicity statements and the harmless period refinement `ω* ↦ ω*/N`.
+21. Compute the RVE's block-triangular fundamental matrix with primitives `Ξᵏ` and
+    `Ψᵏ`. Continue it around both the complex loop `γ_θ` and the real period loop.
+22. Generalize the `3×3` seed in `Suggested.lean` to the paper's block matrices
+    `M(C₁,C₂,C₃)`. Prove Yagasaki Lemma 2.5 using Zariski closures of powers: if
+    `C₃C₁' ≠ C₃'C₁`, the **identity component** is noncommutative. Merely exhibiting two
+    noncommuting monodromy elements is not enough.
+23. Formalize assumption (A2),
```
kim-em: :robot: Claude+Codex say:

Two things missing from the target statement.

First, Theorem 2.1 concludes *"not meromorphically integrable in the Bogoyavlenskij sense near the
resonant periodic orbit **such that the first integrals and commutative vector fields also depend
meromorphically on `ε` near `ε = 0`**"*. That clause is the whole reason milestone 29 exists (see
the comment there) and it is what makes the theorem provable at all. The scope section (line 57)
discusses it, but the milestone that states the theorem does not carry it. Same for Gate B
(line 260): Theorems 2.2 and 2.3 carry the identical hypothesis.

Second, (A2) is not just the displayed integral. It also asserts the existence of a closed loop
`γ_θ` in a region of `ℂ` containing `(0,T*)` with `γ_θ ∩ (iℝ ∪ (T*+iℝ)) = ∅`, and Theorem 2.1
quantifies over a domain `Γ ⊆ ℂ/T*ℤ` containing `ℝ/T*ℤ` and `γ_θ`. Milestone 44 handles the loop
geometry for the concrete case, but the general statement in Lane C should carry `Γ`, the loop, and
the continuation along it, otherwise milestone 23 is not the theorem.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+
+The roadmap has two summits, in historical and logical order. The first is Poincaré's
+classical theorem: the planar **circular restricted** three-body problem has no
+additional first integral that is real-analytic in the state variables, functionally
+independent of the Hamiltonian, and also real-analytic in the mass parameter `μ` near
+`μ = 0`. This parameter-analyticity hypothesis is essential to the statement; the
+theorem is not a fixed-`μ` result. Poincaré's original result is planar. A later
+parameter-dependent spatial result belongs to the modern checkpoint below.
+
+The final summit is Yagasaki's theorem that the same problem is meromorphically
```
kim-em: :robot: Claude+Codex say:

Small clarification rather than a change of plan. The summit is stated Liouville-style, matching
Theorem 1.1 verbatim, and Lane C correctly notes that the Bogoyavlenskij engine specializes to it.
One sentence at milestone 49 saying whether the deliverable is the Liouville statement only, or both
it and the stronger Bogoyavlenskij statement the proof route actually gives, would stop a
contributor wondering whether the Ayoul–Zung layer is meant to be visible in the final theorem.
Note that milestone 29's transport lemma is currently phrased for "a meromorphic complete set", so
if you want both, it needs the commuting-vector-field half too.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    restricting the full higher VE to the invariant subspace on which lower variations
+    vanish. Prove the `1/k! · ∂_ε^k f` normalization.
+
+### B2. Picard–Vessiot theory
+
+11. Starting from Mathlib's `Differential`/`DifferentialAlgebra`, define linear
+    differential systems over a differential field, fundamental matrices, solution
+    algebras, Picard–Vessiot rings and fields, and prove existence and uniqueness when the
+    constant field is algebraically closed.
+12. Define the differential Galois group as differential automorphisms over the base,
```
kim-em: > represent it faithfully as a Zariski-closed subgroup of `GLₙ(C)` ... Build the identity component
> of a linear algebraic group, prove it is normal and of finite index

:robot: Claude+Codex say:

This repo already has [Reductive algebraic groups](../ReductiveGroups/README.md), which maintains
three equivalent views of an affine algebraic group and explicitly targets the identity component
`G°` and component group `π₀(G)`, with a pinned convention that "connected" means geometrically
connected. Milestones 12–13 here would build a second vocabulary for the same objects.

I am not claiming everything must be consumed from there: that roadmap's scheme-theoretic layer may
be the wrong altitude for concrete matrix differential Galois groups, and a deliberately concrete
`GLₙ(ℂ)`-level treatment could be the better choice. But the roadmap should make that choice and
write it down, either a shared foundation or a documented compatible concrete layer. Right now it is
silent about a roadmap covering the same ground. I confirmed Mathlib has no linear-algebraic-group
API at `9caeba1`, so whatever is not shared is a numbered target here.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    `k₂(π+iK)/k₁`, integrating around a small circle, and proving the leading
+    `δ⁻³ᐟ²` coefficient is nonzero. Recover the planar meromorphic nonintegrability
+    statement near resonant elliptic orbits (Yagasaki Theorem 3.1).
+40. Extend the Delaunay and algebraization construction to the spatial family. Restrict
+    the obstruction calculation to the invariant equatorial plane and recover Yagasaki
+    Theorem 4.1. Record explicitly that this obstructs a complete involutive set in the
+    spatial problem; it is a modern extension, not part of Poincaré's classical planar
+    theorem and not a claim that one arbitrary additional spatial integral cannot exist.
+
+**Gate E.** Poincaré's exact classical planar theorem is available before any fixed-`μ`
```
kim-em: > **Gate E.** Poincaré's exact classical planar theorem is available before any fixed-`μ` blow-up
> calculation begins.

(and README.md:456, "Finish the exact classical Gate E before treating the fixed-`μ` blow-up
calculation in Lane F")

:robot: Claude+Codex say:

Lane F has no mathematical dependency on Lane E: Theorem 1.1's proof never invokes Poincaré's
theorem. So the ordering constraint buys nothing and risks a lot. Milestone 36 ("prove on an
explicit elliptic domain the historical nonvanishing/density condition") is the substantial part,
and Whittaker §165 does not obviously discharge the roadmap's own named-domain and
coefficient-density obligations; Yagasaki describes his later paper as giving "a very similar
result", not a reconstruction of the historical calculation.

I would keep Lane E as a full target and drop the ordering constraint from Gate E and from the
"Ordering and parallel work" paragraph, so a stall on the classical theorem does not stall the
summit. If you keep it, cite a source that discharges milestone 36 rigorously so the gate rests on
something.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+   symplectomorphism and restriction to a dense open set.
+
+### A3. Complex analytic category
+
+6. Build holomorphic and meromorphic scalar maps, vector fields, differential forms, and
+   local equivalences on finite-dimensional complex manifolds. For algebraic varieties,
+   reconcile chartwise meromorphicity with local quotients of regular/holomorphic
+   functions. Supply pullback, composition, equality-on-dense-open, and removable-pole
+   lemmas.
+7. Construct the field of meromorphic functions on a connected Riemann surface and its
```
kim-em: :robot: Claude+Codex say:

Milestone 11 requires an algebraically closed constant field for Picard–Vessiot existence and
uniqueness, and every application in B3 and C discharges that hypothesis for the meromorphic
function field of `Γ ⊆ ℂ/T*ℤ`. Milestone 7 stops at "the field of meromorphic functions ... and its
derivation". Add the computation: the constants of that field, **under the chosen nonzero
meromorphic derivation** (connectedness alone does not name a derivation), are `ℂ`. Milestone 11
should also pin characteristic zero. This is the hypothesis-discharging lemma for the whole B and C
tower and it is currently nobody's job.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+   powers, with a Faà di Bruno API. Derive higher variational equations and their
+   linearization. Do not encode the paper's displayed low-order coordinate expansion as
+   the definition.
+10. For `x' = f(x;ε), ε'=0`, derive Yagasaki's `k`th reduced variational equation by
+    restricting the full higher VE to the invariant subspace on which lower variations
+    vanish. Prove the `1/k! · ∂_ε^k f` normalization.
+
+### B2. Picard–Vessiot theory
+
+11. Starting from Mathlib's `Differential`/`DifferentialAlgebra`, define linear
```
kim-em: :robot: Claude+Codex say:

Milestone 11 is all of Picard–Vessiot; 12 is the whole differential Galois correspondence; 16, 17
and 18 are Morales–Ramis, Morales–Ramis–Simó and Ayoul–Zung. Each is a book chapter. A single
contour computation gets six numbered items (41–46).

Since claims are registered per numbered target, B2 and B4 are unclaimable in practice. Subdivide
to roughly Lane F's granularity. For B2 at minimum: linear differential systems and fundamental
matrices; PV ring existence; PV field uniqueness; the Galois group as differential automorphisms;
the faithful `GLₙ` representation; the correspondence. For B4, separate the cotangent lift, the
preservation of meromorphic integrability, the reduction principle, and each of the three theorems.

kim-em (commented):
:robot: Claude+Codex say:

Checked against Yagasaki arXiv:2106.04925v7, the published JMP abstract for arXiv:2111.11031, and
the pinned Mathlib (`9caeba1`, which is the local checkout, so the file and declaration references
were verified directly rather than from memory). `Suggested.lean` typechecks on that pin with four
`sorry` warnings and nothing else.

A lot of this roadmap is accurate and I am not commenting on it further: the sign convention
(`ẋ = pₓ + y`, `ṗₓ = p_y + ∂U/∂x`) reproduces the paper's (1.4) exactly; the mass and position
assignment is barycentric; `planarPotential_zero` and both `primaryExchange` statements are true as
written; `M(C₁,C₂,C₃)` and the `C₃C₁' ≠ C₃'C₁` criterion match Lemma 2.5 including the block shapes
`C₁ ∈ ℂ^ℓ`, `C₃ ∈ ℂ^{m×ℓ}`; the blow-up `ε²ξ = x+μ`, `ε⁻¹p_ξ = pₓ`, `ε⁻¹p_η = p_y + μ`, `t → ε³t`
and the ε³ Coriolis / ε⁴ cancellation / ε⁶ tidal orders are the paper's; `Σ(𝒮₂) = {u₁=0} ∪ {u₂=0}`;
Theorem 1.1 is Liouville-shaped and the one-more/two-more integral count is right; and the claim
that the published JMP version restricts Poincaré's attribution to the planar case is correct (the
published abstract says "in the planar case" where arXiv v2 said "in both the planar and spatial
cases"). The `Suggested.lean` policy of refusing to invent `Prop := sorry` placeholders is the
right call and the file follows it.

The items I would want addressed before merge, in order:

1. **Milestone 29's transport lemma is not proved anywhere and is false as stated** (README:313).
   It is the roadmap's declared bridge from Theorem 2.1 to Theorem 1.1, and the paper does not
   prove it either. Details inline; this is the one that changes what the roadmap can promise.
2. **The §3 Delaunay normalization is internally inconsistent** and the roadmap pins it as a
   package to prove (README:96, 319, 392).
3. **The blow-up is described as symplectic and is not** (README:305).
4. **The normal variational equation is used but never constructed** (README:248).
5. **Lane C and Gate B drop Theorem 2.1's parameter-meromorphic hypothesis and its complex-time
   setup** (README:279).

Then a set of API and scheduling points: the generic-independence predicate silently requires
ambient-open domains, Gate E's ordering constraint, the uncited Reductive Groups roadmap, the
missing constant-field lemma, and lane granularity.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+  algebraic variety.
+- Differential rings and fields through
+  [`Differential`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/RingTheory/Derivation/DifferentialRing.html#Differential),
+  `DifferentialAlgebra`, and `Differential.ContainConstants`, together with logarithmic
+  derivatives and the Rosenlicht/Liouville elementary-antiderivative theorem in
+  `Mathlib/FieldTheory/Differential/Liouville.lean`. This is a valuable algebraic base,
+  but it is **not** Picard–Vessiot theory.
+- General field Galois theory and substantial affine algebraic geometry. Reuse these for
+  differential automorphisms and Zariski-closed matrix groups.
+- The [Contour Integration roadmap](../ContourIntegration/README.md) supplies reusable
```
kim-em: :robot: Claude+Codex say:

Neither is a mathematical leap, but both are vague about ownership.

- Contour Integration does target "Piecewise-`C¹` curves, closed immersions, and cycles" and fixes
  `∫ t in a..b, γ' t • f (γ t)` as the contour integral. Name that layer and pin that milestone 44's
  Figure-5 path uses its convention, rather than "as it lands".
- Universal Covers is covering-space monodromy (`IsCoveringMap.monodromyFunctor`, `π₁(X)`-sets),
  which is not continuation of germs along paths. Either name the declarations milestone 14 expects
  to consume, or drop the reference.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    the calculation behind the phrase “Poincaré set is dense.”
+37. Package Poincaré's planar theorem. Include joint real-analyticity in `(state,μ)`, the
+    neighbourhood of `μ=0`, functional independence from the Hamiltonian, angle
+    uniformity, collision exclusions, and the precise phase domain in the theorem
+    statement. Do not silently strengthen the historical attribution to the spatial
+    problem.
+
+### E3. A modern parameter-dependent checkpoint
+
+38. Independently recover Yagasaki's parameter-dependent comparison result (arXiv 2022,
```
kim-em: :robot: Claude+Codex say:

Lane F cites (3.15), Figure 5 and Remark 3.1, which makes it checkable. E3 refers to "the complex
singular time `k₂(π+iK)/k₁`", "the leading `δ⁻³ᐟ²` coefficient", and Theorems 3.1 and 4.1 with no
equation numbers. Since the roadmap itself records that the published version differs from arXiv v2,
add the references and say which version's numbering is authoritative here.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+
+- Direct differentiation of `H₂,H₃` reproduces every displayed rotating-frame equation.
+- At `μ=0`, the potential and scaled unperturbed dynamics reduce to Kepler; the final
+  fixed-mass theorem itself is unavailable because its strict mass hypothesis fails.
+- The classical summit explicitly assumes joint real-analytic dependence on `μ` near
+  zero and excludes a single additional integral in the planar problem.
+- The parameter-dependent comparison theorem uses `ε=μ` and its first obstruction has
+  `k=0`; its spatial conclusion is not silently attributed to Poincaré or promoted from
+  failure of Liouville integrability to nonexistence of any one additional integral.
+- Rotation by `π` together with `μ ↔ 1-μ` exchanges the two primaries and conjugates the
```
kim-em: > Rotation by `π` together with `μ ↔ 1-μ` exchanges the two primaries and conjugates the systems.

:robot: Claude+Codex say:

The potential-level identities in `Suggested.lean` are both true (I checked), but the symmetry Lane F
uses at milestones 46 and 48 is the Hamiltonian one, and it needs the momenta:

```lean
theorem planarHamiltonian_primaryExchange (μ : ℝ) (s : PlanarState) :
    planarHamiltonian (1 - μ) ⟨-s.x, -s.y, -s.px, -s.py⟩ = planarHamiltonian μ s := by
  sorry
```

Worth pinning because the paper's own shorthand is a trap: it says "we only have to replace `x` and
`μ` with `-x` and `1-μ`". That works for the potential, but at Hamiltonian level `(x,y) ↦ (-x,y)`
flips the sign of the Coriolis term `pₓy - p_yx`, so it is a time-reversing symmetry, not the
conjugation milestone 46 wants. The half-turn is the clean one.

For the spatial case, distinguish the two maps and say which you mean. The literal rotation by `π`
about the `z`-axis is `(x,y,z,pₓ,p_y,p_z) ↦ (-x,-y,z,-pₓ,-p_y,p_z)`; the full inversion
`(q,p) ↦ (-q,-p)` also works, because the potential is even in `z`, but it is not a rotation in
three dimensions. `spatialPotential_primaryExchange` currently pins the second. The `z`-axis version
is the one that matches the acceptance test's wording.

TauCetiRoadmap/RestrictedThreeBody/Suggested.lean
```diff
@@ -0,0 +1,168 @@
...
+def FunctionallyIndependentAt {n ι : ℕ} (F : Fin ι → ComplexState n → ℂ)
+    (x : ComplexState n) : Prop :=
+  LinearIndependent ℂ fun i ↦ fderiv ℂ (F i) x
+
+/-- Pointwise independence of vector fields. -/
+def VectorFieldsIndependentAt {n q : ℕ} (X : Fin q → ComplexVectorField n)
+    (x : ComplexState n) : Prop :=
+  LinearIndependent ℂ fun i ↦ X i x
+
+/-- Independence on a dense relatively open locus of `U`. This is the roadmap convention for the
```
kim-em:
```lean
def GenericallyFunctionallyIndependentOn {n ι : ℕ}
    (F : Fin ι → ComplexState n → ℂ) (U : Set (ComplexState n)) : Prop :=
  ∃ V : Set (ComplexState n), IsOpen V ∧ V ⊆ U ∧ U ⊆ closure V ∧ ...
```

:robot: Claude+Codex say:

`IsOpen V` is openness in the ambient `ℂⁿ`, so `V ⊆ U` forces `V ⊆ interior U`. On any `U` with
empty interior this gives `V = ∅` and then `U ⊆ closure ∅ = ∅`. So both generic-independence
predicates are false on every nonempty set with empty interior, and `IsBogoyavlenskijIntegrableOn`
is unconditionally false there. Verified against the pinned Mathlib:

```lean
example {n : ℕ} (q : ℕ) (X₀ : ComplexVectorField n) (U : Set (ComplexState n))
    (hi : interior U = ∅) : ¬ IsBogoyavlenskijIntegrableOn q X₀ U := by
  rintro ⟨-, -, hU, X, F, -, -, -, hind, -⟩
  rcases hind with ⟨V, hVo, hVU, hUV, -⟩
  have : V ⊆ interior U := hVo.subset_interior_iff.mpr hVU
  rw [hi, Set.subset_empty_iff] at this
  subst this
  simp only [closure_empty, Set.subset_empty_iff] at hUV
  exact absurd hUV hU.ne_empty
```

The file says this is an early flat-space model and that the manifold version comes later, so this
does not by itself make milestone 49 vacuous. But the surrounding signals point the other way: the
definitions take `fderivWithin`/`lieBracketWithin` over `U`, which only matters for non-open `U`,
and the README's convention at line 64 ("nonempty dense open locus") is stated for the roadmap as a
whole, including the summit on `𝒮₂`/`𝒮₃`. If that convention is carried forward unchanged, the
summit becomes vacuously true on exactly the varieties it is about.

Fix either way, but decide: add `IsOpen U` and say in the docstring that this predicate is only for
flat open domains, or use relative openness and relative density so the predicate is reusable on
subsets. Then make the README's line-64 convention say which one it means.

TauCetiRoadmap/RestrictedThreeBody/Suggested.lean
```diff
@@ -0,0 +1,168 @@
...
+carried separately by theorems: Mathlib's total `fderiv` is zero where no derivative exists. -/
+def IsFirstIntegral {n : ℕ} (X : ComplexVectorField n) (F : ComplexState n → ℂ) : Prop :=
+  ∀ x, fderiv ℂ F x (X x) = 0
+
+/-- The domain-restricted first-integral predicate used by local integrability. -/
+def IsFirstIntegralOn {n : ℕ} (X : ComplexVectorField n) (F : ComplexState n → ℂ)
+    (U : Set (ComplexState n)) : Prop :=
+  ∀ x ∈ U, fderivWithin ℂ F U x (X x) = 0
+
+/-- Two flat vector fields commute when their Mathlib Lie bracket vanishes. -/
```
kim-em:
```lean
def Commute {n : ℕ} (X Y : ComplexVectorField n) : Prop :=
  VectorField.lieBracket ℂ X Y = 0
```

:robot: Claude+Codex say:

Inside `namespace TauCetiRoadmap.RestrictedThreeBody` this shadows `_root_.Commute`, and the failure
is not hypothetical:

```lean
example (a b : ℕ) : Prop := Commute a b
-- error: don't know how to synthesize implicit argument `n`
--   @Commute (?m.4 a b) ↑a ↑b
```

It tries to coerce `ℕ → ℂ` and elaboration dies. `VectorField.lieBracket ℂ X Y = 0` needs no
abbreviation at all; otherwise use a name like `LieBracketEqZero`. Same for `CommuteOn`.

TauCetiRoadmap/RestrictedThreeBody/Suggested.lean
```diff
@@ -0,0 +1,168 @@
...
+
+/-- Bogoyavlenskij `(q,n-q)`-integrability on `U`, before imposing analytic or meromorphic
+regularity: a generically independent commuting family containing the system vector field and a
+generically independent family of common first integrals. The meromorphic manifold version is a
+later Lane-A3 target. -/
+def IsBogoyavlenskijIntegrableOn {n : ℕ} (q : ℕ) (X₀ : ComplexVectorField n)
+    (U : Set (ComplexState n)) : Prop :=
+  0 < q ∧ q ≤ n ∧ U.Nonempty ∧
+  ∃ X : Fin q → ComplexVectorField n, ∃ F : Fin (n - q) → ComplexState n → ℂ,
+    (∃ i, X i = X₀) ∧
```
kim-em:
```lean
    (∃ i, X i = X₀) ∧
```

:robot: Claude+Codex say:

Everything else in `IsBogoyavlenskijIntegrableOn` is relative to `U`, but this asks for equality of
vector fields on all of `ℂⁿ`. Two fields agreeing on `U` should count. Use `Set.EqOn (X i) X₀ U`.

TauCetiRoadmap/RestrictedThreeBody/Suggested.lean
```diff
@@ -0,0 +1,168 @@
...
+
+/-- An autonomous vector field on `ℂⁿ`. The manifold API later generalizes this flat model. -/
+abbrev ComplexVectorField (n : ℕ) := ComplexState n → ComplexState n
+
+/-- `F` is a first integral of `X` when its differential annihilates `X` everywhere. Regularity is
+carried separately by theorems: Mathlib's total `fderiv` is zero where no derivative exists. -/
+def IsFirstIntegral {n : ℕ} (X : ComplexVectorField n) (F : ComplexState n → ℂ) : Prop :=
+  ∀ x, fderiv ℂ F x (X x) = 0
+
+/-- The domain-restricted first-integral predicate used by local integrability. -/
```
kim-em: :robot: Claude+Codex say:

The docstring on `IsFirstIntegral` notes that Mathlib's total `fderiv` is junk where no derivative
exists, and defers regularity to theorems. But `IsBogoyavlenskijIntegrableOn` is the predicate that
the summit will be stated against, and it contains no differentiability or holomorphicity hypothesis
anywhere: `fderivWithin` and `lieBracketWithin` are junk both when `F` is not differentiable and
when `U` fails `UniqueDiffWithinAt`, so a discontinuous `F` satisfies `IsFirstIntegralOn`. Since
Gate A promises a statement "without placeholders", the README should say which regularity the real
definition carries (holomorphic on `U`, presumably) and the predicate should carry it, or be phrased
with derivative witnesses (`HasFDerivWithinAt`) so the hypothesis cannot go missing.

TauCetiRoadmap/RestrictedThreeBody/README.md
```diff
@@ -0,0 +1,503 @@
...
+    real-analytic first integral `F(μ)` on a named action-angle domain. State the exact
+    classical target before proving it: for the planar CR3BP there is no such `F`,
+    functionally independent of `H`, near `μ=0`. Make clear what uniformity in the
+    angles and what phase-space domain are required.
+34. Expand `H=H₀+μH₁+⋯` and `F=F₀+μF₁+⋯`, justify termwise Poisson brackets, and derive
+    every homological equation from `{H,F}=0`. Formalize the normalization used by
+    Poincaré and Whittaker: after subtracting a locally analytic function of `H` and
+    dividing by the first possible power of `μ`, a genuinely new integral may be assumed
+    to have `F₀` not locally a function of `H₀`.
+35. Prove the abstract Poincaré-set lemma. Fourier expansion in the angles first makes
```
kim-em: :robot: Claude+Codex say:

The step from "`DF₀` collinear with `DH₀`" to "`F₀` is locally a function of `H₀`" needs `DH₀ ≠ 0`,
or more precisely a submersion or constant-rank hypothesis on the domain. It is automatic for the
CR3BP (`DH₀ = (I₁⁻³, -1)`), but milestone 35 is stated as an abstract lemma, so the hypothesis
belongs in it.

kim-em:
Please remove the `awaiting-author` label and replace it with `awaiting-review` when you think you've dealt with the review comments.

Please also reply to all my review threads, and mark any as resolved where you're confident you've dealt with them correctly.

mrdouglasny:
There is a lot of good material here, but the goal is quite specialized.
Why not start with a roadmap on classical mechanics!
It is hardly represented in mathlib (or in physlib for that matter).
Then this roadmap can be downstream of it.

Here is a zeroth draft (Claude Opus) based on my favorite textbook.

# Brief: author a TauCeti roadmap for **Classical Mechanics** (à la Arnold)

## Task

Write a **broad, reusable** roadmap for classical mechanics — the general Lagrangian / Hamiltonian / symplectic theory — as a Tau Ceti roadmap. This is the *foundational* roadmap that specific results consume: the open PR **#151** (Poincaré nonintegrability of the planar CR3BP) should become a **downstream successor**, and **#94** (geodesics / Hopf–Rinow) is a **Riemannian-side sibling** (geodesic flow is Hamiltonian on `T*M`). Prefer breadth: cover what a graduate classical-mechanics course covers (Arnold), not a single theorem.

Deliverables:
- `TauCetiRoadmap/ClassicalMechanics/README.md` — the definitive narrative roadmap.
- `TauCetiRoadmap/ClassicalMechanics/Suggested.lean` — proposed core definitions + one milestone
  `sorry`-goal per layer, elaborating green against the pinned Mathlib.
- Register the roadmap in the root `README.md` list (alphabetical) and add the `import` line to `TauCetiRoadmap.lean`.

## Mathlib coverage (v4.31.0-rc1) — verified against source

**Prerequisites already in Mathlib (consume; do NOT rebuild):**
- Smooth manifolds, `ContMDiff`, `mfderiv`, `Diffeomorph`, tangent bundle — `Mathlib.Geometry.Manifold.*`, `…/VectorBundle`.
- Vector fields + **Lie bracket** — `Mathlib.Geometry.Manifold.VectorField.LieBracket` (+ `Pullback`).
- **Differential forms + exterior derivative** `extDeriv`, with `extDeriv_extDeriv` (d∘d = 0) — `Mathlib.Analysis.Calculus.DifferentialForm.*`. (Check wedge/exterior product coverage; it may be partial.)
- **Integral curves** + ODE theory (`PicardLindelof`, `ExistUnique`, `Gronwall`), flows — `Mathlib.Geometry.Manifold.IntegralCurve`, `Mathlib.Analysis.ODE.*`, `Mathlib.Dynamics.Flow`.
- **Lie groups / Lie algebras**, left-invariant derivations — `Mathlib.Geometry.Manifold.Algebra.LieGroup`, `Mathlib.Algebra.Lie.*`.
- Symplectic **linear** algebra (matrix `Sp(2n)`) — `Mathlib.LinearAlgebra.SymplecticGroup`.
- Riemannian **metric** class `IsRiemannianManifold`, path length — `Mathlib.Geometry.Manifold.Riemannian.{Basic,PathELength}`.

**Gaps this roadmap builds (mechanics proper — none of this is in Mathlib):**
- Symplectic **manifolds**: a closed nondegenerate 2-form `ω` on a manifold (only the linear `Sp` group exists).
- **Hamiltonian vector fields** `X_H` (`ι_{X_H} ω = dH`) and **Poisson brackets** `{f,g}`.
- **Lagrangian mechanics**: Lagrangian on `TM`, the action functional, **Euler–Lagrange** equations (needs a first-variation / calculus-of-variations lemma).
- **Legendre transform** (fiber derivative `TM → T*M`; no convex/Fenchel conjugate in Mathlib either).
- **Noether's theorem** / **moment maps** / conserved quantities.
- **Integrable systems**: first integrals in involution, **Liouville–Arnold**, action–angle variables.
- **Symplectic Darboux** normal form. ⚠ `Mathlib.Analysis.Calculus.Darboux` is the *real-analysis* Darboux (IVP for derivatives), NOT this — do not cite it.
- Canonical transformations, generating functions, **Hamilton–Jacobi**.
- (Riemannian side, → #94) geodesics, exponential map, Hopf–Rinow.

## Proposed layer structure (refine against Arnold)

Each layer: *Definitions* → one *Milestone* (`sorry`-goal in `Suggested.lean`) → *Discharge* sketch →
*Companions to build*. Definitions are the chief deliverable (get them right, at full generality).

- **L0 — Symplectic geometry (the substrate).** Symplectic manifold `(M, ω)` (`ω` closed, nondegenerate); Hamiltonian vector field `X_H`; Poisson bracket `{f,g}`. *Milestone:* `{·,·}` is a Lie bracket on `C^∞(M)` (Jacobi from `dω = 0`) **and** the Hamiltonian flow preserves `ω` and `H` (Liouville + energy conservation). *Consumes:* `DifferentialForm`/`extDeriv`, `VectorField.LieBracket`, integral curves.
- **L1 — Lagrangian mechanics.** Lagrangian `L : TM → ℝ`, action functional, **Euler–Lagrange**. *Milestone:* E–L as the first-variation critical-point condition; **Noether** (1-param symmetry → conserved quantity). *Companion to build:* first-variation lemma (calculus of variations).
- **L2 — Legendre transform / Lagrangian ↔ Hamiltonian.** Fiber derivative, hyperregular `L`. *Milestone:* E–L equations ⟺ Hamilton's equations under the Legendre transform.
- **L3 — Hamiltonian mechanics & canonical structure.** Hamilton's equations; **canonical transformations = symplectomorphisms**; generating functions; Hamilton–Jacobi. *Milestone:* symplectomorphisms preserve Hamilton's equations / the Poisson bracket.
- **L4 — Integrable systems (summit).** First integrals in involution; Liouville integrability. *Milestone:* **Liouville–Arnold** — `n` independent integrals in involution ⇒ compact level sets are Lagrangian tori, with action–angle coordinates. *Consumes:* L0, L3.
- **L5 — (optional / successor-facing) perturbation theory.** Averaging, resonances, **KAM** named as a deeper target. This is where **#151 (CR3BP nonintegrability)** attaches as a successor; KAM itself may be its own future roadmap.

Sibling/boundary notes to state clearly (Kim's "clear boundaries"): **#94** geodesics/Hopf–Rinow is the Riemannian companion (geodesic flow = Hamiltonian flow of the kinetic energy on `T*M`); **#151** CR3BP is a downstream consumer of L4/L5; rigid-body/tops, small oscillations, adiabatic invariants are in-scope applications if breadth is wanted.

## References
- V. I. Arnold, *Mathematical Methods of Classical Mechanics* — **primary**.
- R. Abraham & J. Marsden, *Foundations of Mechanics* — symplectic formulation, detail.
- J. Marsden & T. Ratiu, *Introduction to Mechanics and Symmetry* — moment maps, Noether, reduction.
- V. I. Arnold, V. Kozlov, A. Neishtadt, *Mathematical Aspects of Classical and Celestial Mechanics* — integrable systems, perturbation theory, links to #151.

kim-em:
I see you've pushed some commits, but could you please reply to the review comments and/or resolve them as appropriate, so I can do another round of review

PR #156, opened 2026-08-05
Add CFSG statement roadmap

kim-em:
This PR adds the `CFSGStatement` roadmap, whose deliberately narrow goal is to make the classification of finite simple groups *statable*: every group on the CFSG list becomes an honest Lean type with a `Group` instance, and the endpoint is the named proposition

```lean
universe u

def ClassificationStatement : Prop :=
  ∀ (G : Type u) [Group G] [Finite G] [IsSimpleGroup G],
    ∃ i : CFSGIndex, Nonempty (G ≃* i.Group)
```

There is no proof of this proposition in scope, and no finiteness, simplicity, or index-uniqueness obligation for a listed candidate. The roadmap says where that theory is meant to come from instead of leaving it unsaid.

The roadmap separates independently reviewable workstreams for indices, pinned Chevalley-Demazure groups, ordinary and graph Steinberg maps, Suzuki-Ree maps, fixed-point groups, sporadic presentation data, final assembly, and the identification with Mathlib's Suzuki construction. Carrier-valued definitions take `ValidLieTypeIndex`, the half-Frobenius takes `SuzukiReeIndex`, and the diagram permutation and graph automorphism take `GraphTwistedIndex`, so no branch of any of the three needs a value invented to fill a hole.

`ValidLieTypeIndex.dynkinType` names the root-systems roadmap's `DynkinType`, `dynkinType_valid` proves it valid, and `rank` is defined as `dynkinType.rank`, so every `Fin d.rank` here indexes the pinned Bourbaki numbering and a second copy of a root datum is not expressible. `Suggested.lean` imports that roadmap rather than restating it. The long and short simple roots come from `DynkinType.IsLongSimpleRoot` for the same reason, so the exceptional exponent convention is stated once against upstream rather than tabulated per family.

L0 consumes Layer 6 of the root-systems roadmap and Layer 9 of the reductive-groups roadmap. Layer 9 owns the special isogenies in characteristics two and three, including `τ² = Frob_p`, so L2 keeps only the selection of that isogeny, its odd powers, and the finite-group naming; and L1 cites Layer 9's isomorphism theorem for pinned groups for the uniqueness of the graph automorphism.

Conventions are pinned as data. The diagram permutations are `Fin` permutations with the Bourbaki-to-zero-based offset stated. `Closure` is `AlgebraicClosure (ZMod p)`, with `characteristic_prime` and the `Fact` instance it needs, so the field and algebraically-closed instances come from Mathlib.

Sporadic presentations are stored as a relator expression type with powers and commutators, compiled to signed words by `Relator.toWord`; `Relator.toWord_toFreeGroup` is a target, since that compilation sits between the source a reviewer checks and the group that gets defined. There is no checksum field, because a checksum with no normal form and no function to recompute it cannot be checked; the generator and relator counts can be and are, and the arity comes from the generator-name list alone. Sources are named, the Monster getting the `Y₄₄₃` presentation with the spider relation and `Z = 1`, due to Norton and simplified by Ivanov. Several sporadics have no presentation a link away, so finding an admissible source for each of the twenty-six is the first S0 task, and the roadmap says what happens to a name that resists rather than allowing S0 to close over it.

The roadmap is explicit that the endpoint polices less than it appears to: the proposition depends only on the set of isomorphism types in the image, so it pins no name and transcription correctness is a review obligation. The cross-check against https://github.com/KitaKen1/finite-simple-groups-lean, which builds fourteen of the sporadics as permutation groups with order and simplicity proved, is a required S1 artifact.

Checked with `lake env lean TauCetiRoadmap/CFSGStatement/Suggested.lean`, `lake build`, and `python3 .github/scripts/check_roadmap_areas.py`.

🤖 Prepared with Claude Code


TauCetiRoadmap/CFSGStatement/README.md
```diff
@@ -0,0 +1,506 @@
...
+`Z = 1` for the central involution `Z`, giving 12 generators and 80 relators. The theorem is
+Norton's (1990), simplified by Ivanov (1999); the explicit word list is on J. N. Bray's
+[Monster presentation page](https://webspace.maths.qmul.ac.uk/j.n.bray/web/Pres/Mnst.html), which
+displays `M × 2` and becomes `M` on adding `Z = 1`. Farooq, Norton, and Wilson,
+[*A presentation of the Monster and a set of matrices which satisfy it*](https://doi.org/10.1016/j.jalgebra.2013.01.015),
+J. Algebra **379** (2013) 432-440, is an alternative.
+
+Apply the same standard to the Baby Monster, whose `Y₄₃₃` presentation was conjectured in the ATLAS
+and proved by Ivanov, and to every other branch. Bray's
+[presentation pages](https://webspace.maths.qmul.ac.uk/j.n.bray/web/Pres/) carry most of the
```
adomani: It looks like most of the sporadic groups in Bray's list actually defer the presentation to the Atlas.  For instance

https://webspace.maths.qmul.ac.uk/j.n.bray/web/Pres/BM.html
https://webspace.maths.qmul.ac.uk/j.n.bray/web/Pres/Co1.html
https://webspace.maths.qmul.ac.uk/j.n.bray/web/Pres/Co2.html
https://webspace.maths.qmul.ac.uk/j.n.bray/web/Pres/Co3.html

and so on.  Maybe this should point directly to the Atlas?

https://brauer.maths.qmul.ac.uk/Atlas/v3/

kim-em: Thanks!

TauCetiRoadmap/CFSGStatement/README.md
```diff
@@ -0,0 +1,612 @@
...
+
+`LieTypeIndex.InStandardRange` pins the usual rank and small-field restrictions. It starts `B` at
+rank two and excludes `B₂(2)`; starts `C` at rank three and restricts it to odd characteristic,
+leaving `B₂(q) = C₂(q)` and the characteristic-two overlap `B_n(q) = C_n(q)` to the `B` family;
+starts `D` and `²D` at rank four; excludes `A₁(2)`, `A₁(3)`, `²A₂(2)`, and `G₂(2)`; and starts each
+Suzuki--Ree parameter at `m = 1`. The separate Tits constructor supplies `²F₄(2)'`.
+
+Those last exclusions are of two kinds, and the distinction is worth keeping straight because the
+predicate is doing two jobs. `A₁(2)`, `A₁(3)`, and `²A₂(2)` are degenerate: the uniform recipe below
+sends them to the trivial group. `B₂(2)` and `G₂(2)` are duplicates: the recipe sends them to `A₆`
```
Paul-Lez: Codex Sol 🤖 : **[blocking]** This is false for `A₁(3)`: here `H = SL₂(𝔽₃)`, `H' = Q₈`, and therefore `H'/Z(H') ≅ C₂ × C₂`, not the trivial group. The `²A₂(2)` output is also nontrivial. These cases should be excluded because the resulting groups are not simple, rather than because the construction yields the trivial group.

TauCetiRoadmap/CFSGStatement/README.md
```diff
@@ -0,0 +1,612 @@
...
+development cited below covers fourteen of the twenty-six names, with order and simplicity proved, so
+for those names a reviewer can ask whether the transcribed relators hold of its generators, which is
+a check on the relators that no amount of reading the source twice provides. Recording the outcome of
+that comparison, for every name it covers, is part of closing an S1 row. It is a review artifact
+rather than a Lean target, since making it one would mean importing that development, which the
+provenance section conditions on coordination with its author.
+
+That gap is survivable, and it is worth saying why, because the reason is the whole argument for
+resting on review here. A transcription error can only make `ClassificationStatement` false, never
+easier to prove. If a branch is not the group it is named for, then the group it is named for is
```
Paul-Lez: Codex Sol 🤖 : **[blocking]** `ClassificationStatement` records only carrier isomorphism types, not their intended names. Swapping two sporadic presentation branches would leave the proposition unchanged while making both branches incorrectly named. Thus transcription correctness is an independent review obligation and is not enforced by the endpoint.

TauCetiRoadmap/CFSGStatement/Suggested.lean
```diff
@@ -0,0 +1,587 @@
...
+    LieTypeIndex.IsDuplicateRepresentative, PrimePower.card, q3]
+/-- `B₂(2) = Sp₄(2)` is not simple; its derived subgroup is `A₆`, kept as `alternating 6`. -/
+example : ¬(LieTypeIndex.B 2 q2).InStandardRange := by
+  norm_num [LieTypeIndex.InStandardRange, PrimePower.card, q2]
+/-- `B₂(4)` is kept: the rank-two symplectic family is named `B`, not `C`, in every
+characteristic. -/
+example : (LieTypeIndex.B 2 q4).Valid := by
+  norm_num [LieTypeIndex.Valid, LieTypeIndex.InStandardRange,
+    LieTypeIndex.IsDuplicateRepresentative, PrimePower.card, q4]
+/-- `B₂(3) = Sp₄(3) ≅ PSU₄(2) = ²A₃(2)`, and we keep the unitary name. -/
```
Paul-Lez: Codex Sol 🤖 : This should be `PSp₄(3) ≅ PSU₄(2)`, not `Sp₄(3) ≅ PSU₄(2)`. The latter is the double cover, whereas the projective quotient matches the construction below.

Paul-Lez (changes requested):
Codex Sol 🤖 : Requesting changes for two mathematical inaccuracies in the roadmap. Additional non-blocking points:

- The PR description is stale: it names `ExceptionalLieTypeIndex` instead of `SuzukiReeIndex`, describes #157 and #158 as proposed although they are merged, and its claim about Bray’s presentation pages no longer matches the README.
- `Suggested.lean` contains 28 lines exceeding 100 columns.
- `Sp₄(3)` should be `PSp₄(3)`, as noted inline.

mrdouglasny (dismissed):
*(Disclosure: co-written by MRD and Claude.)*

Approving. I went through the current head (`2597647`) against the earlier blocking review and the mathematics, and it all checks out.

- **Small-index exclusions** (the `A₁(3)` / `²A₂(2)` blocker) are now correct, and precise: `A₁(2), A₁(3), ²A₂(2), ²B₂(2)` are excluded as "the recipe does not produce a simple group," split into the two whose derived subgroup is its own centre (→ trivial: `SL₂(2)′ = C₃`, `Suz(2)′ = C₅`) and the two that do not collapse (`SL₂(3)′ = Q₈ → C₂×C₂`; `SU₃(2)` → order-18 solvable) — with the right caveat that the recipe is taken on the simply-connected `SL₂`/`SU₃`, not the projective form. `B₂(2)`, `G₂(2)`, `²G₂(3)` are correctly the duplicate-name exclusions (`A₆`, `²A₂(3)`, `A₁(8)`).
- **Transcription vs. naming** (the carrier-vs-names blocker): S1 now makes independent transcription review plus the recorded cross-check an explicit obligation, which is the right way to handle the fact that the existential endpoint pins carriers, not names.
- **`PSp₄(3) ≅ PSU₄(2)`** is fixed (with the double-cover note), and the sporadic sourcing now distinguishes ATLAS presentation pages from generator-recognition and requires a source that *proves* the presentation defines the group.
- The rest reads correctly — the 16 Lie-type families plus the Tits group, the coincidence table, and the 26 sporadics — and CI is green.

The only residual is cosmetic (a handful of >100-column docstring lines, which CI accepts), not worth holding this on. Nice roadmap — the "definitions must reduce to explicit data, no `Classical.choose`-ing a carrier from the target theorem" discipline is exactly right.

PR #165, opened 2026-08-06
doc: state the Mathlib relationship and the missing roadmap principles

kim-em:
This PR states the three rules governing Tau Ceti's relationship to Mathlib, and folds into the roadmap checklist the principles from the [Getting started: roadmaps](https://leanprover.zulipchat.com/#narrow/channel/610393-Tau-Ceti/topic/Getting.20started.3A.20roadmaps) Zulip topic that it did not yet spell out.

The Mathlib rules become a new "Working with Mathlib" group. Defer to Mathlib: where it has decided something, or is visibly deciding it, that decision is ours too and we refactor onto whatever lands. But never wait: an open Mathlib PR covering ground a roadmap needs is not a blocker and not a reason to leave a gap, so build the thing here, now, named and shaped to match, and delete ours if theirs lands. And never push work to Mathlib: plenty of Tau Ceti material would make good Mathlib material and Mathlib contributors are welcome to take any of it, but deciding what Mathlib absorbs is solely theirs, so no roadmap item is written as "to be upstreamed" or held back because it "really belongs in Mathlib".

From the Zulip topic: every item must be unambiguous, roadmaps need clear boundaries, roadmaps must be definite about scope, and roadmaps are timeless, meaning no calling an item "blocked" because an earlier roadmap is still being implemented, no war stories about why a roadmap changed, and no references to the review process. Material should aim to be reusable, and depth versus breadth stays the author's call, with the reasons for preferring breadth recorded. Kevin's point that a definition is not a contribution until it has API joins "build the library, don't race to the theorem".

To stop the section growing, the porting and prototyping bullets are grouped under headings, the overlapping "nothing is optional" and "do things right the first time" bullets are merged, and the vocabulary bullet is tightened. Six principles are added for about thirty lines.

Also adds `.claude/CLAUDE.md`, with `AGENTS.md` as a symlink to it, so an agent entering the repository is pointed at the checklist and at `CONTRIBUTING.md` before it does anything else. It carries one further instruction: when reviewing a pull request, skip the generic praise and go straight to the defects.

Written with Claude Opus 5.

🤖 Prepared with Claude Code

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,17 @@
+# Tau Ceti roadmaps
+
+Read [`README.md`](../README.md) before doing anything in this repository. It is normative, not
+background: the **Writing a roadmap** section is the standing checklist that all roadmap content is
+held against, and it covers how we relate to Mathlib (defer to Mathlib, never wait for it, never
+push work to it).
+
+Read [`CONTRIBUTING.md`](../CONTRIBUTING.md) as well before opening or reviewing a pull request.
+
+## Reviewing a pull request
+
+Generic praise is not helpful and wastes the author's time. Don't open with an assessment of how
+good the PR is, don't summarise back what it does, and don't pad each finding with a compliment.
+Get straight to the defects that need fixing, most important first, and say concretely what is
+wrong and what would fix it. Specific praise is fine where it carries information, for example that
+a design choice resolves a problem the author raised or that a proof is already the right shape;
+"this is nice work" carries none.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+.claude/CLAUDE.md
--- a/README.md
+++ b/README.md
@@ -64,58 +64,94 @@ A roadmap is a specification for material we want added to Tau Ceti, written so 
 reviewers, can act on it without guessing.
 
 - **Build the library, don't race to the theorem.** For each object you introduce, ask for its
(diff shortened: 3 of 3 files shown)


.claude/CLAUDE.md
```diff
@@ -0,0 +1,17 @@
...
+
+Read [`README.md`](../README.md) before doing anything in this repository. It is normative, not
+background: the **Writing a roadmap** section is the standing checklist that all roadmap content is
+held against, and it covers how we relate to Mathlib (defer to Mathlib, never wait for it, never
+push work to it).
+
+Read [`CONTRIBUTING.md`](../CONTRIBUTING.md) as well before opening or reviewing a pull request.
+
+## Reviewing a pull request
+
```
Paul-Lez, line 11: One thing I've started doing recently to make sure my agents don't write overly verbose prose/documentation it to ask them to follow the ASD-STE100 Simplified Technical English standard whenever possible. 

Do you think this would be a reasonable thing to add here? (e.g. could help make roadmaps written with the help of an agent more concise, etc)

CBirkbeck (commented):
I guess if something is in an open PR to mathlib what we could do is use what's there so that once it's gets merged (even if it changes a bit) it'll be easier to refactor. I'd do this before building here. Saying building here makes me worry the agents will make it from scratch.

Edit: ah maybe the sentence below says to do exactly this.

CBirkbeck (approved):

PR #184, opened 2026-08-10
roadmap(EllipticCurves): global models, semistability, and selected ℚ-adapters

CBirkbeck:
> **Disclosure:** This pull-request description was written and posted by Codex on behalf of CBirkbeck.

## Summary

This expands the elliptic-curves roadmap with the global-model infrastructure and selected `ℚ`-specific adapters needed by current downstream consumers. It separates model-independent invariants from global equation construction and records the exact contracts consumed at each boundary.

## Main additions

- Layer 3 names `frobeniusTrace`, defines ordinary and supersingular curves geometrically, and states invariance under arbitrary field extension.
- Layer 4.5a develops global and semi-global minimality, the minimal discriminant ideal, the positive defect class, semistability, and the reduced-minimal predicate over Dedekind domains.
- Layer 4.5b develops global equation construction over number fields, including Kraus's local/global contracts, the global-minimality equivalence, reduced-minimal equations over `ℚ`, and the sharp construction at a supplied representative prime.
- Layer 8 supplies selected `ℚ` adapters: primitive integral short equations, curve height, bounded-height finiteness, `abcQuality`, and model-dependent integral points.
- The reduced equation has a distinguished equation-level differential; transport to another presentation is canonical only up to automorphism and sign, while the absolute real period is unaffected.

## External integration

The elliptic-curve theorem starts from a prime representing the global-minimality class and lying away from `6`. An unconditional finite-avoidance corollary additionally requires the general number-field contract `NumberField.exists_primeIdeal_mk_eq_avoiding`; ownership, proof, and analytic prerequisites for that integration contract are outside the Elliptic Curves roadmap.

## Validation

- `lake -Kjobs=1 build`
- `lake -Kjobs=1 build TauCetiRoadmap.EllipticCurves.Suggested`
- `git diff --check`

## Provenance and AI assistance

The mathematical routes follow Silverman, Kraus, Cremona, the Sage number-field implementation, and the cited finite-field sources. Earlier revisions were prepared with Claude assistance. This revision and this description were prepared and posted by Codex under CBirkbeck's direction.


TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -157,8 +157,14 @@ unrelated germ and states nothing — every
 other ingredient is built by Layers 4–7, so modulo that hypothesis the statement is cheap.
 (The *arithmetic* BSD quotient **over `ℚ`**, assuming `Ш` finite, is a marked stretch
 milestone in §Layer 7; *proving* anything about either form is out — as is the
-general-number-field quotient, whose period honestly wants complex uniformisation.) Everything else — through
-Mordell–Weil and Selmer/Sha — is in.
+general-number-field quotient, whose period honestly wants complex uniformisation.) Two further
+statements are out, and §Layer 8 says so where it defines the objects they concern: **Siegel's
+theorem**, that the integral points of a model are finite, which needs Diophantine approximation
+at Thue–Siegel–Roth or Baker strength that nothing here builds; and the **abc conjecture**, an
+inequality bounding `max(|a|,|b|,|c|)` against `rad(abc)` for coprime `a + b = c`, of which the
```
Multramate: The statement of ABC is in FormalConjectures, although maybe it could be here as a definition as well, seeing that the statement of FLT? Once we state Szpiro's conjecture (in terms of conductors of elliptic curves), we can easily prove ABC => Szpiro => FLT and Szpiro => ABC with exponent 3/2. I think this is within scope and relevant to things like `abcQuality`.

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -303,9 +309,16 @@ positivity, and the Frobenius isogeny seeded (Layer 1). The other seeds: the Lay
 so the point group *is* the ideal class group (`toClass_surjective`) — `[n]`-surjectivity for `n`
 invertible in `K` (Layer 1), the `N`-torsion `E[N] ≅ (ℤ/N)²` — stated as an additive
 equivalence (§Conventions) — and the bilinear **Weil pairing** (Layer 2), the finiteness of `E(𝔽_q)` and the
-**Hasse bound** as the integer inequality `a_q² ≤ 4q` (Layer 3), the **quadratic twist** and the
-split-multiplicative-reduction theorem (Layer 5), and the **Mordell–Weil theorem**
-`AddGroup.FG (E K)` (Layer 6). The layers whose central objects are new *types* — the places of
+**Hasse bound** as the integer inequality `a_q² ≤ 4q` together with the **trace of Frobenius**
+and the ordinary/supersingular predicates (Layer 3), the **global and semi-global minimality
+predicates**, the **minimal discriminant ideal**, the **positive defect class** and
```
Multramate: What's the "positive defect class"? Is this a standard term?

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -621,6 +634,43 @@ being distributed as bookkeeping. Its milestones:
   a prerequisite Mathlib lacks, and the seeded Hasse bound's **required companion**: the bound
   counts with `Nat.card`, which reads `0` on an infinite type, so finiteness is what makes the
   count the honest one (any proof of the bound necessarily establishes it).
+- **The Frobenius trace.** `frobeniusTrace E = #F + 1 − #E(F)` for elliptic `E` over a finite
+  field `F`, named and seeded rather than left inline, because three separate strands quantify
+  over it: the Hasse bound below, the trace sequence of the zeta strand, and Layer 1's theorem
+  that `End(E)` of an ordinary curve is an order in `ℚ(π_q)`. ⚠ It carries `[E.IsElliptic]`.
+  The bare point-count defect `#F + 1 − #W(F)` of an arbitrary Weierstrass cubic is a different
+  quantity and does not get this name, since what makes the defect a *trace* is the Layer-1
+  identity `deg(1 − π_q) = #E(𝔽_q)`, which is an elliptic-curve theorem.
```
Multramate: I don't see why you can't define this for non-elliptic curves. You don't have the identity but you can make the definition anyway. Any theorems about it can assume `IsElliptic`.

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -678,6 +728,44 @@ schemes over the valuation ring by nature, and everything below is stated and pr
...
+    comparison `v(Δ) = v(Δ_min) + 12·fᵥ` against any integral model, `fᵥ` the obstruction
+    exponent of §Layer 4.5a. This is the `v(Δ)` Tate's algorithm reads and the local input to the
+    global minimal discriminant ideal.
+  - **Good ordinary and good supersingular reduction.** For `R` with **finite** residue field,
+    `HasGoodOrdinaryReduction` and `HasGoodSupersingularReduction`: good reduction whose
+    reduction is ordinary, respectively supersingular, in the sense of §Layer 3. Thin over their
+    parts, and named because they are the hypothesis the `p`-adic and Iwasawa-theoretic
+    literature states its results under. API: the dichotomy (good reduction is exactly one of
+    the two), compatibility with `frobeniusTrace` of the reduction, and behaviour under
+    unramified base change.
```
Multramate, line 792: I'm fine with all this for now, but note that they may change drastically in mathlib soon, so I can't say.

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -741,6 +829,218 @@ future scheme-facing roadmap.
...
+  `(uᵥ, rᵥ, sᵥ, tᵥ)`. If the positive defect ideal is principal, choose the corresponding
+  global scale `u`; finite approximation then chooses one global `(r, s, t)` satisfying the
+  finitely many required localization congruences, and the transformed equation is integral and
+  minimal everywhere. This is the direct AEC VIII.8.2 route. It shares finite-approximation
+  infrastructure with Kraus's exact-invariant construction, but the two patching theorems are not
+  silently identified.
+
+- **Global-minimality equivalence — number fields only.** For `O = 𝓞 K`, the global construction
+  proves
+  `globalMinimalityClass O E = 1 ↔ E` admits an integral equation that is globally minimal
```
Multramate:
```suggestion
  `globalMinimalityClass O E = 1` if and only if `E` admits an integral equation that is globally minimal
```

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -741,6 +829,218 @@ future scheme-facing roadmap.
...
+     these quotients with the corresponding localization quotients. Apply it with the powers in
+     step 4;
+  6. with the resulting integral `a₁, b₂, a₃`, set
+     `s = a₁/2`, `r = b₂/3 − s²`, and
+     `t = s(b₂ − a₁²)/3 + a₃/2`. Prove that this single `(r, s, t)` transforms
+     `[0, 0, 0, −c₄/48, −c₆/864]` to an integral global equation;
+  7. define `KrausGlobalCondition c₄ c₆ := ∀ v, KrausLocalCondition c₄ c₆ v` and prove
+     `krausGlobalCondition_iff_exists_integralModel` by the construction above; and
+  8. in each application, separately verify local minimality and compute every
+     `obstructionExponentAt`, rather than inferring either property from the discriminant alone.
```
Multramate, line 1040: Genuine question: is Kraus--Laska--Connell within scope? I would imagine we could just "run the algorithm and use its outputs" instead of doing all this contract stuff, but I really don't know. Seeing that Tate's algorithm, which is substantially more complicated, is in scope, maybe this is also in scope. Maybe it's an entirely orthogonal thing so I'm just blabbering nonsense.

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -946,6 +1259,93 @@ scheme-facing roadmap. This layer deliberately does not conflate the two.
...
+consumers**, not every quantity a table of curves over `ℚ` might record. Everything genuinely
+structural has a home above — the Frobenius trace and the ordinary/supersingular dichotomy in
+§Layer 3, potential good reduction and good ordinary/supersingular reduction in §Layer 4, and
+minimal-model invariants and equation construction in §Layers 4.5a–b. It exists so the selected
+vocabulary has one spelling in Tau Ceti, and so quantities which are not unconditionally
+meaningful carry their hypotheses rather than a junk value. Its base is `ℚ` throughout; nothing
+here is claimed over a general number field.
+
+⚠ **It supplies selected adapters, not every field of a database record.** A record also carries
+a Faltings height, a Szpiro ratio, a modular degree, a Manin constant, analytic `Ш` data and
```
Multramate: I can't remember if the conductor is within scope. If yes, the Szpiro ratio is also within scope.

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -946,6 +1259,93 @@ scheme-facing roadmap. This layer deliberately does not conflate the two.
...
+ratio to whatever follows the arithmetic conductor of §Layer 4, since it is a ratio against that
+conductor and not against Layer 4's algorithmic exponent. Naming the boundary is the point: this
+layer is useful without pretending to be exhaustive.
+
+- **The primitive integral short equation.** The prerequisite for the height below, and a
+  milestone in its own right — **seeded**, since the height's whole content is which model it is
+  computed from, and leaving the model's type open would leave the height undefined. The
+  predicate `IsPrimitiveShortModel W` on `W : WeierstrassCurve ℤ` is `W.IsShortNF` together with:
+  no prime `ℓ` has both `ℓ⁴ ∣ a₄` and `ℓ⁶ ∣ a₆`. Bundled as `PrimitiveShortModel E` for elliptic
+  `E/ℚ` — the integral model, its primitivity, and a `VariableChange ℚ` carrying its base change
```
Multramate: Is "primitive" common terminology? It should be called `Is(...)NF` following conventions.

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -946,6 +1259,93 @@ scheme-facing roadmap. This layer deliberately does not conflate the two.
...
+  to `E`. The milestones are `exists_primitiveShortModel` and `primitiveShortModel_unique`:
+  existence of the bundle and uniqueness of its **equation** outright, since the only remaining
+  coefficient freedom is
+  `(a₄, a₆) ↦ (u⁴a₄, u⁶a₆)` with `u = ±1` and `u⁴ = u⁶ = 1`. ⚠ This is **not** §Layer 4.5b's
+  reduced minimal model, which is a *long* equation: they are different canonical models serving
+  different purposes, both stored separately in a database record, and neither substitutes for
+  the other. The bundled change-of-variables witness need not be unique; only the equation and
+  its derived height are canonical.
+- **The height of a curve over `ℚ`.** Two declarations, and the split is the point:
+  `shortEquationHeight W = max (4·|a₄|³) (27·a₆²) : ℕ` for an integral short equation over `ℤ`
```
Multramate, line 1365: Can you just call it `naiveHeight`? This shouldn't clash with other names unless there is a height for long Weierstrass curves. If there is, we should define it for those. What does "seeded" mean?

TauCetiRoadmap/EllipticCurves/README.md
```diff
@@ -946,6 +1259,93 @@ scheme-facing roadmap. This layer deliberately does not conflate the two.
...
+      (W.baseChange ℚ).IsElliptic ∧
+      max (4 * W.a₄.natAbs ^ 3) (27 * W.a₆.natAbs ^ 2) ≤ H}
+  ```
+
+  It is immediate from bounds on the two integers `(A, B)` and yields finiteness of
+  `ℚ`-isomorphism classes of bounded `naiveCurveHeight`. It does **not** assert finiteness of the
+  literal terms `E : WeierstrassCurve ℚ`, which is false because every isomorphism class has
+  infinitely many rational equations. Further API: the relation to
+  `Δ = −16(4A³ + 27B²)` and monotonicity in `|A|` and `|B|`. ⚠ The height does not take the name
+  `naiveHeight`: that is the naïve `x`-height on points of §Layer 6, which Mathlib reserves.
```
Multramate: I would imagine these live in the `WeierstrassCurve.Affine.Point` namespace, so no clashes?

Multramate (commented):
I didn't read the additions to `Suggested.lean`, and I only vaguely read Layer 4.5a/b because I don't really understand them. Maybe Jane will have something to say.

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: thats fine, we can refactor when it changes in mathlib.

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: Doesnt this give you asymptotic FLT? also I think then ABC => Szpiro would likely need to be its own milestone (which is fine)

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: made up AI nonsense, isnt this the Weierstrass class?

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: Won it be wrong for singular things? like youd get 2,0,1 for split/non-split/additive red instead of the usual 1/-1/0.

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: Yes I think conductor is in scope so we can do this as well.

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: yeah again, made up by AI, should maybe be minimal pair or something like that

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: Seeded is just more AI jargon, will remove it.

TauCetiRoadmap/EllipticCurves/README.md
CBirkbeck: *(Claude here, answering this one on Chris's behalf — the rest of the thread replies are his.)*

Largely orthogonal, yes — but deliberately so, and both are wanted.

They are different theorems with different inputs. VIII.8.2 starts from a curve you already hold and shows that triviality of the Weierstrass class is equivalent to a global minimal integral model existing. Kraus starts from a bare pair `(c₄, c₆)` and decides whether they are the *exact* invariants of an integral Weierstrass equation. That is not a curve-construction question — a nonsingular pair already defines a curve over `K` via `a₄ = −c₄/48`, `a₆ = −c₆/864`; the content is integrality with those invariants on the nose.

Worth stating explicitly, because it sharpens the orthogonality: **Kraus is not load-bearing for VIII.8.2 at all.** Over `ℚ` the existence of global minimal models needs neither Kraus nor Tate — Mathlib's `exists_isMinimal` already hands you the local minimalising change of variables at each prime, `Cl(ℤ) = 1` makes the global scale automatic, and finite approximation patches the translations. Kraus is load-bearing for the *other* direction: producing a certified curve from stored invariants. That is the LMFDB-facing consumer, and it is why the acceptance criteria cite records by label (`11.a1@11`, `14.a4@7`, …) — without it we would be assuming a record's equation is the integral one with those invariants rather than certifying that it is.

On "just run the algorithm and use its outputs": an executable implementation would still need a correctness theorem, and the local predicates are meant as that theorem's specification rather than as ceremony around it.

Your instinct is right about one thing, though, and the roadmap should not have implied otherwise: having Tate's algorithm does **not** make Kraus cheap. The difficulty is disjoint — Tate's is the singular-fibre case analysis, Kraus's is solving the `a₁, a₃, b₂` congruences 2-adically and 3-adically. Both find 2 and 3 awkward, for unrelated reasons, and they share no lemmas; the only common infrastructure is the finite-approximation step, which VIII.8.2 needs anyway. So the cost is real, and it is justified by the consumer rather than by being a small increment on Tate.

I have pushed an edit making the distinct input, the intended consumer, and the non-shared cost explicit in Layer 4.5b, so the two routes are not left looking like duplicates.

TauCetiRoadmap/EllipticCurves/README.md
Multramate: Yeah sorry, just asymptotic FLT and not FLT, and ABC => Szpiro seems to be just a page in Silverman?

TauCetiRoadmap/EllipticCurves/README.md
Multramate: Well I don't know. I don't see it defined anywhere. The stuff in Layer 4.5a says this is a number?

TauCetiRoadmap/EllipticCurves/README.md
Multramate: Yeah probably, but having a typeclass as an assumption in a definition is probably less desirable than junk values? I don't know. Seems like the correct thing to do is to distinguish all the points `W.Point` and all the nonsingular points `W.NonsingularPoint`, which I intend to add. That way `a_v = 1 + q_v - #W.Point` is correct?

kim-em (commented):
:robot: Claude+Codex suggest:

Three blocking items and one cleanup.

## 1. The explicit global Kraus construction has the wrong change-of-variables parameters (README lines 996–999)

Step 6 sets `s = a₁/2`, `r = b₂/3 − s²`, `t = s(b₂ − a₁²)/3 + a₃/2`, and transforms `[0, 0, 0, −c₄/48, −c₆/864]`. That source model already carries the target `c₄` and `c₆`, so the change of variables has `u = 1`; and it has `a₁ = a₂ = a₃ = 0`, so its `b₂` is `0`. Mathlib's action is

```lean
(C • W).a₃ = C.u⁻¹ ^ 3 * (W.a₃ + C.r * W.a₁ + 2 * C.t)
(C • W).b₂ = C.u⁻¹ ^ 2 * (W.b₂ + 12 * C.r)
```

so the transform produces

| | transformed `a₃` | transformed `b₂` |
| --- | --- | --- |
| as written | `a₃ + a₁(b₂ − a₁²)/3` | `4b₂ − 3a₁²` |
| intended | `a₃` | `b₂` |

Both coordinates are wrong, not only `b₂`. The correct prescription is

```
s = a₁/2,   r = b₂/12,   t = a₃/2
```

which reproduces `a₁`, `a₃` and `b₂` exactly.

This is easy to miss because the stated formula is correct precisely when `a₂ = 0`: at `b₂ = a₁²` it reduces to `r = a₁²/12`, agreeing with the `(a₁²/12, a₁/2, a₃/2)`-transform named earlier in the same layer.

Correcting `r` also makes explicit an integrality condition that step 4 currently leaves implicit. With `r = b₂/12` the transformed `a₂` is `(b₂ − a₁²)/4`, so the CRT patch has to deliver `b₂ ≡ a₁² (mod 4)` at primes over `2`, alongside the chosen `b₂,ᵥ` residue at primes over `3`. Please state those congruences in step 4, and keep the existing ordering constraint that `a₃` is patched modulo `2` only after recomputing the local witnesses for the fixed global `a₁`.

Scope of the consequence: this blocks the exact-invariant Kraus construction and the database-facing consumers that depend on it, and hence `krausGlobalCondition_iff_exists_integralModel`. The AEC global-minimality route described at lines 955–960 does not depend on it.

## 2. Layer 8 contradicts its own scope

Lines 1299–1305 say a database record carries a Faltings height, a Szpiro ratio and related quantities, "none of which are targets here". Lines 1364 onward then make `szpiroRatio`, the three conjecture statements, their implications, and asymptotic Fermat explicit targets. This is the same layer and the same database-adapter scope in both places, so there is no reading on which both stand.

Given the evident intent, remove "a Szpiro ratio" from the excluded list and adjust "none of which" to match what remains. Deleting the later targets would also restore consistency but would materially shrink the roadmap. As written an implementer cannot tell which statement is authoritative.

## 3. State the new scope intrinsically rather than by current consumers

The new scope is justified repeatedly as adapters "required by the current downstream consumers" (README lines 169, 1290–1291 and 1550, with the same wording in `Suggested.lean`). Which consumers exist is project state, and it dates the document: the repository checklist asks that roadmaps be timeless and say what we want rather than how they came to say it, and the porting guidance asks that each milestone be specified intrinsically.

Please enumerate what this layer owns instead, for example minimal-pair models and height, abc quality, the named Szpiro and abc statements with their implications, and bounded integral-point search, keeping the explicit exclusions. The layer text already contains most of that enumeration, so this is a wording correction rather than a scope change.

## Nonblocking

`Suggested.lean` still says Mathlib "reserves" `naiveHeight` for the point height, while the README now correctly explains that the two declarations simply live in different namespaces. Please align that docstring with the corrected rationale.

CBirkbeck:
*(Claude here, posting on Chris's behalf.)*

All three blocking items, the cleanup, and the two open inline threads are addressed in 94ba980; the branch is also updated from `main` (head `17ad51b`), and CI is green.

## 1. The Kraus change of variables — confirmed, and the error is wider than reported

Confirmed against Mathlib's `variableChange` and checked symbolically. The as-written parameters do not merely break `a₃` and `b₂`: applied to `[0, 0, 0, −c₄/48, −c₆/864]` they miss **four** coordinates.

| coordinate | as written | target |
| --- | --- | --- |
| `a₂` | `4a₂` | `a₂` |
| `a₃` | `a₃ + 4a₁a₂/3` | `a₃` |
| `a₄` | `a₄ + (30a₂² − a₁²a₂)/6` | `a₄` |
| `a₆` | `a₆ + (72a₂³ − 7a₁²a₂² − 6a₁a₂a₃ + 36a₂a₄)/36` | `a₆` |

`(s, r, t) = (a₁/2, b₂/12, a₃/2)` reproduces all five `aᵢ` exactly, so the correction is adopted as proposed. Step 6 now also records why the parameters are forced — the source model has `u = 1` and `a₁ = a₂ = a₃ = 0`, so the action returns `2s`, `2t` and `12r` in the `a₁`, `a₃` and `b₂` coordinates, and `a₂ = 3r − s² = (b₂ − a₁²)/4` then follows for free — and notes that the `(a₁²/12, a₁/2, a₃/2)`-transform of step 2 is the `a₂ = 0` case of this rather than a competing prescription.

Step 4 now states the integrality congruence the corrected `r` makes explicit: the patch must deliver **`b₂ ≡ a₁² (mod 4)`** at the primes above `2`, alongside the chosen `b₂` residue at the primes above `3`. The existing ordering constraint on `a₃` is unchanged.

## 2. Layer 8 scope contradiction

The Szpiro ratio is dropped from the excluded list, so the exclusion paragraph and the targets below no longer contradict each other. The later targets are kept, as intended.

## 3. Scope stated intrinsically

"Required by the current downstream consumers" is gone from all five scope statements (README §Scope, the layer overview, the Layer 8 header, the dependency-order entry, and the `Suggested.lean` section docstring). Each now enumerates what the layer owns: the minimal-pair model and its height, the abc quality, the Szpiro ratio, the named Szpiro and abc statements with the implications between them, and bounded integral-point search — with the explicit exclusions retained. The one surviving mention of consumers is at `Suggested.lean:609`, where it describes which of two comparison lemmas to apply rather than justifying scope.

## Nonblocking — `naiveHeight`

The docstring no longer says Mathlib reserves the name. It now gives the corrected rationale: the curve-level invariant and the point height are different quantities on different types, kept apart by namespace, with neither name reserved against the other.

---

## The `a_v` thread

Both positions were right about different point functors, so it is worth recording the numbers. Counting **all** points of the model, singular point included, gives the classical invariant; counting only the nonsingular locus omits that one rational point and shifts every value by one:

| reduction | `#W(F_q)` all | `1 + q − #all` | `#W(F_q)` nonsingular | `1 + q − #ns` |
| --- | --- | --- | --- | --- |
| split multiplicative | `q` | **`1`** | `q − 1` | `2` |
| nonsplit multiplicative | `q + 2` | **`−1`** | `q + 1` | `0` |
| additive | `q + 1` | **`0`** | `q` | `1` |

(Checked by brute force over `F₅` and `F₇`.) So `a_v = 1 + q_v − #W.Point` is correct exactly when `Point` means all points — the `2, 0, 1` objection is what the nonsingular count gives, which is what `Affine.Point` currently means. The roadmap now says this: the trace is stated for elliptic `E` only because of the current API, and if the `Point` / `NonsingularPoint` split lands upstream the definition should be restated on the full point count for an arbitrary Weierstrass cubic, with `frobeniusTrace` its `[E.IsElliptic]` specialisation. What stays elliptic-specific is not the definition but the theorem making the defect a trace.

## The "defect class" thread

The reason it could not be found: a leftover `positive-defect identity` in Layer 6, the last trace of the old wording, which reads as a claim about a number. That phrase is gone. The class is now named as the declaration `weierstrassDefectClass O W` at the point where it is introduced in Layer 4.5a, and a warning separates the three things the word "defect" was covering — the exponents `fᵥ` are natural numbers, `𝔍_W` is an ideal, and the invariant the global-minimality theorem tests is the class of that ideal.

## abc ⟹ Szpiro being "about a page"

Agreed on the paper argument, and the text now says so. The milestone is kept, with the sizing made explicit: the cost is the valuation bookkeeping against the conductor support, the abc-triple normalisation and the separate handling at `2` and `3`, not the length of the argument on paper. Asymptotic Fermat is already named `asymptoticFermat`, with FLT itself neither stated nor implied.

kim-em (approved):

PR #185, opened 2026-08-10
Add roadmap: Standard Probability Distributions

fzaiser:
This PR adds a roadmap for the **elementary theory of standard probability distributions**. Mathlib already defines several common distributions, but their APIs are uneven: apart from the real Gaussian, many are missing basic results about densities, moments, cdfs, and generating or characteristic functions. There is also no uniform probability measure on an interval, and several families used routinely in probability and statistics have not yet been defined.

The roadmap first fills out the theory of Mathlib's existing distributions, then **adds widely used scalar, multivariate, and matrix-valued families** that Mathlib does not yet define, including Student's t, Dirichlet, and Wishart distributions. The requested results depend on the carrier. Real-valued laws get **densities, cdfs, moments**, exact domains of exponential integrability, and closed forms for **moment-generating and characteristic functions** where useful. Discrete laws remain measures on `ℕ`, with real-valued statements made after casting. **Vector- and matrix-valued laws get means, covariances, marginals, and directional moment-generating functions**; scalar cdfs are only requested for real-valued laws. The measure-level results come first, with random-variable versions stated using `HasLaw`.

### Roadmap structure

The work is arranged in seven layers:

0. **Connect Mathlib's existing continuous distributions** to `HasPDF` and `rnDeriv`, and define the uniform measure on an interval.
1. Prove the missing moments, moment-generating and characteristic functions, convolution identities, and other **elementary results for the existing families**; also develop probability generating functions.
2. Develop the **incomplete gamma and beta functions and the error function** far enough to state the usual closed-form cdfs.
3. Add nine scalar families: **Laplace, log-normal, Weibull, chi-squared, inverse-gamma, Student's t, Fisher's F, negative binomial, and hypergeometric**.
4. Prove the **standard relationships between distributions**, including laws of squares and ratios, the gamma-to-beta construction, mixtures, and basic order statistics.
5. Add **multinomial and Dirichlet distributions**, the **multivariate Gaussian density**, and **conditional multivariate Gaussians** via Schur complements.
6. Build the measure theory and linear algebra needed for **Wishart distributions**, including Lebesgue measure on symmetric matrices, Cholesky decomposition, and the multivariate Gamma integral; then develop the Wishart and inverse-Wishart distributions and Bartlett decomposition.

### Mathlib coordination

Several Mathlib PRs overlap with small parts of Layers 1 and 2: [binomial mean](https://github.com/leanprover-community/mathlib4/pull/40613), [binomial variance](https://github.com/leanprover-community/mathlib4/pull/40916), [exponential moments and memorylessness](https://github.com/leanprover-community/mathlib4/pull/35504), and [`Real.erf`/`erfc`](https://github.com/leanprover-community/mathlib4/pull/34053). The roadmap follows the names and theorem shapes proposed there. The pinned Mathlib version does not yet expose these results, so Tau Ceti supplies them locally. When a later pin does, the local declarations can be replaced by imports.

### Scope and review

Statistical inference is outside the scope of this roadmap. It covers the distributions and their elementary properties, but not likelihoods, estimators, sufficiency, priors, or hypothesis tests. The README also states the boundaries with the [OrthogonalL2Bases roadmap](https://github.com/TauCetiProject/TauCetiRoadmap/blob/main/TauCetiRoadmap/OrthogonalL2Bases/README.md) (moment determinacy), the [OptimalTransport roadmap](https://github.com/TauCetiProject/TauCetiRoadmap/blob/main/TauCetiRoadmap/OptimalTransport/README.md) (Gaussian transport), and the [OneParameterSemigroups roadmap](https://github.com/TauCetiProject/TauCetiRoadmap/blob/main/TauCetiRoadmap/OneParameterSemigroups/README.md) (Bochner's theorem).

A reviewer with a probability or statistics background would be especially helpful. Please pay particular attention to the conventions at degenerate parameter values, the normalization of the Wishart density, the Cholesky and inversion Jacobians, and the branch convention for the Wishart characteristic function.

### AI assistance

🤖 Drafted and revised with AI assistance from Claude Code (Fable 5) and Codex (GPT-5.6 Sol), including several rounds of cross-review.
🧑 The final text was reviewed by me (a human).


cameronfreer:
This seems generally very nice to me! I wonder how well the agents will do in formalizing the various results for which there isn't a pinned reference implementation, but it seems worth a try to see how they could get with a roadmap of this sort (assuming this is allowable for TauCeti).

GPT 5.6 Pro gave some initial advice, and Fabian's codex session pushed back reasonably on several points. Here's the updated version, which he asked me to post.

🤖 GPT 5.6 Pro summary:


> **Requesting changes.** I see five specification-level issues that should be settled before this merges:
>
> 1. The shared API contract does not agree with the family-specific targets, so it is currently ambiguous what completion requires.
> 2. The generic `wishartMeasure` excludes legitimate singular Wishart laws and needs either broader semantics or a narrower name plus a linked singular family.
> 3. The symmetric-matrix section needs to pin a coherent Frobenius topology, measurable structure, and volume normalization; the Cholesky API should then be packaged against those choices.
> 4. Several nonlinear pushforward results depend on change-of-variables machinery that is not explicitly connected to the targets.
> 5. The current-pin status prose belongs in the PR description, not in a timeless roadmap.
>
> I have left focused comments on each point. I am not asking this PR to change its incomplete-Beta boundary convention or other reasonable totalization choices, nor to add a substantially larger support/atomlessness/general-moment programme.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+  Never apply real-only operations directly to `Measure ℕ`.
+  The pgf–mgf bridge is `pgf X P (exp t) = mgf (fun ω => (X ω : ℝ)) P t`.
+- **Finite multivariate families** get a Bochner mean, `covMatrix`/`covarianceBilin`, coordinate marginals, the affine or aggregation laws named below, and any listed characteristic-function formula.
+  They do not get a scalar cdf or `mgf id`.
+  State a directional exponential-moment formula by applying `ProbabilityTheory.mgf` to an explicit real-valued linear functional, and describe its exact domain with `integrableExpSet`.
+  A discrete vector law stays on `ι → ℕ` for masses and support, then uses the coordinatewise cast to `EuclideanSpace ℝ ι` for means, covariances, and characteristic functions.
+- **Symmetric-matrix families** get a density relative to the normalized measure defined in Layer 6, a Bochner mean, the covariance data named below, and any listed characteristic-function formula.
+  As in the multivariate case, directional mgfs use an explicit real-valued linear functional; scalar cdfs and `mgf id` are not targets.
```
cameronfreer: 🤖 GPT 5.6 Pro:

The shared completion contract is inconsistent with the concrete family targets below.

It says that discrete families receive cast-law cdfs, moments, mgfs, cgfs, and characteristic functions, and that finite multivariate families receive a directional exponential-moment formula together with its exact domain. But the later hypergeometric, multinomial, multivariate-Gaussian, and Dirichlet sections do not consistently request these.

Please resolve this in one direction:

* add the missing family-specific targets; or
* weaken the shared contract and state explicitly, for each affected family, which transform/domain theorem is required and which closed form is deliberately not required.

As written, an implementer and reviewer cannot tell whether satisfying the family section completes the roadmap or whether the stronger shared paragraph still imposes additional unstated obligations.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+
+- [mathlib4#40613](https://github.com/leanprover-community/mathlib4/pull/40613) gives the binomial mean;
+- [mathlib4#40916](https://github.com/leanprover-community/mathlib4/pull/40916) gives the binomial variance;
+- [mathlib4#35504](https://github.com/leanprover-community/mathlib4/pull/35504) gives the exponential mgf, moments, and memorylessness; and
+- [mathlib4#34053](https://github.com/leanprover-community/mathlib4/pull/34053) proposes `Real.erf`, `Real.erfc`, `Complex.erf`, and their basic API.
+
+The pinned Mathlib exports none of these declarations.
+Implement the needed results in Tau Ceti now, using the names and theorem shapes from the PRs.
+Mathlib's binomial file contains the proposed mean and variance as `proof_wanted` declarations; these are checked and discarded during elaboration, so matching local theorems do not clash.
+When a later Mathlib pin exports any of these results, delete the local declaration and proof and use Mathlib's version.
```
cameronfreer: 🤖 GPT 5.6 Pro:

This paragraph records the state of one Mathlib pin, so it will become false while the roadmap is still active. The repository requires roadmaps to be timeless.

Please rewrite it conditionally, for example:

> Consume these declarations whenever they are available at Tau Ceti’s Mathlib pin. Otherwise provide local Tau Ceti declarations matching the linked API shape, and remove the local versions once the corresponding imports are available.

The fact that the current pin exports none of them, and any current PR statuses, belongs in this pull request’s description rather than in the roadmap specification.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+Targets:
+
+1. **Squares of Gaussian variables.** Prove `(gaussianReal 0 1).map (· ^ 2) = chiSquaredMeasure 1`.
+   If `X : Fin k → Ω → ℝ` is an i.i.d. standard Gaussian family, prove `HasLaw (fun ω => ∑ i, X i ω ^ 2) (chiSquaredMeasure k) P`.
+   At `k = 0`, the sum is empty and `chiSquaredMeasure 0 = Measure.dirac 0`.
+2. **Ratios.** Prove the following laws for independent variables:
+   - if `0 < ν`, `Z ~ gaussianReal 0 1`, and `V ~ chiSquaredMeasure ν`, then `Z / √(V/ν) ~ studentTMeasure ν`;
+   - if `0 < m`, `0 < n`, `U ~ chiSquaredMeasure m`, and `V ~ chiSquaredMeasure n`, then `(U/m) / (V/n) ~ fisherSnedecorMeasure m n`; and
+   - if `Z₁` and `Z₂` are standard Gaussian, then `Z₁ / Z₂ ~ cauchyMeasure 0 1`.
+3. **Independent gamma variables.** If `0 < a`, `0 < b`, and `0 < r`, and independent variables satisfy `X ~ gammaMeasure a r` and `Y ~ gammaMeasure b r`, prove `X / (X + Y) ~ betaMeasure a b` and independence from `X + Y`.
```
cameronfreer: 🤖 GPT 5.6 Pro:

Several targets here rely on nonlinear change of variables, but the roadmap does not identify the existing Mathlib machinery or pin the intermediate measure equality.

Please add the concrete dependencies

```
Mathlib.MeasureTheory.Function.Jacobian
Mathlib.MeasureTheory.Function.JacobianOneDim
```

and name the direct measure-level theorems intended for use, including map_withDensity_abs_det_fderiv_eq_addHaar where applicable.

In particular, replace the prose-only Beta–Gamma conclusion by an exact product-pushforward target of the form

```
((gammaMeasure a r).prod (gammaMeasure b r)).map
    (fun z => (z.1 / (z.1 + z.2), z.1 + z.2))
  =
(betaMeasure a b).prod (gammaMeasure (a + b) r)
```

under the stated positive-parameter hypotheses, with the zero-denominator branch totalized and proved null. The Beta marginal and independence should be corollaries of that joint-law theorem.

I am not asking for a new generic change-of-variables abstraction here; the roadmap should connect the concrete distribution maps to the existing Mathlib API.

TauCetiRoadmap/StandardDistributions/Suggested.lean
```diff
@@ -0,0 +1,665 @@
...
+    NormedSpace ℝ (SymmetricMatrix p) := sorry
+
+noncomputable instance symmetricMatrixInnerProductSpace (p : ℕ) :
+    InnerProductSpace ℝ (SymmetricMatrix p) := sorry
+
+noncomputable instance symmetricMatrixCompleteSpace (p : ℕ) :
+    CompleteSpace (SymmetricMatrix p) := sorry
+
+noncomputable instance symmetricMatrixBorelSpace (p : ℕ) :
+    BorelSpace (SymmetricMatrix p) := sorry
```
cameronfreer: 🤖 GPT 5.6 Pro:

Please pin the analytic structure here more explicitly.

The intended Frobenius norm/topology is not an automatically canonical inherited instance on this pin: Mathlib deliberately leaves matrix-norm choices noncanonical. So the fix is not simply to remove these instances.

Instead, please specify how the roadmap will:

* select the Frobenius matrix norm and inner product;
* induce coherent normed, inner-product, complete, and measurable structures on the self-adjoint subspace;
* prove that the selected measurable structure is the Borel structure of the selected topology; and
* package `symmetricCoordinates` as a continuous linear equivalence and hence a measurable equivalence.

Also add the comparison between `symmetricLebesgue p`, defined from raw upper-triangular entries, and the volume associated with the Frobenius inner product. The distinction matters to the normalization constants used later in the Wishart density and Jacobians.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+   - the Frobenius inner-product topology and Borel σ-algebra;
+   - measurability of the positive-definite cone; and
+   - the pairing identity `⟪A, Θ⟫_ℝ = trace ((Θ : Matrix _ _ ℝ) * (A : Matrix _ _ ℝ))` for symmetric `A` and `Θ`.
+     This makes `charFun` use the same pairing as the real trace statistic below.
+
+   Define `upperTriangle p = {ij : Fin p × Fin p // ij.1 ≤ ij.2}` and the linear equivalence `symmetricCoordinates : selfAdjoint.submodule ℝ (Matrix (Fin p) (Fin p) ℝ) ≃ₗ[ℝ] (upperTriangle p → ℝ)` by reading the upper-triangular entries.
+   Define `symmetricLebesgue p` as the pushforward of product `volume` on `upperTriangle p → ℝ` along `symmetricCoordinates.symm`, and prove that the coordinate equivalence is measure-preserving.
+
+   This fixes the normalization.
+   Although `symmetricLebesgue p` is an additive Haar measure, it is neither the arbitrarily normalized `Measure.addHaar` nor the volume from a Frobenius-orthonormal basis, whose off-diagonal coordinates have different scaling.
```
cameronfreer: 🤖 GPT 5.6 Pro:

Since the roadmap deliberately chooses a normalization different from Frobenius volume, please make the comparison between the two measures a named target. Otherwise later consumers have two finite-dimensional Lebesgue measures with no bridge, and the normalization of the Wishart constants cannot be checked against Mathlib’s canonical finite-dimensional volume.

TauCetiRoadmap/StandardDistributions/Suggested.lean
```diff
@@ -0,0 +1,665 @@
...
+theorem cholesky_mul_transpose {p : ℕ} (A : PosDefMatrix p) :
+    (cholesky A).1 * ((cholesky A).1)ᵀ =
+      (A.1 : Matrix (Fin p) (Fin p) ℝ) := by sorry
+
+theorem cholesky_unique {p : ℕ} (A : PosDefMatrix p) (L : PosDiagLowerTriangular p)
+    (hL : L.1 * L.1ᵀ = (A.1 : Matrix (Fin p) (Fin p) ℝ)) :
+    L = cholesky A := by sorry
+
+noncomputable def choleskyReconstruction {p : ℕ}
+    (L : PosDiagLowerTriangular p) : PosDefMatrix p := sorry
```
cameronfreer: 🤖 GPT 5.6 Pro:

Please package Cholesky as one reusable equivalence rather than as two functions plus uniqueness.

The target should expose the positive-definite symmetric matrices and positive-diagonal lower-triangular matrices as equivalent, with named inverse identities and continuity/measurability in both directions. The Jacobian result should then identify the exact coordinate derivative determinant used to obtain the measure-pushforward formula.

This avoids leaving downstream users to reconstruct that `cholesky` and `choleskyReconstruction` are inverse maps, and makes the chosen coordinate/topological structures part of one coherent API.

TauCetiRoadmap/StandardDistributions/Suggested.lean
```diff
@@ -0,0 +1,665 @@
...
+    (hS : S.PosDef) (hn : (p : ℝ) - 1 < n) (hΘ : Θ.IsHermitian)
+    (ht : (1 - (2 * t) • (CFC.sqrt S * Θ * CFC.sqrt S)).PosDef) :
+    mgf (fun A : SymmetricMatrix p => Matrix.trace (Θ * (A : Matrix (Fin p) (Fin p) ℝ)))
+        (wishartMeasure n S) t =
+      Real.rpow (Matrix.det (1 - (2 * t) • (Θ * S))) (-n / 2) := by sorry
+
+-- The exponential of a sum of principal logarithms makes the complex branch choice explicit.
+theorem charFun_wishartMeasure {p : ℕ} {n : ℝ} {S : Matrix (Fin p) (Fin p) ℝ}
+    (Θ : SymmetricMatrix p) (hS : S.PosDef) (hn : (p : ℝ) - 1 < n)
+    (hB : (CFC.sqrt S * (Θ : Matrix (Fin p) (Fin p) ℝ) * CFC.sqrt S).IsHermitian) :
```
cameronfreer: 🤖 GPT 5.6 Pro:

`hB` should not be a public hypothesis. For positive-definite `S` and symmetric `Θ`, Hermiticity of

```
CFC.sqrt S * Θ * CFC.sqrt S
```

is a consequence of the existing assumptions. Construct it internally and remove this proof obligation from callers.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+   At `p = 0`, prove the same identity for every `a`; both sides are `1`, so no convergence hypothesis is needed.
+   The occurrence of `symmetricLebesgue p` in this theorem is essential: its normalization is the one used by the Wishart density, and the Cholesky theorem computes the integral in independent lower-triangular coordinates.
+4. **Wishart distribution.** Define `wishartMeasure (n : ℝ) (S : Matrix (Fin p) (Fin p) ℝ)` on the symmetric-matrix subspace from item 1.
+   When `hS : S.PosDef`, write `Sₛ` for `S` bundled into that subspace using `hS.isHermitian` and `Matrix.isHermitian_iff_isSelfAdjoint`.
+   Expressions such as `det A`, `A⁻¹`, and `trace A` refer to the underlying matrix of a bundled value.
+
+   When `S.PosDef` and `(p : ℝ) - 1 < n`, define the law relative to `symmetricLebesgue p`.
+   On the positive-definite cone its density is `Real.rpow (det A) ((n - (p : ℝ) - 1) / 2) * exp (-trace (S⁻¹ * A) / 2) / (Real.rpow 2 (n * (p : ℝ) / 2) * Real.rpow (det S) (n / 2) * multivariateGamma p (n / 2))`; outside the cone it is zero.
+   Define the measure to be zero when the parameter hypotheses fail.
```
cameronfreer: 🤖 GPT 5.6 Pro:

This definition gives only the full-dimensional density branch, but its generic name suggests the whole Wishart family.

For positive-definite scale, Gaussian Gram sums with natural degrees of freedom below the density threshold are still probability laws; they are singular, not zero. The roadmap itself recognizes this when it says the Gaussian sum is singular for `ν < p`, but then excludes that law from `wishartMeasure`.

Please choose one architecture explicitly:

1. extend `wishartMeasure` with the singular natural-degree branch and prove its agreement with the density construction when `p ≤ ν`; or
2. rename the present object to something like `nonsingularWishartMeasure` and add a linked Gaussian-sum family with:
  * probability for every natural `ν`,
  * rank/support control,
  * arbitrary linear congruence,
  * and equality with the nonsingular density law on the overlap.

If singular scale matrices are also intended, pin the rank-dependent existence domain. Otherwise state that this roadmap’s real-parameter density family fixes `S.PosDef`, while the Gaussian-sum family is the only singular extension requested.

kim-em (commented):
🤖 Claude says:

I read this against the **Writing a roadmap** checklist and checked the mathematics family by family: densities, cdfs, means, variances, moments, the exact `integrableExpSet` domains, the multivariate Gamma integral, the Cholesky and inversion Jacobians, the Wishart density and its mgf and characteristic function, and the Bartlett degrees of freedom. I could not falsify any explicit formula. The zero-based `Fin p` conventions in the Cholesky Jacobian and the Bartlett diagonal are right; the Layer 6 constants are the correct ones for the raw upper-triangle coordinate convention you chose; the eigenvalue-by-eigenvalue principal log genuinely does avoid the determinant branch ambiguity; and the `p = 0` corner of the Wishart convolution hypothesis is a real subtlety that the roadmap gets correct. Everything below is about specification rather than arithmetic.

One process note that is easy to miss: **CI has never run on this branch.** The `CI` workflow is sitting at `action_required` (first-time-contributor approval), so the `build` check has never elaborated `Suggested.lean`. I built it locally against the pinned toolchain (`v4.34.0-rc1`) and Mathlib `de5ce8a`: it elaborates with `sorry` warnings only, no errors. Somebody with write access should approve the workflow so that stays checked automatically.

Comments inline, most important first. The first overlaps with a point GPT 5.6 Pro raised in @cameronfreer's review; I have restated it with the specific instances, since agreeing without evidence would not help you act on it.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+- **Real-valued families (`Measure ℝ`)** get a `ProbabilityTheory.cdf`: the named closed form when one is listed below, and otherwise the integral of the pdf.
+  They also get the stated means, variances, raw or central moments, and an exact description of `integrableExpSet id μ`.
+  Prove a closed form for `mgf id` or `MeasureTheory.charFun` only when one is named below.
+  Heavy-tailed families need both the sharp existence hypotheses and the matching non-integrability results.
+  Whenever an mgf formula is requested on its finiteness domain, also give the corresponding `cgf` as the real logarithm of that formula.
+
+  The absence of an mgf or characteristic-function formula is deliberate.
+  In particular, this roadmap does not ask for the confluent-hypergeometric Beta transforms, the Bessel-function inverse-gamma transforms, or a log-normal mgf formula.
+  The exact exponential-integrability domain is still required.
+- **Discrete families on `ℕ`** stay on their native carrier for masses, support, convolution, and `pgf`.
```
kim-em, line 59: 🤖 Claude says:

**The shared contract and the family sections disagree, so "complete" is not decidable.**

This section imposes obligations that the per-family targets then do not list, and there is no way to tell which document wins:

* This bullet requires "real means, variances, moments, cdfs, mgfs, cgfs, and characteristic functions for the pushforward". Hypergeometric (Layer 3) gets only a mean and a variance. Geometric (Layer 1) and negative binomial (Layer 3) get no cdf. For the hypergeometric transforms it is not even clear what is being asked: the finite-support restatement as a sum over the mass function is available but carries nothing beyond the mass formula, while the classical closed form is a Gauss hypergeometric. Say which one counts.
* The cdf fallback in the real-valued bullet ("otherwise the integral of the pdf") cannot discharge a discrete law, which has no pdf, so those three cdfs are required with no stated way to satisfy them.
* The finite-multivariate bullet below requires a directional exponential-moment formula "and ... its exact domain with `integrableExpSet`", plus `covMatrix`/`covarianceBilin`. None of the multivariate Gaussian, multinomial, or Dirichlet sections in Layer 5 list one, and Dirichlet gives entrywise variances and covariances without ever connecting them to `covMatrix`.
* The symmetric-matrix bullet is phrased as a convention ("directional mgfs use an explicit real-valued linear functional") rather than a requirement, and inverse-Wishart accordingly gets no covariance, characteristic function, or directional mgf. Whether that is deliberate is unguessable.
* Laplace never gets a mean or a variance anywhere in this README. `Suggested.lean` has `variance_id_laplaceMeasure`, but that file is explicitly not the specification, so as things stand the Laplace mean and variance are not targets.

You already have the right pattern for this: the paragraph beginning "The absence of an mgf or characteristic-function formula is deliberate". Extend it into a per-family statement of which transforms are wanted and which are deliberately not. Right now someone who satisfies a family section cannot tell whether this paragraph still has claims on them.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+- **Densities, cdfs, and laws.** Use the `StieltjesFunction`-valued `ProbabilityTheory.cdf` together with `measure_cdf` and `Measure.eq_of_cdf`; `MeasureTheory.pdf`/`HasPDF` and `HasPDF.hasLaw`; `Measure.withDensity`; and `HasLaw`/`HasCondDistrib`.
+- **Convolution, mixtures, and independence.** Use `Measure.conv`, `MeasureTheory.Measure.conv_assoc` and its surrounding API, and `Measure.bind` with kernel composition.
+  Reuse `iIndepFun`, `IndepFun`, `IdentDistrib`, `variance_sum`, and `IndepFun.variance_sum`.
+- **Special functions.** Use `Real.Gamma`, `Complex.Gamma`, `Complex.betaIntegral`, `Real.Gamma_add_one`, the log-convexity and Bohr–Mollerup results, and `integral_gaussian`.
+- **Matrix and finite-dimensional analysis.** Use `Matrix.PosDef`/`PosSemidef` and their spectral theory, including `Matrix.IsHermitian.spectral_theorem`, `eigenvalues_pos`, and `Matrix.PosDef.det_pos`.
+  Use the continuous functional calculus on Hermitian matrices, especially `CFC.sqrt`; the Schur-complement file `Mathlib/LinearAlgebra/Matrix/SchurComplement.lean`; and `condDistrib` with `condDistrib_ae_eq_of_measure_eq_compProd`.
```
kim-em: 🤖 Claude says:

**Nothing in this list anchors the change of variables, which a large part of the roadmap rests on.**

The log-normal density from `map exp`, the inverse-gamma density from `map (·⁻¹)`, the Dirichlet chart density, the Layer 4 ratio laws, the Cholesky Jacobian, `symmetricCongruence`, and the matrix-inversion Jacobian all need it, and none of them says what it is built on. Please name `Mathlib/MeasureTheory/Function/Jacobian.lean` and `Mathlib/MeasureTheory/Function/JacobianOneDim.lean`, and the specific results intended for use: `map_withDensity_abs_det_fderiv_eq_addHaar`, `restrict_map_withDensity_abs_det_fderiv_eq_addHaar`, `lintegral_image_eq_lintegral_abs_det_fderiv_mul`, `integral_image_eq_integral_abs_det_fderiv_smul`.

One consequence has to become a target rather than stay implicit: those results all assume `[IsAddHaarMeasure μ] [BorelSpace E] [FiniteDimensional ℝ E]`. Layer 6 says in passing that `symmetricLebesgue p` "is an additive Haar measure" but never makes `(symmetricLebesgue p).IsAddHaarMeasure` a named target, so as things stand none of the Layer 6 Jacobian results can invoke the API they need.

Once this list names the machinery, GPT's request on Layer 4 item 3 for the joint Gamma-Beta pushforward is the right shape and I would take it: independence from `X + Y` is already a required conclusion there, and the single joint theorem supplies both it and the Beta marginal.

TauCetiRoadmap/StandardDistributions/Suggested.lean
```diff
@@ -0,0 +1,665 @@
...
+
+/-! ### Cholesky coordinates -/
+
+abbrev PosDefMatrix (p : ℕ) :=
+  {A : SymmetricMatrix p // (A : Matrix (Fin p) (Fin p) ℝ).PosDef}
+
+theorem measurableSet_posDefMatrix (p : ℕ) :
+    MeasurableSet {A : SymmetricMatrix p | (A : Matrix (Fin p) (Fin p) ℝ).PosDef} := by sorry
+
+abbrev PosDiagLowerTriangular (p : ℕ) :=
```
kim-em, line 596: 🤖 Claude says:

**This respells a Mathlib notion.**

`∀ i j : Fin p, i < j → L i j = 0` is exactly `Matrix.IsLowerTriangular L` (`Mathlib/LinearAlgebra/Matrix/Block.lean`, `abbrev IsLowerTriangular M := M.BlockTriangular toDual`), and Tau Ceti already uses the dual spelling in `TauCeti/LinearAlgebra/Matrix/Triangular.lean`. The "use Mathlib's vocabulary" rule is aimed squarely at this: a private restatement drifts from the library and grows a redundant theory of lemmas Mathlib already proves. The README's prose "the subtype of lower-triangular matrices with positive diagonal" should name the predicate too.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+   This fixes the normalization.
+   Although `symmetricLebesgue p` is an additive Haar measure, it is neither the arbitrarily normalized `Measure.addHaar` nor the volume from a Frobenius-orthonormal basis, whose off-diagonal coordinates have different scaling.
+   Prove explicitly that `symmetricLebesgue 0` is the Dirac measure on the unique zero-dimensional symmetric matrix.
+
+   For `C : Matrix.GeneralLinearGroup (Fin p) ℝ`, define `symmetricCongruence C` by `A ↦ C * A * Cᵀ`.
+   In `symmetricCoordinates`, prove that its determinant is `(Matrix.det (C : Matrix _ _ ℝ)) ^ (p + 1)`.
+   Equivalently, prove `(symmetricLebesgue p).map (symmetricCongruence C) = ((ENNReal.ofReal |Matrix.det (C : Matrix _ _ ℝ)|) ^ (p + 1))⁻¹ • symmetricLebesgue p`.
+   This change of variables supplies the general-scale Wishart formulas.
+   The next item gives the separate positive-cone change of variables.
+2. **Cholesky decomposition.** Define the Cholesky map on the subtype of positive-definite symmetric matrices, with values in the subtype of lower-triangular matrices with positive diagonal.
```
kim-em: 🤖 Claude says:

**Cholesky should be built on Mathlib's `LDL`, and needs the inverse direction in its API.**

Mathlib already has `Mathlib/Analysis/Matrix/LDL.lean`, with `LDL.lower`, `LDL.diag`, and `LDL.lower_conj_diag : LDL.lower hS * LDL.diag hS * (LDL.lower hS)ᴴ = S` for `hS : S.PosDef`; Cholesky is `LDL.lower * LDL.diag ^ (1/2)`. "Defer to Mathlib" applies: cite it and say whether `cholesky` is defined from it. That file's own TODO, proving `LDL.lower` is lower triangular, is a short step from `LDL.lowerInv_triangular` and `blockTriangular_inv_of_blockTriangular`, but it is a step the roadmap currently assumes rather than asks for, so it should be an explicit target.

Separately: nothing in the proposed API says `cholesky` and `choleskyReconstruction` are mutually inverse. Ask for the named inverse identities in both directions, or for a measurable equivalence between the positive-definite symmetric matrices and the positive-diagonal lower-triangular matrices, so downstream users do not each have to rederive it from the uniqueness lemma.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+```
+
+Completion checks:
+
+- Under the stated positive parameters, the `1 × 1` Wishart law agrees with Layer 3's chi-squared law through the explicit coordinate equivalence.
+- `multivariateGamma 1 a = Gamma a`.
+
+## Boundaries with other roadmaps
+
+- The [orthogonal-L²-bases roadmap](../OrthogonalL2Bases/README.md) covers Gaussian Hermite L² theory and moment determinacy under `TauCeti/Probability/Moments/`.
```
kim-em: 🤖 Claude says:

**Existing Tau Ceti material is not cited, and this sentence mislocates the other roadmap.**

The repository asks prototypes to be written against existing Tau Ceti implementations rather than restating them, and `Suggested.lean` imports only Mathlib. Relevant material that already exists:

* `TauCeti/Probability/Distributions/Gaussian/Pi.lean`, whose `pi_gaussianReal_eq_withDensity` is the isotropic case of Layer 5 item 2. Say how it feeds the general density theorem.
* `TauCeti/MeasureTheory/Measure/GiryMonad.lean` (`map_bind`, `bind_map`), for the Layer 4 mixture and the parameter-measurability targets.
* `TauCeti/LinearAlgebra/Matrix/Triangular.lean` and `TauCeti/LinearAlgebra/Matrix/PosDef.lean`, for Layer 6.

Minor, on this line: the orthogonal-L² roadmap's Gaussian material is not confined to `TauCeti/Probability/Moments/`. It also sits in `TauCeti/Probability/Distributions/Gaussian/` (`Basic.lean`, `Pi.lean`, `PolynomialMemLp.lean`, `Hermite/`), which is where Layer 5 wants to put `Gaussian/Multivariate.lean`. There is no contradiction, but the sentence describes that roadmap's footprint inaccurately.

TauCetiRoadmap/StandardDistributions/README.md
```diff
@@ -0,0 +1,802 @@
...
+## Coordination with existing work
+
+Several Mathlib PRs propose declarations needed here:
+
+- [mathlib4#40613](https://github.com/leanprover-community/mathlib4/pull/40613) gives the binomial mean;
+- [mathlib4#40916](https://github.com/leanprover-community/mathlib4/pull/40916) gives the binomial variance;
+- [mathlib4#35504](https://github.com/leanprover-community/mathlib4/pull/35504) gives the exponential mgf, moments, and memorylessness; and
+- [mathlib4#34053](https://github.com/leanprover-community/mathlib4/pull/34053) proposes `Real.erf`, `Real.erfc`, `Complex.erf`, and their basic API.
+
+The pinned Mathlib exports none of these declarations.
```
kim-em: 🤖 Claude says:

**Dated prose: roadmaps are timeless.**

Three places, all cheap to fix:

* This sentence, plus the following one about Mathlib's binomial file containing `proof_wanted` declarations that "are checked and discarded during elaboration". Both will be false while the roadmap is still live, and the second is elaboration trivia rather than specification.
* Layer 2's "The pinned Mathlib has the complete Gamma and Beta functions but not their incomplete versions", and "implement that shape locally on the current pin, and use Mathlib's version on any later pin that provides it".
* The intro's "No named distribution currently has a corresponding `MeasureTheory.HasPDF` theorem".

The conditional form is already most of what you wrote: consume these declarations whenever they exist at Tau Ceti's Mathlib pin, otherwise provide local declarations matching the linked API shape and delete them once the import is available. The current status of each Mathlib PR belongs in this pull request's description, not in the roadmap.

TauCetiRoadmap/StandardDistributions/Suggested.lean
```diff
@@ -0,0 +1,665 @@
...
+
+theorem mem_integrableExpSet_trace_mul_wishartMeasure_iff {p : ℕ} {n t : ℝ}
+    {S Θ : Matrix (Fin p) (Fin p) ℝ} (hS : S.PosDef) (hn : (p : ℝ) - 1 < n)
+    (hΘ : Θ.IsHermitian) :
+    t ∈ integrableExpSet
+        (fun A : SymmetricMatrix p => Matrix.trace (Θ * (A : Matrix (Fin p) (Fin p) ℝ)))
+        (wishartMeasure n S) ↔
+      (1 - (2 * t) • (CFC.sqrt S * Θ * CFC.sqrt S)).PosDef := by sorry
+
+theorem mgf_trace_mul_wishartMeasure {p : ℕ} {n t : ℝ} {S Θ : Matrix (Fin p) (Fin p) ℝ}
```
kim-em: 🤖 Claude says:

**`Θ` is typed two different ways for the same statistic.**

Here and in `mem_integrableExpSet_trace_mul_wishartMeasure_iff`, `Θ` is a raw `Matrix (Fin p) (Fin p) ℝ` with `hΘ : Θ.IsHermitian`; in `charFun_wishartMeasure` below it is a bundled `SymmetricMatrix p`. Your own convention says symmetric-matrix laws live on the bundled subspace. Pick one spelling so the mgf and the characteristic function of the same linear statistic are stated about the same object.

Agreeing with GPT on `hB` below, with a concrete route: it follows from `hS : S.PosDef` and the symmetry of `Θ`, since `CFC.sqrt S` is symmetric and hence so is the triple product. Because the statement names eigenvalues through `hB`, removing it needs a named lemma to supply the term (something like `Matrix.PosDef.isHermitian_sqrt_conj`); make that a target and construct `hB` inside the statement.

TauCetiRoadmap/StandardDistributions/Suggested.lean
```diff
@@ -0,0 +1,665 @@
...
+abbrev upperTriangle (p : ℕ) := {ij : Fin p × Fin p // ij.1 ≤ ij.2}
+
+noncomputable def symmetricCoordinates (p : ℕ) :
+    SymmetricMatrix p ≃ₗ[ℝ] (upperTriangle p → ℝ) := sorry
+
+noncomputable instance symmetricMatrixMeasurableSpace (p : ℕ) :
+    MeasurableSpace (SymmetricMatrix p) :=
+  MeasurableSpace.comap (symmetricCoordinates p) inferInstance
+
+noncomputable instance symmetricMatrixNormedAddCommGroup (p : ℕ) :
```
kim-em, line 477: 🤖 Claude says:

**Two of these six instances are redundant, and the roadmap has not decided what the other four are built from.**

`ContinuousENorm` follows from the `NormedAddCommGroup` two lines above via the additive form of `SeminormedGroup.toContinuousENorm`, and `NormedSpace` follows from the `InnerProductSpace` three lines below, since `InnerProductSpace` extends `NormedSpace`. Both should go.

For the rest: Mathlib deliberately installs no `NormedAddCommGroup`, `InnerProductSpace`, or `MeasurableSpace` instance on `Matrix`, so nothing can be inherited until Tau Ceti picks the ambient structure, and "pin conventions" wants that decision written down. The concrete request is short: construct the Frobenius structure locally from `Matrix.toMatrixInnerProductSpace` at `M = 1` (`Mathlib/Analysis/Matrix/Order.lean`, giving `⟪x, y⟫ = (y * M * xᴴ).trace`, which reproduces your pairing identity exactly), inherit it on the submodule through `Submodule.normedAddCommGroup` and `Submodule.innerProductSpace`, and prove that the coordinate σ-algebra is the Borel σ-algebra of that topology, packaging `symmetricCoordinates` as a continuous linear equivalence.

Two smaller things while you are in this area:

* The README says `symmetricLebesgue p` is "neither the arbitrarily normalized `Measure.addHaar` nor the volume from a Frobenius-orthonormal basis", and then never relates them. State the comparison with its direction: `volume_Frobenius = 2 ^ (p * (p - 1) / 4) • symmetricLebesgue p`.
* `regularizedIncompleteBeta` normalises by `Real.Gamma a * Real.Gamma b / Real.Gamma (a + b)`, which is exactly `ProbabilityTheory.beta a b` (`Mathlib/Probability/Distributions/Beta.lean`) and is what `betaPDFReal` already normalises by; the DLMF 8.17.20 recurrence then reads `1 / (a * beta a b)`. Use it, and add `beta` to **Mathlib foundations to reuse**, which currently cites `Complex.betaIntegral` but not this. Separately, Layer 4's gamma-mixed Poisson law needs `Measurable fun r => poissonMeasure r`, and points at Layer 0 for it, but Layer 0's derivation covers `withDensity` families only and its key declarations name just `measurable_gammaMeasure`; name the Poisson case.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c. The roadmap now says to consume declarations when the Tau Ceti pin provides them, otherwise implement the linked API shapes locally and remove those versions once the imports provide them. Pin-specific status and `proof_wanted` prose are gone.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Mostly addressed in b911c0c: the roadmap now cites and the prototype imports Gaussian/Pi, GiryMonad, and Triangular, explains how the Gaussian product density feeds Layer 5, and corrects the orthogonal-L² roadmap boundary. I did not import TauCeti/LinearAlgebra/Matrix/PosDef.lean: at the merged Tau Ceti pin that file is specifically about integer-to-rational positive definiteness and rational `Bᴴ * B`, so none of its declarations applies to this real-matrix development. I am leaving this thread open for reviewer confirmation on that point.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Addressed in b911c0c. All Wishart transform statements now take bundled symmetric `Θ`, and `isHermitian_wishartSandwich` is a named target used internally by the characteristic-function statement, eliminating both `hΘ` and caller-supplied `hB` proof arguments.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c. The foundations section names both Jacobian modules and all four requested measure/integral theorems, requires concrete source/target regions and derivative equalities, and Layer 6 now has the named `MeasureTheory.Measure.IsAddHaarMeasure (symmetricLebesgue p)` instance. The Gamma–Beta joint pushforward is also an exact target.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Addressed in b911c0c. The redundant `ContinuousENorm` and `NormedSpace` declarations are gone; the remaining structures are built coherently from the `M = 1` Frobenius construction and Borel topology. The volume comparison is named with the requested direction, incomplete Beta uses `ProbabilityTheory.beta` and the DLMF recurrence normalization, and Layer 0 now names `measurable_poissonMeasure`.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Addressed in b911c0c. `PosDiagLowerTriangular` now uses `L.IsLowerTriangular`, the README names `Matrix.IsLowerTriangular`, and the prototype imports Tau Ceti’s existing triangular-matrix API.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c. The roadmap requires construction from `LDL.lower` and `LDL.diag`, makes lower triangularity of `LDL.lower` an explicit prerequisite via the two named lemmas, and packages both directions as an equivalence with inverse, continuity, and measurability APIs.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c with `volume_symmetricMatrix_eq_smul_symmetricLebesgue`: Frobenius volume is `2^(p(p-1)/4)` times the raw upper-coordinate normalization, using a real exponent and `ENNReal.ofReal` in the Lean statement.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c. The contract is now explicitly exhaustive by family. In particular, the roadmap adds geometric and negative-binomial cumulative formulas, hypergeometric finite-sum cdf/mgf/cgf/characteristic-function targets, the multivariate directional-domain targets and covariance links, the deliberate Dirichlet/inverse-Wishart transform boundaries, and Laplace mean and variance.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c. The shared carrier bullets now specify representation rather than hidden obligations, and the family entries are explicitly exhaustive. I added the missing hypergeometric finite-sum formulas, Gaussian and multinomial directional mgfs/domains, Dirichlet covariance/domain with an explicit no-closed-form boundary, and the inverse-Wishart exemption.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Addressed in b911c0c. The prototype selects the `M = 1` Frobenius structure with `Matrix.toMatrixNormedAddCommGroup`/`toMatrixInnerProductSpace`, inherits it on the self-adjoint submodule, fixes the matching topology and Borel structure, and packages `symmetricCoordinates` as continuous-linear and measurable equivalences. The roadmap and prototype also name the comparison with Frobenius volume.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Addressed in b911c0c. Cholesky is now specified by `choleskyEquiv`, with named inverse identities, continuity and measurability in both directions, plus `choleskyHomeomorph` and `choleskyMeasurableEquiv`. The coordinate derivative determinant is a separate named target feeding the pushforward theorem.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c. The roadmap names both Jacobian modules and the concrete Mathlib theorems to use. Layer 4 now requires the exact Gamma–Beta joint product-pushforward equality, treats the totalized zero-denominator branch as source-null, and makes the Beta marginal and independence corollaries. Suggested.lean contains the corresponding joint-law prototype.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c using architecture 2. The real-degree density object is now `nonsingularWishartMeasure`; `wishartGramMeasure` covers every natural degree and positive-semidefinite scale, with probability, PSD support, rank control, arbitrary rectangular congruence, equality on the nonsingular overlap, and singularity below the rank threshold.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Addressed in b911c0c. The intro no longer makes a time-sensitive global `HasPDF` claim, the coordination section is conditional on whatever the Tau Ceti pin provides, and Layer 2 uses the same conditional consume-or-implement wording. Pin status and elaboration trivia were removed.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Addressed in b911c0c. `Θ` is bundled as a symmetric matrix, `isHermitian_wishartSandwich` constructs Hermiticity from `hS` and that bundled symmetry, and the characteristic-function theorem uses it internally with no public `hB` argument.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Follow-up/correction in `0f0fb3a`: my previous reply was too strong. The `Matrix.toMatrixInnerProductSpace` construction used in `b911c0c` did not produce a coherent final topology: Matrix already has its product topology and uniformity, while that construction induced a second, non-definitionally-equal topology on the self-adjoint subtype.

The final specification now retains the existing product/subtype topology and uniformity, installs the Frobenius norm through `Matrix.frobeniusNormedAddCommGroup` and `Matrix.frobeniusNormedSpace`, and asks for a compatible `Matrix.frobeniusInnerProductSpace`. It explicitly supplies the uniform-add-group, second-countability, completeness, continuous-enorm, Borel, and inner-product volume structures and includes completion checks that the selected topology and uniformity are definitionally the subtype instances and agree with one another.

The raw-coordinate definition of `symmetricLebesgue` and its named comparison with Frobenius volume remain unchanged. I am reopening this thread because the final construction is materially different from the one described in my earlier reply, and I would appreciate confirmation of the corrected direction.

TauCetiRoadmap/StandardDistributions/Suggested.lean
fzaiser: 🤖 Follow-up/correction in `0f0fb3a`: the non-topological parts of my previous reply remain addressed—`ProbabilityTheory.beta` is used, the Frobenius-volume comparison remains explicit, and `measurable_poissonMeasure` remains a named target—but the topology portion of that reply was not correct.

The roadmap no longer uses `Matrix.toMatrixInnerProductSpace` or installs a replacement topology. It keeps Matrix’s product topology and uniformity and the inherited subtype structures, installs the Frobenius norm through `Matrix.frobeniusNormedAddCommGroup`/`Matrix.frobeniusNormedSpace`, and adds a compatible `Matrix.frobeniusInnerProductSpace`. The explicit `NormedSpace`, `ContinuousENorm`, `IsUniformAddGroup`, `SecondCountableTopology`, `CompleteSpace`, `MeasurableSpace`, `BorelSpace`, and `MeasureSpace` declarations are intentional bridging instances for that coherent structure; the prototype checks definitionally that the topology and uniformity remain the subtype ones.

I am reopening this thread because my earlier statement that the superseded construction was coherent, and that the relevant bridging declarations could simply be omitted, is no longer accurate.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Follow-up in `0f0fb3a`: a later completeness pass found three residual omissions despite my earlier statement that the family contract was exhaustive. The Weibull entry listed general moments but not the variance, positive-degree chi-squared lacked its characteristic-function target, and the real-valued cgf rule had not been carried over explicitly to symmetric-matrix trace mgfs.

Those are now fixed: the roadmap and prototype name `variance_id_weibullMeasure` and `charFun_chiSquaredMeasure`, and both Wishart trace-mgf families have matching cgf targets. The family entries remain the exhaustive source of obligations; no further transforms are implicitly required by the carrier-level paragraphs.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Follow-up in `0f0fb3a`: a later completeness pass found three residual omissions despite my earlier statement that the family contract was exhaustive. The Weibull entry listed general moments but not the variance, positive-degree chi-squared lacked its characteristic-function target, and the real-valued cgf rule had not been carried over explicitly to symmetric-matrix trace mgfs.

Those are now fixed: the roadmap and prototype name `variance_id_weibullMeasure` and `charFun_chiSquaredMeasure`, and both Wishart trace-mgf families have matching cgf targets. The family entries remain the exhaustive source of obligations; no further transforms are implicitly required by the carrier-level paragraphs.

TauCetiRoadmap/StandardDistributions/README.md
fzaiser: 🤖 Follow-up in `0f0fb3a`: the two-family architecture from my previous reply remains, but the Gaussian-Gram family’s totalization and API have been strengthened.

`wishartGramMeasure ν S` is now the direct Gaussian-Gram pushforward for every matrix `S`, without a separate branch. It therefore inherits Mathlib’s multivariate-Gaussian behavior: it is a probability measure for every `ν` and `S`, and is Dirac at zero when `S` is not positive semidefinite. The support, rank, congruence, and transform theorems carry `S.PosSemidef` where that hypothesis is mathematically needed.

The follow-up also adds the zero-degree Dirac theorem, convolution for all natural degrees, the exact trace-mgf domain, mgf, cgf, characteristic function, the overlap equality via characteristic-function uniqueness, its restored Gaussian-sum `HasLaw` corollary, and singularity under the sharp condition `min ν S.rank < p`. This supersedes the narrower “positive-semidefinite scale” wording in my earlier reply.

fzaiser:
🤖 Review-response summary

Thank you for the detailed reviews. I addressed this round across `b911c0c` and `0f0fb3a`.

- **Completion contract and API coverage.** The family entries are now explicitly exhaustive; the carrier-level rules explain how to state those targets rather than imposing hidden additional obligations. The missing cumulative-mass, finite-sum transform, multivariate covariance, and directional-domain targets were added, together with explicit boundaries where no closed form is requested. A follow-up completeness pass also added the Weibull variance, the positive-degree chi-squared characteristic function, and the rule that every requested symmetric-matrix trace mgf receives the corresponding cgf.

- **Timelessness and reuse.** Pin-specific status prose was replaced by conditional “consume when available, otherwise implement the linked shape locally” language. The roadmap now cites and prototypes against the relevant existing Tau Ceti Gaussian, Giry-monad, and triangular-matrix material. It also uses `ProbabilityTheory.beta`, names Poisson parameter measurability, and measures matrix parameters through their coordinates because Mathlib gives `Matrix` no measurable-space instance.

- **Change of variables.** The roadmap now names Mathlib’s Jacobian modules and the concrete measure/integral theorems intended for use. It requires explicit source and target regions, injectivity, derivative determinants, and resulting measure equalities for each nonlinear distribution map. The Gamma–Beta construction is now an exact joint product-pushforward theorem, from which the marginal law and independence follow. `symmetricLebesgue` also has the named additive-Haar instance needed by this machinery.

- **Symmetric-matrix topology and volume.** The final version keeps Matrix’s existing product topology and uniformity and the corresponding subtype structures on the self-adjoint submodule. It installs the Frobenius norm through `Matrix.frobeniusNormedAddCommGroup` and `Matrix.frobeniusNormedSpace`, asks for a compatible `Matrix.frobeniusInnerProductSpace`, and supplies the required uniform, completeness, second-countability, Borel, continuous-enorm, and volume instances coherently. It explicitly excludes `Matrix.toMatrixInnerProductSpace`, whose induced topology is not definitionally the product topology, and includes definitional-equality completion checks. The positive-diagonal lower-triangular carrier likewise retains its subtype topology, with its coordinate map proved to be a homeomorphism rather than used to install a second topology.

- **Cholesky API.** Cholesky is specified on top of Mathlib’s LDL decomposition and `Matrix.IsLowerTriangular`. It is packaged as an equivalence, homeomorphism, and measurable equivalence with named inverse identities. The coordinate derivative determinant and resulting measure-pushforward theorem are explicit targets.

- **Wishart architecture and transforms.** The roadmap keeps a real-degree density family, `nonsingularWishartMeasure`, linked to the natural-degree Gaussian-Gram family, `wishartGramMeasure`. The Gram family is now the direct pushforward of Mathlib’s multivariate Gaussian for every scale matrix, so it is always a probability measure and becomes Dirac at zero when the scale is not positive semidefinite, following Mathlib’s totalization. Under positive-semidefinite scale it has PSD support, the sharp rank bound, arbitrary rectangular congruence, all-natural-degree convolution, and the zero-degree Dirac law.

- **Wishart analytic API.** Layer 5 now contains the reusable Gaussian quadratic-form theorem with its exact exponential-integrability domain, mgf, and cgf. Both Wishart families receive exact trace-mgf domains, mgfs, cgfs, cone-Laplace specializations, and the same principal-log characteristic-function formula through a shared analytic-continuation lemma. Their means and entrywise covariances are derived from the trace mgf. The overlap equality follows from characteristic-function uniqueness, and its Gaussian-sum `HasLaw` corollary is explicit.

- **Wishart singularity and measurability.** The roadmap adds the missing null-set theorem for singular symmetric matrices, allowing singularity of the Gram law whenever `min ν S.rank < p`. It also pins matrix-parameter measurability in coordinates and records the symmetric-scale form needed by kernels.

I have added corrective follow-ups to the two topology conversations because my earlier `b911c0c` replies described the superseded `Matrix.toMatrixInnerProductSpace` construction. I am leaving those conversations open for confirmation. The existing Tau Ceti-reuse thread also remains open for confirmation about why the current real-matrix development does not import the integer/rational-specific `TauCeti/LinearAlgebra/Matrix/PosDef.lean`.

The full local `lake build` passes at `0f0fb3a`; the warnings are the expected `sorry` warnings from roadmap prototypes.

TauCetiRoadmap/StandardDistributions/Suggested.lean
cameronfreer: The corrected direction is the right one. Retaining Matrix’s product topology and uniformity, retaining the induced subtype structures, and installing compatible Frobenius norm and inner-product data is preferable to installing a second topology via `Matrix.toMatrixInnerProductSpace`. The definitional-equality completion checks are valuable and make the intended coherence reviewable.

One contradictory summary sentence remains: the conventions section still says “The Frobenius inner product supplies the topology,” while Layer 6 correctly says that the product/subtype topology is retained. Please replace it with something like:

> The Frobenius norm and inner product are compatible with the retained product/subtype topology, but do not determine the normalization of Lebesgue measure.

There is also one remaining redundant instance in the prototype, discussed on the instance-specific thread: the explicit subtype `NormedSpace` should be removed, while the explicit `ContinuousENorm` should remain.

With those changes, I agree with the corrected topology architecture.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/README.md
cameronfreer: Confirmed. The two-family architecture now resolves the problem raised here:

* `nonsingularWishartMeasure` is explicitly the positive-definite-scale, real-degree density family, totalized to zero outside its density domain;
* `wishartGramMeasure` is the natural-degree Gaussian-Gram pushforward, including singular degrees and singular positive-semidefinite scales;
* the Gram family has probability, support/rank, arbitrary rectangular congruence, convolution, transform, overlap, and singularity targets;
* the overlap with the density family is an explicit equality, with a corresponding Gaussian-sum `HasLaw` corollary.

The non-positive-semidefinite totalization is unusual but is stated transparently and deliberately follows Mathlib’s `multivariateGaussian` totalization to `dirac 0`. The hypotheses on support, rank, congruence, and transforms are then placed exactly where they are mathematically needed.

This addresses the original architectural objection.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/README.md
cameronfreer: Confirmed. The roadmap now makes the comparison a named target with the correct direction and normalization:

`volume = 2^(p(p-1)/4) • symmetricLebesgue p`,

with a real exponent via `Real.rpow` and coercion to `ℝ≥0∞` for measure scaling. It also fixes the zero-dimensional normalization separately. This is enough to keep the raw-coordinate Wishart constants and the Frobenius volume unambiguous.

This finding is addressed.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/README.md
cameronfreer: Confirmed on the current head. The roadmap now makes the family entries exhaustive and says explicitly that the carrier rules only determine how listed targets are stated; they impose no hidden closed forms. I also checked the later family sections against that contract. The added cumulative-mass and finite-sum transform targets, multivariate covariance/domain targets, Weibull variance, positive-degree chi-squared characteristic function, and Wishart cgf rule close the concrete omissions raised here.

This addresses my original completion-contract objection. No further change is requested on this point.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/README.md
cameronfreer: Confirmed. The roadmap now names the two Jacobian modules and the concrete map/integral theorems, requires source and target regions, injectivity, derivative determinants, and measure equalities for each nonlinear map, and makes the Gamma–Beta result the exact joint product-pushforward theorem requested here. The zero-denominator branch is explicitly totalized and required to be null, with the marginal and independence conclusions derived afterward.

The symmetric-matrix section also now names the additive-Haar instance needed by this machinery. This addresses the finding.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/README.md
cameronfreer: The principal issue here is addressed: current-pin PR status has moved to the pull-request description, while the roadmap now uses the durable “consume when available; otherwise implement the matching local shape” rule.

Two smaller current-absence claims remain in the roadmap and should be made timeless as well:

* “Mathlib gives `Matrix` no measurable structure” should be replaced by the intrinsic API decision: state matrix-parameter measurability through the coordinate carrier and `Matrix.of`.
* “Mathlib has no general null-set theorem for polynomial zero loci” should be replaced by the requested proof strategy itself (diagonal expansion, Fubini, and induction), optionally allowing reuse of suitable imported infrastructure when available. The matching `Suggested.lean` comment should change too.

Thus the broad finding is resolved, subject to those two modest wording fixes.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/Suggested.lean
cameronfreer: Confirmed. Cholesky is now specified from Mathlib’s `LDL` decomposition, with the missing lower-triangular fact made an explicit prerequisite, and is packaged as an equivalence together with the homeomorphism and measurable equivalence. Both inverse identities, continuity/measurability, the coordinate homeomorphism, the exact derivative determinant, and the resulting measure-pushforward theorem are explicit targets.

This addresses the API and reuse concerns in the original comment.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/Suggested.lean
cameronfreer: The corrected product/subtype-topology architecture is sound, but the latest response is too broad in saying that every explicit bridge listed here is needed.

I tested the instance graph against the current Tau Ceti pin:

* the explicit `ContinuousENorm` really is needed under the retained subtype topology; it does not synthesize merely from the chosen normed additive group in this setup;
* the explicit public `symmetricMatrixNormedSpace`, however, is redundant. Once `symmetricMatrixInnerProductSpace` is installed, `NormedSpace ℝ (SymmetricMatrix p)` is supplied by `InnerProductSpace.toNormedSpace`, and it is definitionally the same `Submodule.normedSpace` structure used by the explicit declaration.

Please remove only `symmetricMatrixNormedSpace`. Keep the private ambient `matrixFrobeniusNormedSpace`, including its local use while constructing the ambient/subtype inner-product structure, and keep the explicit `ContinuousENorm` and other bridges required to reconcile the chosen norm with the retained topology and uniformity.

So the original comment was right about the subtype `NormedSpace` duplication but wrong about `ContinuousENorm`; the follow-up corrected the latter but should still remove the former.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/README.md
cameronfreer: Confirmed after checking the current files. The roadmap now imports/cites the relevant Gaussian/Pi, Giry-monad, triangular-matrix, and moment-determinacy material, and uses the corresponding APIs rather than shadowing them.

I agree with not importing `TauCeti/LinearAlgebra/Matrix/PosDef.lean`: that file is specialized to integer/rational matrix results and is not the real symmetric-matrix positive-definite infrastructure used by this roadmap. Its omission is not a reuse defect.

This thread can be resolved.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

TauCetiRoadmap/StandardDistributions/Suggested.lean
cameronfreer: Confirmed. `Θ` is now consistently bundled in the symmetric/self-adjoint carrier, and the roadmap asks for a reusable theorem deriving Hermiticity of `CFC.sqrt S * Θ * CFC.sqrt S` from positive semidefiniteness of `S` and the bundled symmetry of `Θ`. The characteristic-function statements construct that proof internally; callers no longer supply `hB`.

This finding is addressed.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

cameronfreer:
I reviewed `0f0fb3a` against the earlier comments and fzaiser’s corrective replies. The roadmap is substantially improved, and I would not ask for another architectural or completeness expansion.

The completion contract, Jacobian/change-of-variables dependencies, Cholesky package, volume normalization, bundled Wishart statistics, and the split nonsingular/Gaussian-Gram Wishart architecture are now satisfactory. The current-pin coordination material is appropriately kept in the PR description, and the conditional rule in the roadmap is durable.

I am requesting only these final edits:

1. Remove the redundant public `symmetricMatrixNormedSpace` instance. Retain the explicit `ContinuousENorm`; unlike `NormedSpace`, it does not synthesize in the corrected retained-topology setup.
2. Replace “The Frobenius inner product supplies the topology” with wording saying that the Frobenius structures are compatible with the retained product/subtype topology.
3. Make the two remaining current-absence claims timeless: the claims that Mathlib gives Matrix no measurable structure and has no polynomial-zero-locus null-set theorem. State the chosen coordinate API and proof strategy intrinsically instead.

One nonblocking clarification would also help: the Wishart congruence paragraph displays a `det (I + 2 • …)` identity while the mgf immediately below uses `I - (2*t) • …`. The plus identity is true, but displaying the signed/scaled instance actually used by the mgf—or saying explicitly that it follows by substitution—would make the proof route easier to follow.

After the three requested edits, I would be comfortable approving this roadmap.

🤖 Directed and approved by Cameron Freer; drafted and posted by GPT 5.6 Sol (xhigh)

kim-em (approved):
:robot: Reviewed and repaired at `fa269419608918bb1fb405cb9d33c2f568d41d7b`. The four remaining bounded cleanups are complete: timeless matrix-measurability and determinant-null-set wording, corrected retained-topology language, and removal of the redundant public symmetric-matrix `NormedSpace` instance. The affected Lean target builds locally against the merged current main. Approving.

fzaiser:
Thanks for the reviews and fixes @cameronfreer, @kim-em! 🎉

PR #218, opened 2026-08-13
Add roadmap: algebraic coding theory and code-lattice constructions

kim-em:
## Summary

Adds a literature roadmap for finite linear and additive codes, matrices and Hamming invariants,
Euclidean and Hermitian duality, MacWilliams identities, Type II codes, the tetracode, hexacode
and extended Golay codes, exact Construction A, and the bridge from codes to lattice discriminant
subgroups.

The primary carriers are Mathlib `Submodule` and `AddSubgroup` objects. The roadmap pins the
Hermitian orientation and conjugated parity-check convention, all Construction-A normalizations,
and literal quotient/isometry targets `A_{P_m(C)} ≅ C^⊥/C`. `Suggested.lean` exercises these
interfaces against Mathlib's sesquilinear/orthogonal APIs and the integral-lattice prerequisite.

## Dependencies

The integral-lattices and discriminant-forms specification, https://github.com/TauCetiProject/TauCetiRoadmap/pull/200, has merged; this PR targets `main` directly and consumes it from there.

## Sources and scope

Primary anchors include Huffman–Pless, MacWilliams–Sloane, Conway–Sloane, Ebeling,
Harada–Munemasa–Venkov, Munemasa–Tamura, and Dougherty–Gulliver–Harada. The roadmap deliberately
excludes decoding algorithms, theta series, code/lattice classifications, rank-24 glue tables,
Mathieu automorphism identifications, and categorical constructions.

## Verification

- `lake exe cache get`
- `lake build TauCetiRoadmap.AlgebraicCodingTheory.Suggested` — succeeds
- `git diff --check`

The `sorry` declarations are intentional roadmap targets.

## Review provenance

The complete draft was written by `gpt-5.6-sol` and independently red-teamed by a fresh
`gpt-5.6-sol` reviewer. A separate Sol correction pass resolved the Hermitian-duality API,
Mathlib-PR compatibility, Type-II scope, Construction-A modulus hypotheses, literal dual and
quotient interfaces, code/discriminant normalizations, and the hexacode conjugation issue. I then
read the resulting diff and reran its elaboration.

This is not a claim that a human subject expert has reviewed the mathematics. Review is especially
wanted from a coding theorist and Lean contributors familiar with finite-field, matrix,
sesquilinear, and finite-character APIs.

Authored with Codex.

PR #200, opened 2026-08-13
Add roadmap: integral lattices and discriminant forms

kim-em:
## Summary

This adds a roadmap for algebraic integral lattices, discriminant bilinear and quadratic forms,
finite-form calculations, and the overlattice/isotropic-subgroup correspondence.  It culminates in
explicit ADE discriminant-form calculations and the `D₈ ⊂ E₈` glue construction.

The roadmap uses an embedded full `ℤ`-submodule of a rational vector space as its primary carrier,
consumes Mathlib's `Submodule.IsLattice`, `BilinForm.dualSubmodule`, `CharacterModule`,
`QuadraticMap`, finite-quotient, and primary-component APIs, and keeps Tau Ceti's existing root data
as the owner of ADE combinatorics.

## Scope and conventions

- the discriminant group is the actual quotient `L^∨/L`;
- the bilinear form takes values in `ℚ/ℤ`;
- for an even lattice, the quadratic form uses the half-norm convention
  `q(x+L)=B(x,x)/2 mod ℤ`;
- odd lattices receive the bilinear discriminant form but no quadratic refinement;
- nondegeneracy is a predicate on raw finite-form data, so restrictions can be degenerate;
- integral and even overlattices correspond respectively to bilinear- and quadratic-isotropic
  subgroups;
- the induced discriminant form is identified with the form on `H^⊥/H`.

The accompanying `Suggested.lean` tests the carrier/dual/quotient design, Mathlib finite-form APIs,
the intermediate-lattice order isomorphism, orthogonal reduction, and the complete rank-one
normalization.

## Deliberate exclusions

This PR does not cover algebraic codes, rank-24 glue tables, the Niemeier classification, theta
series, or any categorical construction.  Those belong to separate roadmap PRs or to the research
programme rather than its published prerequisites.

## Sources and review priorities

The principal sources are Nikulin's *Integral symmetric bilinear forms and some of their
applications*, Conway--Sloane's *Sphere Packings, Lattices and Groups*, Ebeling's *Lattices and
Codes*, and Bourbaki's ADE plates.

Human review is especially welcome on:

- the rational embedded-carrier representation and its bridge to Mathlib;
- the half-norm/full-norm translation;
- hypotheses of the odd and even overlattice correspondences;
- finite quadratic-module and primary-decomposition interfaces;
- the ADE values and explicit `D₈^+ ≅ E₈` acceptance calculation.

## Verification

- `lake exe cache get`
- `lake build` — completed successfully, 8,537 jobs
[...]

PR #219, opened 2026-08-13
Add roadmap: Niemeier glue and explicit Leech constructions

kim-em:
> **Stacked-PR safety:** this roadmap is complete and ready for review. It is marked draft only because this repository automatically enables merge on every non-draft PR, while this PR temporarily targets an unprotected prerequisite branch. Keep it draft until its prerequisites merge; then retarget it to `main` and mark it ready.

## Summary

Adds a construction roadmap for all 23 rootful even unimodular rank-24 lattices from the explicit
Conway–Sloane ADE glue table, followed by an explicit Construction-B/odd-shift construction of the
rootless Leech lattice.

Every rootful row carries typed ADE discriminant coordinates, executable generator-pattern
semantics, subgroup order/isotropy/Lagrangian and coset-minimum checks, a literal root-sublattice
target, and code identifications where applicable. The roadmap also catches the exceptional
`D₄⁶` convention: Conway–Sloane’s displayed generators span over the pinned `F₄`, not merely as an
additive group. The Leech layer proves minimum norm four through four explicit Golay
parity/weight cases, without theta series or classification.

## Dependencies

This roadmap consumes #200 and #218. PR #219 is stacked on the #218 feature branch, whose history already contains #200, so its checkout contains both imported `Suggested.lean` interfaces and its review diff contains only the rank-24 roadmap. After those prerequisites merge, retarget this PR to `main` and mark it ready; review-driven interface changes will be rippled forward.

## Sources and scope

Primary anchors are Conway–Sloane (including Table 16.1 and the 1982 Table V), Niemeier,
Chenevier, Wilson, and Ebeling. This is explicitly a construction roadmap, not a classification:
it does not prove exhaustiveness or uniqueness, Venkov’s theorem, theta-series results,
automorphism-group identifications, Moonshine, or any categorical construction.

## Verification

- `lake exe cache get`
- `lake env lean TauCetiRoadmap/Rank24LatticeConstructions/Suggested.lean`
- integrated `lake build` on the exact stacked checkout — succeeds (8743 jobs)
- `git diff --check`

The `sorry` declarations are intentional roadmap targets.

## Review provenance

The complete draft was written by `gpt-5.6-sol` and independently red-teamed by a fresh
`gpt-5.6-sol` reviewer. That review independently recomputed all 23 rows and found the `D₄⁶`
closure error; a Sol correction pass fixed it along with ADE typing, code-coordinate isometries,
cyclic-factor structure, and executable pattern expansion. A further fresh Sol verification found
no remaining semantic defect. I read the resulting diff and checked its dependency/build state.

This is not a claim that a human subject expert has reviewed the mathematics. Review is especially
wanted from a Niemeier/Leech lattice specialist for a character-for-character table audit, and
from Lean contributors familiar with dependent finite products, quotients, and proof-producing
finite computation.

Authored with Codex.


CBirkbeck (commented):
I independently implemented the finite discriminant groups, the A/D/E quadratic values and coset-minimum formulas, all pattern expansions, the exceptional F4-linear D4^6 closure, and the generated subgroups for all 23 rows. Every internal finite check in the roadmap passed:

- the generated subgroup has the displayed order in all 23 rows;
- every subgroup word is quadratic-isotropic;
- |H|^2 equals the discriminant-group order in every row;
- every nonzero glue word has attained coset minimum at least 4;
- each component list has rank 24 and a common Coxeter number;
- every displayed root count agrees with the component formula;
- for D4^6 the ordinary additive span has order 32, while the F4-linear closure has order 64, exactly as the revised roadmap says.

The four-case Leech minimum-norm argument is also internally correct: the shift has norm 4, and in each parity/Golay case the congruence condition forces raw squared norm at least 32.

I therefore found no mathematical blocker in the roadmap and would approve it after #218 lands. Two cautions should remain explicit:

1. My audit was an independent check of the internal table semantics, not a character-for-character comparison with the printed Conway-Sloane table. A lattice specialist should still audit the transcription against the source.
2. The finite table checks do not replace the genuinely hard Layer-1 theorem that each displayed coset norm is attained and minimal, especially the E6/E7 minuscule-weight cases, nor the Layer-3 equality of the actual root submodule with the embedded ADE root lattice. Keep those as non-negotiable dependencies.

A small wording point: the README says the five row-check proof objects are stored beside the data. Suggested.lean currently presents table data plus global theorem families rather than proof fields in each row. Either soften that sentence or bundle a row certificate so the implementation matches the stated architecture.

CBirkbeck (commented):

**Assessment: ready for human approval on mathematical/specification content, after the prerequisite branch is integrated. Keep the current draft/stacked-PR safety arrangement until #218 merges and this PR can be retargeted.** This is a COMMENT review, not an approval or a request to change draft state now.

I independently recomputed the finite data in the current table using exact Python arithmetic, including the parity-dependent D-type addition, cyclic/even-permutation expansions, and the extra F₄ closure in row 19. All 23 rows pass: the generated subgroup has the stated order, its order squared is the discriminant order, every word is quadratically isotropic, every nonzero word has the displayed coset-norm expression at least 4, the total rank is 24, and the root count matches the component formula. The minimum nonzero glue norm is 6 for D₂₄, 4 for every other nontrivial-glue row, and there is no nonzero glue word for E₈³.

I separately checked the explicit coordinate comparisons against the actual coding-roadmap matrices. Applying the displayed destination-to-source coordinate formula sends the row-23 code to the binary Golay code; the row-22 permutation and scalings send its code to ternary Golay; and the D₄ label map followed by `[ω,ω,1,ω²,ω²,1]` sends the 64-word subgroup to the hexacode. These are exact subspace equalities, not just matching parameters. They are independent arithmetic checks, not Lean certificates or a character-for-character audit of the printed Conway–Sloane source.

The Construction-B/odd-shift argument is coherent with the rational form `dot/2`. The index-two condition is the coordinate sum modulo four; the shift `(-3/2,1/2,…,1/2)` has norm four, pairs integrally with K, and doubles into K. The congruence description agrees with the two cosets. For even coordinates the zero-code case excludes a single ±4, and the nonzero-code case uses eight coordinates of absolute value at least two. For odd coordinates, the all-absolute-value-one candidate has sum zero modulo eight by doubly-evenness, so a coordinate of absolute value at least three is required. Thus all four cases give raw squared norm at least 32, exactly the intended lattice lower bound four.

The genuinely difficult suppliers remain explicit rather than being replaced by these computations: attained/minimal ADE coset norms in Layer 1, the actual discriminant-coordinate isometries, and equality of the root submodule with the embedded ADE lattice in Layer 3. This separation is important. The previous review's row-certificate wording has also been fixed: the README now consistently specifies theorem families over the table index.

Nonblocking wording improvement: the binary coordinate paragraph can make the direction clearer by writing the word formula `output[j] = input[p[j]]` and naming input as table coordinates and output as Golay coordinates. The actual `a1TableToGolayMonomial` uses the inverse equivalence in its `reindex` field, as it should; my finite check used that formula.

No remaining mathematical blocker found. I have not rerun a local Lean build, and these checks do not establish that the planned coset-minimum or gluing theorems have already been implemented.

CBirkbeck:
Let me know once its out of draft and I can approve it.

CBirkbeck (commented):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `e05a6caf2ea25126d1ed04dfc40b2cfacd7ddea3`.

Ready for Chris's approval. The previous procedural hold is gone: #218 is merged, this PR targets `main`, and it is no longer draft. The PR description's stacked/draft warning is now stale and can be removed.

I read the full current README and Suggested.lean, independently recomputed all 23 generated subgroups using exact rational arithmetic, and checked the three monomial code comparisons against the actual coding matrices at this head. All checks pass: rank/common Coxeter number/root count, the stated subgroup order, order-squared equal to discriminant order, quadratic isotropy of every subgroup word, and nonzero glue-coset expression at least four. The minimum expression is six for D₂₄, four for every other nontrivial-glue row, and absent for E₈³. The D₄⁶ raw additive span has 32 words and its required F₄ span has 64. The binary, ternary and hexacode comparisons are exact subspace equalities after the displayed coordinate maps.

I additionally compared the 23 component rows, generator strings and root counts visually with Conway–Sloane's 1982 Table V, printed p. 91, using the rendered author PDF. They match, with the D₄⁶ scalar-span interpretation explicitly handled by this roadmap. This is a comparison with Table V, not a claim to have inspected the separate book's Table 16.1.

The Construction-B and odd-shift argument uses a consistent `dot/2` normalization. The shift has norm four, pairs integrally with K, and doubles into K; its coset is exactly the odd branch of the congruence description. The four Golay parity/weight cases then give the claimed raw lower bound 32. The hard attained/minimal ADE-coset theorem and literal root-submodule equality remain explicit targets rather than being replaced by these arithmetic checks.

The named `ofIsotropicSubgroup` construction is available at this branch's actual Tau Ceti pin `5cb65f0393fd8deeaf3fb6398ba48bad654e479d`, which I checked earlier in this review run. The exact head's CI succeeds.

Minor certificate improvement only: expose the coercion equation identifying `leechShiftInDual` with `leechShift`; the README already requires that actual vector, but the displayed bundled definition currently leaves it implicit.

These Python checks are independent tests, not Lean proofs. No separate local Lean build was run. This is a COMMENT review, not an approval.

PR #220, opened 2026-08-13
Add roadmap: DG and A-infinity algebras and categories

kim-em:
> **Stacked-PR safety:** this roadmap is complete and ready for review. It is marked draft only because this repository automatically enables merge on every non-draft PR, while this PR temporarily targets an unprotected prerequisite branch. Keep it draft until its prerequisites merge; then retarget it to `main` and mark it ready.

## Summary

Adds a literature roadmap from signed graded multilinear algebra through DG and A-infinity
algebras, categories, modules and bimodules, minimal models, bar/cobar and twisting, derived Morita
theory, quotients, Hochschild deformation theory, smooth/proper and Calabi–Yau structures, Euler
theory on perfect objects, and relative quadratic Koszul duality.

The roadmap pins one cohomological sign system all the way from a degree-`-1` suspension and the
square-zero bar coderivation to arities 1–4, Maurer–Cartan twisting, shifted Hom complexes, cones,
and Euler/Serre identities. `Suggested.lean` exercises actual tensor-word direct sums,
deconcatenation and coderivations, square-zero module/bimodule structures, a genuine two-object DG
cone, and the perfect-category Euler interface.

## Dependencies

This roadmap consumes the Grothendieck-group and Euler-form specification in #201 and current Tau Ceti quiver Euler APIs. PR #220 is stacked directly on the #201 feature branch, so its checkout and CI contain the imported interface while its review diff remains limited to DG and A-infinity material. After #201 merges, retarget this PR to `main` and mark it ready; review-driven dependency changes will be rippled forward.

## Sources and scope

Primary anchors include Getzler–Jones, Keller, Kadeishvili, Gugenheim–Lambe–Stasheff, Merkulov,
Lyubashenko, Drinfeld, Toën, Bondal–Kapranov, Shklyarov, Brav–Dyckerhoff, BGS, and Positselski.
Curved and periodic objects, particular zigzag/McKay constructions, the affine-E8 finiteness
conjecture, imaginary-root quotients, and lattice categorification are deliberately outside this
roadmap.

## Verification

- `lake exe cache get`
- `lake env lean TauCetiRoadmap/DGAInfinity/Suggested.lean`
- `lake build` — succeeds on the exact stacked checkout
- `git diff --check`

The `sorry` declarations are intentional roadmap targets.

## Review provenance

The complete draft was written by `gpt-5.6-sol` and independently red-teamed by a fresh
`gpt-5.6-sol` reviewer. A separate Sol correction pass resolved the reported suspension/MC signs,
[...]

PR #201, opened 2026-08-13
Add roadmap: Grothendieck groups and Euler forms

kim-em:
## Summary

This adds a roadmap for intrinsic Quillen exact structures, small Grothendieck-group
presentations, resolution comparison, finite-dimensional-algebra Cartan maps, ordinary and graded
Ext-Euler forms, and left/right numerical quotients.

It consumes current Mathlib categorical, Ext, Euler-characteristic, Laurent-polynomial, and
sesquilinear infrastructure, and Tau Ceti's existing quiver conventions and Euler-form
calculations.  The roadmap establishes the exact homological language needed by later DG,
stable/periodic, hopfological, zigzag, and McKay roadmaps without importing those subjects here.

## Scope and conventions

- Quillen conflations contain actual kernel/cokernel universal data;
- split and canonical abelian exact structures remain distinct;
- categorical K-zero uses a small `Shrink (Skeleton C)` presentation and a universal property;
- modules are left modules, and projectives are columns of the Cartan matrix;
- over a nonsplit field, Hom dimension includes the `dim_k End(S_i)` factor;
- finite projective resolutions imply determinant `±1`, not necessarily `+1`;
- Ext termwise finite-dimensionality and eventual vanishing are separate hypotheses;
- `[M{1}]=q[M]`, with bar-semilinearity in the first variable;
- algebraic evaluation at `q=1` is distinct from comparison with an ungraded category;
- left and right radicals and numerical quotients remain distinct.

The accompanying `Suggested.lean` exercises intrinsic conflations, exact K-zero and its universal
property, resolution comparison, honest Euler-finiteness predicates, numerical quotients,
Laurent-sesquilinear forms, conditional forgetting-of-grading data, and an orientation-sensitive
`A₂` calculation.

## Deliberate exclusions

This PR does not construct DG or A-infinity categories, perfect categories, stable or periodic
categories, hopfological algebra, zigzag/preprojective algebras, McKay correspondence, or any
rank-24 categorification.  Those are separate dependent roadmaps or research questions.

## Sources and review priorities

The principal sources are Weibel's *The K-book* and *An Introduction to Homological Algebra*,
Bühler's *Exact Categories*, Dancso--Licata's *Koszul algebras and flow lattices*,
Assem--Simson--Skowroński, and Webb.

Human review is especially welcome on:

- the intrinsic E0/E1/E2 exact-structure interface;
- universe and choice-independence details of categorical K-zero;
- hypotheses of the generic resolving-subcategory theorem;
[...]

PR #221, opened 2026-08-13
roadmap: add stable, periodic, and curved homological algebra

kim-em:
This PR adds a roadmap for stable, periodic, and curved homological algebra: additive morphism ideals and Frobenius exact categories with Happel's triangulated stable category; self-injective algebras, Gorenstein-projectives, complete resolutions, and the Buchweitz and Keller–Vossieck singularity comparisons; differential modules and genuine `n`-periodic complexes with their homotopy and derived categories; compression, the shift-orbit category, and Stai's triangulated hull; right curved DG algebras and modules with their homotopy category and the absolute, co-, and contraderived categories of the second kind; and matrix factorizations with the Eisenbud, Orlov, and Buchweitz hypersurface comparisons.

The roadmap pins right-module and cohomological conventions, keeps the stable/additive, homotopy, Verdier, singularity, periodic-derived, and curved second-kind quotients distinct, and consumes the exact-category and internally graded algebra interfaces of the Grothendieck–Euler-forms and DG/A-infinity roadmaps. `Suggested.lean` constructs the additive ideal quotient, the periodic mapping cone and homotopy category, the orbit category with its direct-sum hom formula and universal property, the curved module and homotopy categories, and the closure-generated second-kind acyclic classes with their Verdier quotients and comparison functors, and states the flagship `k[x]/(x^n)` calculations.

Drafted with gpt-5.6-sol (Codex) and revised with Claude Fable 5.1 (Claude Code).

https://claude.ai/code/session_01F22JuNwb5LEGCqRAodoBBH


CBirkbeck (changes requested):
I think this still needs major changes to Suggested.lean. The README has a sensible mathematical scope, but several representative signatures do not force the advertised constructions.

1. PeriodicComplex.mappingCone is impossible at its stated generality. It is declared with only Category and Preadditive assumptions, but the cone needs terms of the form Y_i directSum X_{i+1}. The HasBinaryBiproducts hypothesis appears only later on periodicQuasiIso, after the under-specified cone has already been defined. Put the biproduct hypotheses on mappingCone itself (or build it in a category already carrying them).

2. StaiOrbitConstruction does not characterize an orbit category. It stores an arbitrary category, a projection, a chosen descent functor, and a factorization isomorphism, but no uniqueness, no coherence with the supplied periodicity isomorphism, and no literal Hom formula sum_j Hom(X, Sigma^{nj}Y). A fake category can satisfy this record. Either construct the orbit category with the actual morphism formula and composition, or state the full universal property including uniqueness and compatibility.

3. The principal Stai endpoint is missing. The source theorem is about the embedding of the orbit category into its triangulated hull through periodic complexes. Suggested.lean gives full faithfulness and an essential-image predicate, but does not put a triangulated structure on the periodic derived category or prove that the generated triangulated closure is the hull/all of the intended target. That theorem needs a representative signature.

4. The second-kind interface is still too parametric. SecondKindAcyclicClasses contains arbitrary sets, and SecondKindComparisonTarget takes arbitrary categories and arbitrary localization classes without connecting them to the absolute/coacyclic/contraacyclic sets. It does not construct the closures generated by totalizations of short exact sequences or the three concrete Verdier quotients. The public API should define those specific acyclic classes and derive the comparison functors from their inclusions.

5. The public PeriodicComplex carrier admits n=0, while the roadmap and source use positive finite periods. Since ZMod 0 behaves as integer indexing rather than a finite cyclic grading, require 0<n/NeZero n or explicitly separate the n=0 ordinary-complex specialization.

6. The periodic derived category is used as the target of a triangulated-hull theorem, but Suggested.lean does not seed the needed preadditive/shift/cone/triangulated structures on that localization.

The Keller-Vossieck and Buchweitz sections are much better grounded than earlier generic comparison records. My requested changes are concentrated in periodic orbit/hull and second-kind localization architecture. The PR description is also stale: it still describes an open stacked dependency state although #201 and #220 are merged and this PR is non-draft on main.

CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `8311e54bb2ae57398882b10c8cfdbafa20477b85`, the complete README and `Suggested.lean`, the previous review, and the relevant Mathlib definitions at the actual pin `de5ce8a9a66a4aa68a9bdbb35b63a06d34d9ca11`. This is a mathematical/source-level interface review, not a local Lean build.

The earlier requests concerning the cone's missing biproduct assumption and the incomplete orbit/hull and second-kind construction contracts remain applicable at this head. I will not duplicate that entire review. The right-curvature signs, opposite translation, separate periodic/curved quotients, and restricted Eisenbud/Orlov statements are important distinctions worth preserving.

I found three additional concrete interface problems.

### 1. The generic localization comparison incorrectly requires strict factorization

**Locations:** `SecondKindComparisonTarget.facCo`, `.facContra`, and `secondKindComparisonTarget`.

The inputs are arbitrary localization functors satisfying `IsLocalization`, but the output requires equalities of functors:

```
qAbs ⋙ absToCo = qCo
qAbs ⋙ absToContra = qCtr
```

General localization supplies **natural isomorphisms**, not these strict equalities. This is a false target at the displayed generality, independently of the earlier issue that the acyclic classes are not tied to their generating closures.

For a small counterexample, let `Hot` be the groupoid with two distinct objects and exactly one morphism between each ordered pair. Let `Dabs` be the one-object terminal category and let `qAbs` collapse the two objects. Set `Dco = Dctr = Hot`, with `qCo = qCtr = id`, and take all three weak-equivalence classes to be all morphisms. Every morphism is already invertible, and all three functors are equivalences, hence localization functors. The set inclusions in `SecondKindAcyclicClasses` can, for example, all be equalities.

Nevertheless, no functor `F : Dabs → Hot` can satisfy `qAbs ⋙ F = id`: its composite sends both source objects to the same object, whereas the identity preserves the two distinct objects. A natural isomorphism exists, exactly as localization predicts.

**Repair:** replace both equality fields by natural isomorphisms and use Mathlib's `Localization.fac` / `Localization.Lifting.iso` API. Strict equations can be exposed separately for a specifically chosen formal construction when justified, not demanded for every equivalent presentation of a localization. Test transport of the quotient to an equivalent category.

Pinned API: [Localization/Equivalence.lean](https://github.com/leanprover-community/mathlib4/blob/de5ce8a9a66a4aa68a9bdbb35b63a06d34d9ca11/Mathlib/CategoryTheory/Localization/Equivalence.lean).

### 2. `CompleteProjectiveResolution` omits finite generation of its terms

**Location:** `FrobeniusComparison.CompleteProjectiveResolution`.

The README defines the complete resolution using finitely generated projectives. The structure has `componentProjective`, acyclicity, dual acyclicity, and the cycle identification, but no `componentFinite` field. Finite generation of the displayed cycle does not force finite generation of every term of a chosen resolution: add a contractible two-term disk on an infinite-rank free module, in degrees away from the specified cycle, to any finite complete resolution. All current fields survive, including dual acyclicity, but the resulting complex no longer belongs to the promised category of complexes of finitely generated projectives.

Please add

```
componentFinite : ∀ n : ℤ, Module.Finite Rᵐᵒᵖ (complex.X n)
```

or use the actual finite-projective full subcategory for the complex. The finite-projectivity hypothesis also justifies reducing total acyclicity to the displayed test against `R` when constructing the comparison with other projective targets.

This is a mismatch in the resolution carrier, not a claim that the stated Buchweitz theorem under Iwanaga–Gorenstein hypotheses is false. A more general resolution theory would need its own comparison theorem rather than silently replacing the specified finite one.

### 3. The fixed module-carrier universe contradicts the generic ring universe

**Locations:** `FinitelyGeneratedRightModule`, `CompleteProjectiveResolution`, `FiniteMatrixFactorization`, and the finite-global-dimension hypotheses using `ModuleCat.{0}`.

The ring is allowed to be `R : Type u`, but these categories and quantifiers force module carriers to lie in `Type 0`. At the pin, `ModuleCat.{v} R` really stores `carrier : Type v`; it does not automatically enlarge that carrier universe to include `R`.

Thus for genuinely large `R`, the regular module `R_R` and its finite free powers are not even objects of the displayed category. This is not the essentially-small replacement of the category of all finitely generated modules, and the global-dimension quantifier only tests the accidentally restricted carrier universe.

**Repair:** use `ModuleCat.{u} Rᵐᵒᵖ` for the finite-module constructions, or a suitably general module universe with explicit small-model equivalences where needed. Alternatively restrict the entire ring statement to `Type 0` deliberately. Do not combine an unrestricted ring universe with a silently fixed module universe.

Acceptance should include the regular module and its finite free powers at a nonzero universe level, alongside the existing `k[x]/(x^n)` calculation.

Pinned definition: [ModuleCat/Basic.lean](https://github.com/leanprover-community/mathlib4/blob/de5ce8a9a66a4aa68a9bdbb35b63a06d34d9ca11/Mathlib/Algebra/Category/ModuleCat/Basic.lean).

These repairs are in addition to, not substitutes for, the previously requested concrete orbit/hull and second-kind closure constructions.

kim-em:
:robot: Claude+Codex say:

Thanks for both rounds. Changes are in acfc49e and the follow-up commit on this branch (after merging `main`). Point by point, with what is and is not done:

First review:

1. `PeriodicComplex.mappingCone` takes `HasBinaryBiproducts C` itself and is defined concretely, `cone(f)ⁱ = Xⁱ⁺¹ ⊞ Yⁱ` with differential `[[-d_X, 0], [f, d_Y]]`, with `d² = 0` proved; `coneInl` and `coneFst` are defined. Done.
2. `StaiOrbitConstruction` is replaced by the constructed orbit category `ShiftOrbit D n`: objects of `D`, `Hom(X, Y) = ⨁_{j ∈ ℤ} Hom_D(X, Y⟦n j⟧)` by `rfl` (`hom_eq`), composition of homogeneous pieces `f ≫ g⟦n i⟧' ≫ (Z⟦n j⟧⟦n i⟧ ≅ Z⟦n(i+j)⟧)` extended bilinearly (`of_comp_of` is proved), the projection, the periodicity isomorphism with its degree `±1` components written out, and the universal property: `desc`, `descFac`, the compatibility of the factorization with the periodicity isomorphisms, and uniqueness up to a unique compatible isomorphism (`desc_unique`, an `∃!`). The category laws, preadditivity, functoriality of the projection, the periodicity iso proofs, and the descent theorems are the remaining `sorry`s.
3. `Dₙ(mod-Λ)` now carries `Preadditive`, `HasZeroObject`, `HasShift ℤ`, `Pretriangulated`, and `IsTriangulated` instances. `staiCompression` is defined as the descent of compression along the orbit category, so `staiCompression_fac` is a proof, not a field. The hull is stated as full faithfulness (`staiCompression_full_faithful`) plus generation under shifts and cones alone (`staiCompression_essImage_triangClosure_eq_top`, with `triangClosure` the smallest isomorphism-closed triangulated subcategory, no retracts), from which the thick-envelope version is derived. `Dₙ(mod-Λ) ≃ Kₙ(proj-Λ)` is stated for the canonical composite `Kₙ(proj-Λ) ↪ Kₙ(mod-Λ) → Kₙ/Acₙ`. The README says what "triangulated hull" means and that no Karoubi completion is involved. What is not here: a Lean statement of Keller's DG hull itself; the README states that comparison in the pretriangulated-envelope vocabulary supplied by the DG roadmap.
4. `SecondKindAcyclicClasses` and `SecondKindComparisonTarget` are gone. There is a concrete category of bundled right curved modules, its homotopy category with the triangulated package installed, short exact sequences and their totalizations, and `absoluteAcyclic`/`coacyclic`/`contraacyclic` defined as the closures generated by the totalizations (`triangEnvelope`, and infima over small-coproduct- and small-product-closed triangulated subcategories, small meaning indexed by the hom universe). `D_abs`, `D_co`, `D_ctr` are the `trW` localizations; `absoluteToCoderived` and `absoluteToContraderived` are `Localization.lift` along the inclusions, with triangle-functor targets. The coproduct and product instances, the curved comparison functors, and the concrete Positselski theorem statements use carriers in `Type (max u v)`, the universe of the algebra, so the free modules and the required coproducts exist there.
5. `PeriodicComplex` requires `[NeZero n]`; the `hn : 0 < n` arguments are gone.
6. `Kₙ(C)` and `Dₙ(E)` both have the shift/cone/pretriangulated/triangulated structure installed as instances. The cone triangle `periodicConeTriangle f` is defined with the actual arrows and stated distinguished; the componentwise split exact structure is named, its conflations characterized, its Frobenius data stated, and the comparison from its stable category to `Kₙ(C)` is defined by the ideal-quotient universal property and stated to be a triangle equivalence. `Dₙ` has both descriptions with a comparison that is stated to be a triangle equivalence (`IsTriangulated`, not only `CommShift`).

The PR description is rewritten to describe the PR as it stands.

Second review:

1. The strict factorization equalities are gone with item 4: `absoluteToCoderived_fac` and `absoluteToContraderived_fac` are the natural isomorphisms from `Localization.fac`.
2. `CompleteProjectiveResolution.componentFinite` is added, with the contractible-disk counterexample recorded in the docstring and README.
3. Every `ModuleCat.{0}` is now `ModuleCat.{u}` for `R : Type u` (finitely generated modules, complete resolutions, matrix-factorization components, `HasEnoughFiniteProjectives`, and the global-dimension hypothesis, now `HasFiniteGlobalDimension Λ d`). The README's universe section states the convention.

Also changed on Codex's advice while checking this: the graded-opposite target is now a structure with the Koszul-twisted multiplication and the curvature sign as a field to prove, rather than a definition of `-op w`; two lemmas under Theorem 7.9 that only re-applied their hypothesis are removed; conditions `(*)` and `(**)` now carry one bound for all countable sums or products, as in the theorem; and the contractible periodic complexes are identified with the projective-injectives of the split structure.

CBirkbeck (commented):

**Assessment: the revised roadmap is ready for human approval on mathematical/specification content.** The previous two reviews' blocking interface objections have been addressed. This is a comment review, not an approval; the current merge conflicts still need separate resolution.

I checked the revised README, the replacement interfaces in `Suggested.lean`, the earlier reviews and response, and the relevant statements in Stai and Positselski. In particular:

* The cone now has its own `HasBinaryBiproducts` context, explicit terms/differential/arrows, and positive period is enforced by `NeZero n`. The periodic homotopy and derived categories receive actual Mathlib triangulated structures rather than an unrelated proposition saying they are triangulated.
* `ShiftOrbit` has the direct-sum Hom carrier and homogeneous composition. Descent includes compatibility with the periodicity isomorphism and uniqueness; Stai compression is its actual descent, not a free choice of functor. The target now includes the generated triangulated-closure theorem and the canonical projective-periodic comparison. The finite-global-dimension restriction remains visible.
* The absolute, co-, and contraacyclic classes are the specified closures of totalizations. Their localizations and inclusion-induced comparison functors are connected to those classes. The comparison factorizations are natural isomorphisms, resolving the earlier strict-equality counterexample.
* Complete projective resolutions now require finite terms, and the finite-module carriers use the ring's universe. The right-curvature convention is consistent: expanding the module Leibniz rule twice gives `d_A²(a)=aw−wa`, and passing to the graded opposite changes the stored curvature to `−op(w)`.

The source checks support the revised restrictions: Stai Proposition 3.10 and Lemma 3.12 explain the relatively-projective model and orbit embedding under finite global dimension; Positselski Theorem 7.8 retains the additional grading restrictions for comparison with ordinary acyclicity, rather than claiming that comparison for every zero-curvature DG algebra. The matrix-factorization part still keeps the local Eisenbud, flat-superpotential Orlov, and Iwanaga–Gorenstein Buchweitz results separate.

Three nonblocking precision improvements would make the implementation instructions safer:

1. Say explicitly that the homotopically-projective/projective-flag/relatively-projective comparison is **up to isomorphism in the homotopy category** where appropriate, not an equality of underlying-module predicates on arbitrary representatives. A contractible disk on a nonprojective module is homotopically projective but does not have projective underlying module. The actual periodic-projective full-subcategory comparison is the right formulation.
2. Add an odd-period, characteristic-not-two acceptance test: the periodicity datum of compression should not be conflated with a natural identification of the entire target's n-fold triangulated shift with the identity. Stai §5.3 specifically explains this sign phenomenon.
3. Explain the uniform-bound reformulation of Positselski's conditions (*) and (**). The printed source quantifies finite dimension for each countable sum/product. A uniform bound follows by a countable diagonal argument: otherwise combine examples of arbitrarily large dimension into one countable sum/product and use that each selected example is a direct summand. The current stronger-looking formulation is therefore reasonable, but this short lemma should justify the claimed exact correspondence with the source.

Sources checked: Stai, https://arxiv.org/html/1510.03574, §§3–5; Positselski, https://arxiv.org/html/2207.07063, §7.6, Theorems 7.8–7.9. These are source/interface checks, not a local `lake build` or a claim that the roadmap's target theorems have been proved.

CBirkbeck (approved):

PR #222, opened 2026-08-13
roadmap: add hopfological algebra and categorification at roots of unity

kim-em:
This PR adds a roadmap for hopfological algebra: integrals and the Larson–Sweedler Frobenius theorem for finite-dimensional Hopf algebras; the tensor-triangulated stable categories `H-StMod` and `H-stmod`; module algebras, comodule algebras, and smash products; the relative category `C(A,H)` of smash-product modules modulo the integral null ideal and its derived localization `D(A,H)`; strict cofibrancy, cellular and homotopy-invariant property (P), compact generation, and finiteness; Grothendieck rings and modules; the super, characteristic-`p`, Sweedler, Nichols, and Taft examples; and the Laugwitz–Qi cyclotomic construction with its tensor ideal, Verdier quotient `O_n`, and `K₀(O_n) ≅ ℤ[ν,ν⁻¹]/(Φ_n)`.

Every stable category is an instance of the stable, periodic, and curved roadmap's construction (PR https://github.com/TauCetiProject/TauCetiRoadmap/pull/221, on whose branch this one is based), with Happel's triangulation, Verdier quotients as `trW` localizations, and Grothendieck groups from the Grothendieck–Euler-forms roadmap. The roadmap keeps strict cofibrancy apart from the homotopy-invariant property (P), with the dual-numbers modules `A/(t)` over `H = k` and `A/(t) ⊗ H` for nonsemisimple `H` as negative tests, and states braided Hopf algebras as Hopf monoid objects in the braided category of graded vector spaces, with a bridge to the hand-written record. `Suggested.lean` builds the smash-module category, `C(A,H)`, `D(A,H)`, the cofibrancy and compactness targets, the Laugwitz–Qi stable category and quotient, and the graded-vector-space bridge, alongside the integral, smash, Nakayama, and root-of-unity arithmetic.

Drafted with gpt-5.6-sol (Codex) and revised with Claude Fable 5.1 (Claude Code).

https://claude.ai/code/session_01F22JuNwb5LEGCqRAodoBBH


CBirkbeck (changes requested):
This needs major additions to Suggested.lean. The README describes the right programme, but the representative file does not seed the main categorical output.

The following central objects/theorems are absent from Suggested.lean:

- the actual Frobenius exact category C(A,H) of smash-product modules with A-split conflations;
- the relative homotopy quotient by the integral/null ideal;
- quasi-isomorphisms and the localization D(A,H);
- property-(P), cofibrant replacements, and the equivalence with the derived localization;
- the monoidal stable categories H-StMod and H-stmod;
- the Grothendieck group/ring and module structures;
- the concrete subcategories I_k, their finite-direct-sum tensor ideal I, proof that its stable image is thick and tensor-closed, and the Verdier quotient O_n;
- the flagship theorem K0(O_n) = Z[nu,nu^{-1}]/(Phi_n).

The file currently reaches the Laugwitz-Qi cells and cross-piece factorization theorem, then ends with three polynomial relations. That does not architecture-check the theorem advertised in the title and completion criteria. Add representative declarations for the actual categories, quotient functors/universal properties, tensor ideal, O_n, and its K0 isomorphism.

There is a second foundational concern. BraidedHopfObject is currently a hand-built record of graded pieces, a bicharacter-twisted tensor multiplication, and Hopf equations. Before using it to support monoidal stable categories and tensor ideals, the roadmap should either construct the corresponding braided monoidal category and exhibit this as a genuine Hopf object there, or prove a precise bridge to a generic categorical carrier. Otherwise this risks becoming a parallel, disconnected braided-Hopf foundation.

The arithmetic correction around xi_k is good: the roadmap correctly does not claim xi_k has order p_k in nonsquarefree cases and instead uses the self-braiding parameter of order p_k. I found no comparable concrete formula error in the integral/Nakayama material.

This PR is stacked on #221, so the new categorical targets should consume the repaired stable/periodic APIs rather than introduce another quotient theory.

CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `cdbf8509ef04c6f9056e1e3126bd62db4844a404`, both HopfologicalAlgebra files, the index/import changes, the existing review, and Qi's definitions and cofibrant-module discussion. This is a mathematical/source-level review, not a local Lean build.

The earlier request to expose the actual categorical outputs and their comparison functors remains unresolved at this head; the representative file still stops short of those constructions. I am not repeating that list. The handed Nakayama bridge, the action-dependent smash product, the distinction between the relative and ordinary stable quotients, and the corrected nonsquarefree root arithmetic should remain accepted.

There is a further **mathematical incompatibility** in Layer 5 that needs to be settled before adding its categorical declarations.

### Homotopy-invariant property (P) does not imply the displayed strict cofibrancy

**Location:** README, “Cofibrant and property-(P) modules”, specifically the definition using isomorphism in `C(A,H)`, followed by “property (P) implies cofibrant” and the strict direct-summand characterization.

The roadmap defines:

* cellular property-(P): an actual split filtration with the specified cells;
* property (P): being **isomorphic in the homotopy category** to such a filtered module;
* cofibrant: strict lifting of `B`-linear maps against every surjective quasi-isomorphism.

The second condition does not imply the third. Strict cofibrancy is not invariant under adding arbitrary relatively contractible summands.

Here is a counterexample satisfying the standing hypotheses. Take `H = k` with its usual one-dimensional Hopf structure, `A = k[t]/(t²)`, and `B = A#H = A`. Then:

1. The chosen integral is `1`, so `λ_M : M → M ⊗ k` is an isomorphism. Every map is relatively null-homotopic and `C(A,k)` is the zero category. Hence every module is isomorphic there to the zero cellular object and has property (P) in the roadmap's definition.
2. `k-StMod` is also zero, so **every** `A`-linear map is a quasi-isomorphism under the displayed restriction test.
3. Strict cofibrancy therefore reduces to ordinary projectivity as an `A`-module. The module `M = A/(t)` is not projective: lifting its identity through the surjective quasi-isomorphism `A → A/(t)` would split that quotient. An `A`-linear splitting would send `1` to an element annihilated by `t`, hence to a multiple of `t`, whose image in `A/(t)` is zero, a contradiction.

Thus this `M` has the stated homotopy-invariant property (P), but is not cofibrant.

This is not merely an issue with allowing the trivial Hopf algebra. With any finite nonsemisimple `H`, trivial action on the same `A`, and `M = A/(t)`, the relative contractible module `M ⊗ H` is zero in `C(A,H)` but is not projective over `A`. The surjective quasi-isomorphism `A ⊗ H → M ⊗ H` cannot admit the lift of the identity that cofibrancy would require.

### Repair and acceptance tests

Use the **actual cellular-filtration condition**, up to `B`-module isomorphism, for the theorem that property-(P) modules are strictly cofibrant, and for the statement that cofibrant modules are their `B`-module retracts. The existing `PropertyPData` is a useful carrier for precisely this stronger condition.

The homotopy-invariant saturation can still be retained as a separate full subcategory of `C(A,H)`. Its objects have cellular/cofibrant **representatives**, and it is appropriate for the derived-category equivalence and homotopy-Hom comparison. It must not imply strict projectivity or strict lifting for every representative. Alternatively, a homotopical lifting condition must explicitly permit lifting up to homotopy and be distinguished from the strict condition currently stated.

Please include both `H=k, M=A/(t)` and the contractible summand `M⊗H` as negative tests for strict cofibrancy. The positive test should verify strict cofibrancy of the actual bar/cellular replacement, not infer it merely from an isomorphism in `C(A,H)`.

For source precision: [Qi, arXiv:1205.1814, Definition 6.3](https://arxiv.org/pdf/1205.1814) does use the homotopy-category wording, while Definition 6.1 is the strict lifting condition. The point is not that the roadmap misquoted that version; the elementary examples above show that these literal formulations cannot support the claimed representative-level implication without the distinction just described. A formalization needs to resolve it rather than inherit it as an unprovable target.

kim-em:
:robot: Claude+Codex say:

Thanks for both rounds. Changes are in ed83a2b and the two follow-up commits, on top of the repaired #221 branch; the categorical targets consume its `StableCategory`, `FrobeniusExactData`, Happel triangulation, and the Euler-forms `TriangulatedK0` directly. Point by point:

First review, the missing categorical outputs:

- `C(A,H)`: `SmashModule` is a category (`B`-linear maps, abelian) with `relativeExactStructure`, whose conflations are characterized as the `A`-split short exact sequences, and `relativeExactStructure_frobenius`. `RelativeStable S` is the dependency's stable category of it; `relativeStable_rel_iff_isIntegralNull` identifies its null ideal with the integral ideal for the canonical `H`-action on `Hom_A` (now constructed once as `homActionFormula`, not taken as an argument), and `isRelativeProjective_iff_retract_tensorH` identifies the relative projectives.
- `D(A,H)`: `restrictStable : C(A,H) → H-StMod` is induced by the ideal-quotient universal property and its factorization through the two quotients is a proof; it is stated to be a triangle functor. `relativeQuasiIso` is invertibility of its image, `RelativeDerived S` the localization, `RelativeVerdierDerived S` the Verdier quotient by `hAcyclicProperty`, `relativeDerivedComparison_isEquivalence` relates them, and `D(A,H)` carries the triangulated instances. The `H-StMod` action on `B-Mod`, `C(A,H)`, and `D(A,H)` is stated as Mathlib's `MonoidalRightAction` (with associator and unitor), not as a bare bifunctor, with exactness at module level and triangle-functor targets for the action functors on `C(A,H)` and on `D(A,H)`.
- Property (P) and replacements: see the second round below; the bar replacement is a functor with a natural transformation to the identity whose components are surjective quasi-isomorphisms, and the literal full subcategories of `C(A,H)` on cellular and on cofibrant representatives map by equivalences onto `D(A,H)`.
- `H-StMod` and `H-stmod`: `HStMod` is the stable category of the abelian Frobenius structure on all `H`-modules, with Happel's triangulation, a `MonoidalCategory` target, and tensoring on either side stated to be a triangle functor; `Hstmod` is the full subcategory on objects isomorphic to a finite module, made monoidal through `ObjectProperty.IsMonoidal` (with finite-dimensionality of `H` an explicit hypothesis of that instance), with tensor-left and tensor-right triangle-functor targets.
- Grothendieck structures: `K0Hstmod` is `TriangulatedK0` with a ring structure characterized by addition, unit, and class-of-tensor theorems; `K0Hopfological` is `K₀` of the compact full subcategory, a module over `K0Hstmod`ᵐᵒᵖ (a right module, since the ring may be noncommutative), with the scalar action of classes the class of the action and the action shown to preserve compacts. Compact generation is against a finite skeleton of simple modules (`SimpleSkeleton`), with the Ravenel–Neeman thick-closure statement.
- Laugwitz–Qi: `HnGradedModule` (finite-dimensional graded modules), `HnGradedStable` as the dependency's stable category, `lqIdealProperty` (the image of `I`, with the pieces now finite-dimensional and the representative's carrier pinned to the literal direct sum), thickness and the tensor-ideal theorem, `LQVerdierQuotient` as the `trW` localization with a monoidal localization functor, and `laugwitzQi_5_15` as a ring isomorphism `K₀(O_n) ≃+* ℤ[ν,ν⁻¹]/(Φ_n)` normalized so that the class of `k{1}` is `ν`, with the ring structure on `K₀(O_n)` tied to the tensor product by class theorems. The stable relation is stated the same way, with the internal grading shift defined literally on carriers and the trivial module `k` pinned by its action and grading.
- Braided Hopf foundation: `GradedVect G k` is a category with a `MonoidalCategory` target and, per bicharacter, a `BraidedCategory` target; `BraidedHopfObject.hopfObj` exhibits the record as a Mathlib `HopfObj` in it, with multiplication, comultiplication, unit, counit, and antipode recovered on pure tensors; `ofHopfObj` goes back, and both round trips are stated (`ofHopfObj_hopfObj` as an equality, `hopfObj_ofHopfObj` as a `HEq` across the underlying-space equality).

Second review, property (P) versus cofibrancy: agreed, and the correction is in. The README defines cellular property (P) (the strict filtration, `IsCellular`) separately from Qi's homotopy-invariant property (P) (`HasPropertyP`), draws every strict conclusion from the cellular condition (`isCofibrant_of_isCellular`, `isCofibrant_iff_retract_cellular` with an honest `B`-linear split pair, `isCofibrant_of_retract`), states only `hasPropertyP_of_isCofibrant` in the other direction, and notes that the isomorphism closures of the cellular and cofibrant subcategories coincide (so the retract distinction lives among strict `B`-modules). Negative tests: `dualNumbersResidue_hasPropertyP` with `dualNumbersResidue_not_isCofibrant` for `H = k`, and `tensorH_dualNumbersResidue_isRelativeProjective`, `_hasPropertyP`, and `_not_isCofibrant` for every nonzero finite-dimensional `H` with the trivial action (the obstruction is `A`-nonprojectivity of `A/(t) ⊗ H`, so no semisimplicity hypothesis is needed). The residue module's carrier is the literal quotient, the trivial module-algebra action is pinned to the counit, and so is the `H`-action on the residue module. The positive test is `barReplacement_isCofibrant`, proved from the cellular filtration.

Also changed while checking this: `H-Mod` carriers live in the universe of `k` and `H`, smash-module carriers in the universe of `k`, `H`, and `A` together, and the Laugwitz–Qi carriers in the universe of `k`, so the units, regular modules, `A ⊗ V`, and `N ⊗ H` are objects of the categories being quantified over; and the README no longer leaves the internal tensor product of hopfological modules conditional (it is required for commutative `A` and cocommutative `H`).

The PR description is rewritten to describe the PR as it stands.

CBirkbeck (commented):

**Assessment: the earlier categorical and strict-cofibrancy objections are substantially resolved. One finiteness statement still needs correction before approval.** Keep this PR draft/stacked until #221 is integrated; this is a COMMENT review, not an approval.

I checked the revised categorical targets, not only the response to the previous reviews. The relative Frobenius quotient and derived localization now consume #221; strict cellularity, strict cofibrancy and homotopy-invariant property (P) are separate; the actual bar replacement is cellular; and the two dual-numbers counterexamples are retained. The categorical graded-vector-space/HopfObj comparison and the concrete Laugwitz–Qi stable category, ideal, Verdier quotient and normalized K₀ ring theorem are now exposed. The earlier absence-of-targets objection should not be repeated against this head.

### Blocking: finite A-length does not imply the proposed derived Hom is finite over k

Layer 6, item 3, asks for

`Dᶜ(A,H) × Dᶠ(A,H) → Dᶠ(k,H)`

under the Noetherian hypotheses of Layer 5. But `Dᶠ(A,H)` was defined using **finite length over A**, not finite dimension over k. Even Artinian A does not suffice for this assertion.

Take `k=𝔽₂`, the ordinary Hopf algebra `H=k[d]/(d²)` with primitive d, and `A=k(t)` with trivial H-action. Set `P=M=A`, with d acting by zero. Then:

* A is a Noetherian Artinian k-algebra and M has A-length one.
* P is the cellular generator `A⊗k`, hence cofibrant and compact in `D(A,H)`.
* Consequently the proposed derived Hom is represented by `Hom_A(A,A)=A`, with trivial H-action.
* This is not isomorphic in H-StMod to a finite-dimensional H-module. Indeed `stableHom_H(k,A)=A`: ordinary Hom is A and the integral d acts by zero, so no nonzero map is killed. This is infinite-dimensional over k, whereas `stableHom_H(k,N)` is finite-dimensional for every finite-dimensional N.

Thus the value does not lie in the announced target `Dᶠ(k,H)`. This is a counterexample to the finiteness conclusion, not to existence of derived Hom in the large stable category.

**Repair:** initially state this pairing for A finite-dimensional over k, or explicitly assume that the finite-length A-modules under consideration are finite-dimensional over k. Alternatively retain a pairing into the large H-StMod and separately prove the finite-target theorem. Keep `Dᶠ`'s finite-length convention consistent throughout rather than silently changing it. Qi §9.1 prints the broad pairing too; the example above is why the formal roadmap should impose the needed finiteness rather than inherit that literal wording.

Two related precision improvements:

1. Pin “smooth basic” to **Qi's split definition**: every simple A-module is one-dimensional over k, equivalently `A/J(A) ≅ k^r` (Definition 9.3). Over a non-algebraically-closed field, the ordinary weaker use of “basic” permits nontrivial division algebras and does not justify `Hom_A(P_i,S_i)=k` in the K₀ calculation.
2. In the Laugwitz–Qi section, call the descended structure the **monoidal tensor product defined using the ambient braiding**, not a braiding on the H_n-module/stable category. The paper explicitly distinguishes these; for general n it does not supply braiding axioms for the module category. The actual target only needs monoidality.

Sources checked: Qi, https://arxiv.org/html/1205.1814, §§6–9, especially Definitions 9.1/9.3 and the displayed §9.1 pairing; Laugwitz–Qi, https://arxiv.org/abs/1804.01478, Definition 4.5, Proposition 4.16, Theorem 5.15 and the warning in §1.4. I have not run a local Lean build. Apart from the finiteness correction above, the substantial revision addresses the previous blockers.

CBirkbeck (changes requested):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `f6b55b5591fd6194d449c413e34f35bea14e8eb5`.

The earlier objections are resolved at this head: the actual stable and relative categories, derived localization, strict cofibrancy/cellularity versus homotopy-invariant property (P), the bar replacement, and the Laugwitz–Qi quotient and normalized K₀ theorem are all represented. The finite-target pairing is now correctly restricted; the `k(t)` counterexample is retained instead of silently changing the meaning of finite length. I also checked Mathlib's `triangEnvelope` at the pinned `2b1308ee...`: it does include retracts, so it is the right thick envelope here. The concrete Laugwitz–Qi cells and prime-power exception agree with Definition 4.5.

**One remaining carrier-universe error needs repair.** `GradedVect` permits `Carrier : Type x` independently of the universe of `k`, but `GradedVect.unitEquiv` unconditionally identifies the monoidal unit's carrier with `k`. This would make every field small in every selected carrier universe. Taking `x = 0` and a field that is not small in universe zero exposes the contradiction. The later `HMod`/smash-module universe repairs do not repair this earlier graded category.

Choose the graded carrier universe to contain the coefficient-field universe (for example `max u x` for `k : Type u`), or carry an explicit `Small.{x} k` hypothesis through the monoidal construction and the HopfObj bridge. Check the monoidal unit and both HopfObj round trips with explicit, unequal universe levels. The theorem bodies being `sorry` makes this easy to miss in a successful elaboration.

Two smaller implementation notes: the `tensorH` docstring omits `a` in `(a#h)·(n⊗l)`; the first factor should be `a·(h₁·n)`. Also, retain arbitrary H-modules as the primary cell parameter: the prose's equivalence with sums of indecomposables should not be justified by finite-dimensional Krull–Schmidt for arbitrary large modules. This last wording occurs in Qi too, so it is not a misquotation; any literal decomposition claim needs its own hypotheses or argument.

This PR is still draft, and I have left that status unchanged. No local Lean build was run. The new request is about the universe of the categorical unit, not a repetition of the already addressed categorical/cofibrancy/finite-pairing objections.

PR #211, opened 2026-08-13
Add roadmap: zigzag, preprojective, and Ginzburg algebras

kim-em:
> **Stacked-PR safety:** this roadmap is complete and ready for review. It is marked draft only because this repository automatically enables merge on every non-draft PR, while this PR temporarily targets an unprotected prerequisite branch. Keep it draft until its prerequisites merge; then retarget it to `main` and mark it ready.

## Summary

Adds a literature roadmap for zigzag, skew-zigzag, additive preprojective, and Ginzburg DG algebras, including:

- doubled-graph quotients with explicit A1/A2 and disconnected-component conventions;
- bases, Frobenius traces, centers, projectives, and graded Cartan matrices;
- signed versus signless preprojective presentations and quadratic/Koszul duality;
- the 2-dimensional Ginzburg bigrading, derived Koszul duality, and published A-infinity deformation results;
- braid bimodule complexes and precisely scoped spherical-twist comparisons;
- compiled D4, E8, and affine E8 edge/dimension acceptance checks.

The accompanying `Suggested.lean` imports current Tau Ceti path-algebra and Dynkin APIs and prototypes the load-bearing quotient, grading, low-rank, Cartan/Hom, orientation, and named-example interfaces.

## Dependencies

This roadmap consumes #201 and #220. PR #211 is stacked on the #220 feature branch, whose history includes #201, so its checkout and CI contain both imported interfaces while its review diff remains limited to zigzag, preprojective, and Ginzburg algebras. After those prerequisites merge, retarget this PR to `main` and mark it ready; review-driven changes will be rippled forward.

## Sources and scope

Primary anchors include Huerfano–Khovanov, Couture, Ehrig–Tubbenhauer, Etingof–Eu, Keller, Etgü–Lekili, Liu–Wang, Khovanov–Seidel, and Seidel–Thomas. The roadmap deliberately excludes the affine-E8 finiteness conjecture, a categorical null-root quotient, a selected Hopf action, the exterior skew-group construction itself, and any lattice categorification claim.

## Verification

- `lake exe cache get`
- `lake env lean TauCetiRoadmap/ZigzagPreprojective/Suggested.lean`
- `lake build` — succeeds on the exact stacked checkout (8742 jobs)
- `git diff --check`

The `sorry` declarations are intentional roadmap targets.

## Review provenance

[...]

PR #223, opened 2026-08-13
Add roadmap: algebraic McKay correspondence and skew group algebras

kim-em:
> **Stacked-PR safety:** this roadmap is complete and ready for review. It is marked draft only because this repository automatically enables merge on every non-draft PR, while this PR temporarily targets an unprotected prerequisite branch. Keep it draft until its prerequisites merge; then retarget it to `main` and mark it ready.

## Summary

Adds a literature roadmap for the algebraic McKay correspondence: explicit cyclic and binary polyhedral subgroups of `SU(2)`, irreducibles and McKay matrices, named affine ADE graphs and the regular dimension vector `δ`, the algebraic affine-to-finite root-lattice passage, polynomial/exterior skew group algebras, McKay-quiver corners and relative Koszul duality, affine zigzag/preprojective comparisons, deformed preprojective algebras, and the algebraic curved-duplex/Weyl-kernel constructions of Frenkel–Khovanov–Schiffmann.

`Suggested.lean` exercises explicit matrix/quaternion data, McKay matrices, the affine-E8 vector `(6,3,4,2,5,4,3,2,1)`, its Cartan-kernel and order-120 checks, the pinned left skew multiplication, and algebraic deformed-preprojective/curved targets.

## Dependencies

This roadmap consumes #201, #220, #221, and #211. Its temporary base, `integration/mckay-prereqs`, is a tested merge of the #221 and #211 prerequisite branch tips; it is not a PR to `main` and contains no McKay changes. This gives PR #223 a checkout and CI environment containing every imported `Suggested.lean` interface while keeping the review diff limited to McKay and skew-group material. After the prerequisites merge, retarget this PR to `main` and mark it ready; public-name changes will be rippled forward.

## Sources and scope

Primary anchors include Conway–Smith and Choi–Lee for explicit subgroups; McKay and Steinberg; Huerfano–Khovanov; Crawley-Boevey–Holland and Crawley-Boevey–Kimura; and Frenkel–Khovanov–Schiffmann. It deliberately makes no exhaustiveness claim for finite subgroups and excludes GIT, Kleinian resolutions, Nakajima varieties, geometric derived McKay, a categorical quotient by `δ`, E8 homological finiteness, and lattice categorification.

## Verification

- `lake exe cache get`
- `lake env lean TauCetiRoadmap/McKaySkewGroup/Suggested.lean`
- integrated `lake build` on the exact prerequisite merge — succeeds (8750 jobs)
- `git diff --check`

The `sorry` declarations are intentional roadmap targets.

## Review provenance and current state

The complete draft was written by `gpt-5.6-sol` and independently red-teamed by a fresh `gpt-5.6-sol` reviewer. A bounded Sol correction removed the false generic affine-kernel claim, rebuilt the corner and relative-Koszul interfaces around actual constructions, pinned the CBH trace/sign normalization, and corrected the FKS reflection kernel to `C_{a,-x_a}` with concrete stable functor targets. Direct Lean and the full integrated build pass; the PR is ready for human review.

This is not a claim that a human subject expert has reviewed the mathematics. Review will be especially valuable from a representation theorist familiar with McKay quivers/skew group algebras/preprojective algebras, an FKS specialist, and Lean contributors familiar with finite representations and quiver quotients.

Authored with Codex.


CBirkbeck (changes requested):
This needs major correction. There is one concrete mathematical contradiction and a larger FKS architecture problem.

1. The CBH parameter normalization is inconsistent inside the roadmap. Layer 5 correctly says that lambda_i is the scalar by which z acts on rho_i, equivalently Tr_{rho_i}(z)/dim(rho_i), and explicitly warns that the trace is not the weight. But completion criterion 6 says lambda_i = Tr_{rho_i}(z), and the comment before cbhCornerEquivDeformedPreprojective repeats that wording. Suggested.lean's actual cbhVertexWeight definition divides by dimension and is the correct one. Fix the completion criterion and the misleading comment. This matters precisely for the binary dihedral and E6/E7/E8 cases, where irreducibles have dimension greater than one.

2. FKSComplexHomotopyCategory is not constructed. It is definitionally just FKSObject, followed by arbitrary sorry-provided Category and Preadditive instances. There is no morphism carrier of closed degree-zero maps, no null-homotopy relation or morphism ideal, and no quotient. Consequently the later stable category is not known to be the FKS homotopy/stable category at all.

3. The reflection kernel is a bimodule duplex, but the stable category is built from FKSComplex objects rather than FKSDuplex objects. The source's Weyl kernels act on the duplex/2-periodic theory. The roadmap needs either to build the category of duplexes and act there, or prove a folding/equivalence theorem from the complex model to the duplex model before defining the tensor functor.

4. FKSOrbitStableCategory and fksOrbitReflectionFunctor are opaque sorry definitions. No Grothendieck construction, fiber inclusions, parameter transport, or comparison with the concrete tensor kernel is stated. Braid isomorphisms between arbitrary endofunctors of an arbitrary category do not validate the FKS theorem. Pin the orbit category from the actual curvature fibers and require the reflection functor to restrict to fksReflectionTensorFunctor on each fiber.

5. The roadmap says the stable quotient reuses the sibling project's projective-factorization ideal, but the source stable category and its duplex-homotopy model need an explicit equivalence. At present there is an unjustified double abstraction: an arbitrary homotopy category followed by a projective quotient.

6. The handedness of relative quadratic duality is still not finally fixed. The README says the answer is Lambda(V*) smash Gamma "or the precisely stated opposite algebra required by the chosen convention", while Suggested.lean fixes a direct equivalence without an opposite. The normative roadmap should choose the convention now and expose the explicit opposite-algebra bridge if that is what the source requires.

The explicit subgroup and McKay-table architecture is otherwise much improved, and the genericity restriction on the Weyl action is appropriately cautious. But the false lambda normalization and the underconstructed FKS categories are merge blockers.

Review was by ChatGPT 5.6 pro

CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `85c8953e178b7850ed5acfc94af0b38d675de8fa`, the complete README and `Suggested.lean`, both integration changes, and the previous review. This is a mathematical/source-level review, not a local Lean build.

**An important correction to the previous ChatGPT review:** its endorsement of the scalar CBH parameter was wrong. The trace normalization in the completion criterion is the correct one for the standard deformed preprojective relation. Please do not implement the previous request to replace it by the scalar normalization.

### 1. Restore the trace parameter in the standard CBH comparison

**Locations:** README Layer 5; `cbhVertexWeight`, its explanatory comments, and `cbhCornerEquivDeformedPreprojective`.

Write `f_i` for the scalar by which the central element `z` acts on `ρ_i`, and `δ_i = dim ρ_i`. For the standard relation

`Σ_a [a,a*] = Σ_i λ_i e_i`,

the CBH parameter is

`λ_i = Tr(z | ρ_i) = δ_i f_i`,

not `f_i`. This is explicit in Crawley-Boevey's subsequent paper, immediately before Theorem 0.9, and in Tikaradze's account of the CBH equivalence, §2, printed p. 4. The primitive-corner identity `e_i z e_i = f_i e_i` is true, but does not say that the corner of the commutator is the already-normalized preprojective relation. The relation comparison contributes the missing dimension factor.

A useful test uses `Γ = Q₈`, whose affine `D₄` graph has a dimension-one leaf and a dimension-two centre. Give the corresponding two blocks of `z` scalar eigenvalues `1` and `−1`. The standard preprojective algebra with those **scalar** vertex weights has a representation supported with dimension one at each of these two vertices: take the two opposite-arrow maps to be nonzero scalars with the product giving the two signed vertex relations. If the advertised vertex-compatible corner equivalence existed with those weights, its inverse would give a CBH module whose underlying `Γ`-representation is one copy of each irreducible. But then

`Tr([x,y]) = 0`, while `Tr(z) = 1·1 + 2·(−1) = −1`,

a contradiction. This test distinguishes the normalizations; cyclic examples cannot, because all their `δ_i` equal one.

**Repair:** keep the block-scalar map and `cbhVertexWeight_asAlgebraHom` under a name identifying them as scalars, but define the parameter used in the standard preprojective comparison by the trace. The centre's **algebra** equivalence with a product of copies of `ℂ` should continue to use scalars: the trace-coordinate map is linear, not an algebra homomorphism when some `δ_i > 1`. State the normalized relation calculation and the images of the vertex idempotents explicitly.

Sources: [Crawley-Boevey, *Preprojective algebras, differential operators and a Conze embedding*, printed p. 3](https://www.math.uni-bielefeld.de/~wcrawley/preproj2.pdf); [Tikaradze, arXiv:2204.13647, §2](https://arxiv.org/pdf/2204.13647).

### 2. The named McKay graph does not imply determinant one

**Locations:** `NamedMcKayIdentification`, `binaryExteriorCornerEquivZigzag`, `binarySymmetricCornerEquivPreprojective`, and `cbhCornerEquivDeformedPreprojective`.

The shared input records dimension two, faithfulness, self-duality, and the exact affine graph, but not preservation of a nondegenerate alternating form or a trivial determinant. The `IsBinary` condition applies to the **graph tag**, not to the supplied group representation.

The ordinary dihedral group of order eight gives a counterexample to inferring the missing hypothesis. Take its faithful representation generated by

`r = diag(i,−i)`, `s = [[0,1],[1,0]]`.

It is self-dual and has exactly the same affine `D₄` McKay graph and dimensions `(1,1,1,1,2)` as the quaternion group: tensoring its two-dimensional irreducible with any one-dimensional character gives that irreducible, and its tensor square is the sum of the four characters. It therefore fits the current `.binaryDihedral 2` identification. However, `det(s)=−1`.

This changes the actual relations. The top exterior power is the nontrivial determinant character, so the degree-two exterior corner connects a vertex to its determinant twist instead of providing the required loop at every vertex. More decisively, in the CBH target with `z=1`, conjugation by `s` interchanges `x` and `y`. The relation `[x,y]=1` then also gives `−[x,y]=1`, forcing `1=0` over `ℂ`. This is not the standard nonzero affine deformed preprojective algebra.

**Repair:** pass the determinant-one representation, or its invariant nondegenerate alternating form, into the corner and CBH theorems. The explicit quaternion/SU₂ constructions already provide the intended datum; the generic theorem interfaces must retain it. Add the ordinary-dihedral representation as a negative test: it may pass the graph identification but must not pass the SL₂ corner theorem.

### 3. Quotienting the homotopy category by its categorical projective objects collapses it

**Locations:** `fksProjectiveIdeal` and `FKSStableCategory`.

The previous review asked for construction of the homotopy category and clarification of the subsequent projective quotient. There is a definite obstruction to the quotient currently specified: it uses `CategoryTheory.Projective X` **inside the homotopy category**.

In a triangulated category every categorical epimorphism splits. Indeed, completing an epimorphism `f : X → Y` to a triangle forces the next map to be zero; exactness after applying `Hom(Y,−)` then supplies a right inverse to `f`. Therefore every object is projective in this categorical sense. Every identity consequently factors through a projective object—namely itself—and the displayed factor-ideal quotient is the zero category.

Thus implementing `FKSComplexHomotopyCategory` correctly will not make this definition the desired FKS stable category; it will make the unintended collapse explicit.

**Repair:** construct the graded/parity module category over the curved extension algebra used by FKS, quotient by its actual projective modules there, and prove the comparison with the appropriate duplex/homotopy model. Alternatively, identify the intended class to kill in the homotopy category through the actual comparison functor; do not use categorical projectivity in a triangulated category. The source distinguishes these constructions in [FKS, §4](https://arxiv.org/pdf/math/0311485).

The earlier requests for actual morphisms/homotopies, the complex-versus-duplex bridge, and curvature-fibre inclusions in the orbit category remain relevant; they are not replaced by the present correction. Include a known nonzero stable object as a test so that a zero-category model cannot satisfy the acceptance criteria vacuously.

### 4. `rightDual_finiteProjective` drops the finite semisimple base hypothesis

**Location:** `rightDual_finiteProjective` in the relative Koszul section.

The statement accepts an arbitrary `k`-algebra `S`, although the intended relative theory is over the previously defined `SeparableSemisimpleBase`. Finite projectivity of `W` over `S ⊗ Sᵐᵒᵖ` alone does not give the stated finiteness of `Hom_{Sᵐᵒᵖ}(W,S)` as a bimodule.

Take `k=ℚ`, `S=ℚ[t]`, and `W=S⊗_ℚS`, with its outer bimodule actions. Then `W` is free of rank one over `S⊗Sᵐᵒᵖ`. As a right `S`-module it is free on the countably many vectors `t^n⊗1`; each can independently be sent to `0` or `1` by a right-linear map to `S`. Its right dual is therefore uncountable. But `S⊗Sᵐᵒᵖ ≅ ℚ[x,y]` is countable, so every finitely generated module over it is countable. The claimed finite generation is impossible.

**Repair:** carry the intended finite-dimensional separable-semisimple base hypothesis into this theorem, or state the precise weaker hypotheses with a proof. This keeps the group-algebra application intact and prevents a false general-purpose supplier from being exported.

The explicit subgroup and graph calculations remain useful; the requested changes concern information lost between those calculations and the relation/category comparisons.

CBirkbeck (commented):

**Assessment: not yet ready for approval. The CBH, determinant-one and finite-base corrections are present, but the FKS construction still needs substantive repair.** Keep the draft/stacked-PR safety arrangement until the prerequisites merge.

The trace parameter `λ_i = dim(ρ_i)·f_i`, distinct from the block-scalar algebra coordinates, is the right correction. The determinant-one hypothesis is now retained in `NamedMcKayIdentification`, and `rightDual_finiteProjective` now carries `SeparableSemisimpleBase`. Those September 8 requests should not be repeated against this head.

### 1. The stable/homotopy comparison is in the wrong direction

Replacing categorical projectivity by the opaque `FKSInduced` predicate does not yet repair the construction. FKS §4, equation (4.1), gives

`Mod(Ã_c) ≃ Com(A,c) → stableMod(Ã_c) → K(A,c)`,

and the analogous parity statement. It does **not** construct the stable category as a further quotient of the homotopy category. Actual projective `Ã_c`-modules are already zero in the homotopy category.

The distinction is real even in a small example. Put `A=k[t]/(t²)` in degree zero, `c=0`, `M=A/(t)`, and `Ã=A[d]/(d²)` with d of degree one. The induced module `N=Ã⊗_A M` is the contractible two-term disk on M. It is therefore zero in `K(A,0)`. But it is not projective over Ã: restricting any projective Ã-module gives a projective A-module, because Ã is free over A, whereas `N|_A ≅ M⊕M` is not A-projective. Thus N is nonzero in `stableMod(Ã)`.

Consequently a quotient formed **after** passing to K cannot recover this stable category: the displayed nonzero stable object has already been killed. Also distinguish induction from arbitrary A-modules, which produces these relatively contractible objects, from induction of A-projectives, which produces the absolute projectives used by FKS.

**Repair:** construct the graded/parity Ã-module category and its quotient by actual projective modules. Construct the canonical functor from that stable category to the homotopy category. Any claim that it is an equivalence needs additional hypotheses and a proof; it is false in this generality.

### 2. The old concrete-category requirements remain unresolved

In the current `Suggested.lean`, `FKSComplexHomotopyCategory` is still definitionally `FKSObject`, with arbitrary `sorry`-provided Category/Preadditive instances, rather than the specified closed-map/null-homotopy quotient. `FKSInduced` is just another `sorry` predicate, with no curved-extension carrier or comparison functor defining it. The reflection kernel is still a duplex while the target uses the complex-object carrier. `FKSOrbitStableCategory` and `fksOrbitReflectionFunctor` still have no curvature-fibre inclusions or restriction theorem identifying the functor with tensoring by the actual kernel.

Please finish those concrete interfaces using the repaired #221 machinery and the correct stable model above. A `sorry` is appropriate for a difficult theorem in a roadmap; it is not a substitute for specifying which morphisms, quotient, or functor the headline theorem concerns.

### 3. The new nonzero-category test is false at its stated generality

`exists_not_isZero_fksStableCategory` quantifies over every commutative coefficient ring, algebra, grading and curvature. It must instead assert nontriviality for a **specified intended example**. Even a nonzero algebra can have a zero curved homotopy category: take `k=ℚ`, `A=ℚ[u,u⁻¹]`, `deg u=2`, and `c=u`. For every curved object with `d²=u`, the map `h=(1/2)u⁻¹d` has degree −1 and satisfies `dh+hd=1`. Every object is contractible, so the current further quotient of the homotopy category is zero too. FKS §3.2 also records the invertible-curvature obstruction in the duplex setting. Retain this as a negative test alongside a properly scoped positive example.

One earlier specification issue also remains: Layer 3 still says the quadratic dual is `Λ(V*)⋊Γ` “or” an opposite algebra, while the representative equivalence chooses one. Fix the handed convention and its explicit opposite comparison now rather than leave this to the implementor.

Source checked directly: FKS, https://arxiv.org/html/math/0311485, §§3.2–4.2 and equation (4.1). The disk and Laurent-curvature calculations above are direct counterexamples. No local Lean build was run.

CBirkbeck (changes requested):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `2828a22b7ccd1080093a90c7a79e8855b32161f2`.

Still not ready for approval. I read the full current diff and all three previous reviews. The corrected CBH trace parameter, determinant-one hypothesis, and separable-semisimple hypothesis on right-dual finiteness are accepted; the original review's endorsement of scalar CBH weights must not be reinstated.

The September 14 FKS objections remain unresolved at this head. In particular, `FKSStableCategory` is still a further quotient of the homotopy-category carrier. The required direction is the canonical functor from the absolute stable category of curved-extension modules to the homotopy category, not an unconditional equivalence or reverse quotient. The contractible disk on the nonprojective `k[t]/(t²)`-module `k` still distinguishes them. The actual closed-map/null-homotopy quotient, duplex-versus-complex comparison, and curvature-fibre restriction equations for the orbit functor are still missing.

The universal `exists_not_isZero_fksStableCategory` is still false. For the zero algebra every unital module is zero. More substantively, in the current homotopy-based model with `A=ℚ[u,u⁻¹]`, `deg u=2`, `c=u`, the homotopy `h=(1/2)u⁻¹d` contracts every object. Replace this theorem by a specified nonzero intended example, retaining the invertible-curvature example as a negative test.

I found two further concrete convention mismatches while checking the source:

**1. Right-action sign and parity.** In FKS §2.4, the bimodule differential supercommutes with the left action but **commutes with the right action**. Both `supercommutes_right_even` and `supercommutes_right_odd` currently insert `(-1)^p`. With the usual balanced tensor product and total differential, that extra sign prevents compatibility with `(ma)⊗n = m⊗(an)`. Either use the source's right-linearity or define and prove the full twisted-action/tensor comparison; a bare `Aᵐᵒᵖ` action does not supply it. Also expose the shift and action twist in the literal multiplication/comultiplication kernel. FKS §3.1 uses a genuinely parity-graded A-module: odd elements of A exchange its parity components. The current `FKSDuplex` instead gives each component a full A-module structure preserved by every A-action and retains integer gradings. It needs a totalization/folding comparison, not just the label 'parity version'.

**2. Ordinary versus skew zigzag.** The generic FKS functions accept every connected simple graph with more than one vertex but use the ordinary `zigzagAlgebra`. FKS §5.1 uses the orientation-signed algebra; its comparison with the ordinary zigzag algebra is supplied in the bipartite case. Restrict this layer to the intended bipartite ADE cases and transport the trace/volume classes explicitly, or construct the signed algebra and its comparison. The ordinary and signed relations cannot silently be identified on an odd cycle.

Source checked: FKS, arXiv:math/0311485, §§2.4, 3.1, 4, 5.1 and 6; §2.4 was also checked in the rendered PDF. These are specification issues, not requests for proofs of all roadmap targets. Draft/base state is unchanged; no local Lean build was run.

PR #226, opened 2026-08-14
Add roadmap: counting totally ramified extensions of a local field

0stellensatz:
## Summary

This roadmap owns the **counting** of totally ramified extensions of a non-archimedean local
field: the set `σ_K(n)`, the wild exponent `c(L) = d(L) − n + 1` as its weight, the machinery that
evaluates the count, and the two mass formulas of [Serre 1978] with the finiteness, convergence
and orbit statements that accompany them.

```text
∑_{L ∈ σ_K(n)} q^{-c(L)} = n            ∑_{M} 1 / (#Aut_K(M) · q^{c(M)}) = 1
```

The directory is `TauCetiRoadmap/MassFormula/`, renamed from `TotallyRamified` during review: the
old name claimed the arithmetic of total ramification, which this roadmap does not own — #189
does — and `MassFormula` matches the suggested Tau Ceti home. The predicate `IsTotallyRamified`
and the set `totallyRamifiedOfDegree` keep their names, which describe mathematics rather than
this roadmap. The five references to the old directory name in #189's prose are updated here.

The machinery the count needs is not ramification theory, which is what makes it a roadmap rather
than a section of one: a **quantitative Newton estimate** over a complete DVR (Mathlib's
`hensels_lemma` is that estimate over `ℤ_[p]` only, and `HenselianLocalRing` lifts a simple root
with no distance bound, hence no count), the
**lattice-index scaling law** for Haar measure over such a ring (`Measure.addHaar_image_linearMap`
is real-vector-space only), and the parametrization of `σ_K(n)` by the **Eisenstein region** of
coefficient space. The first two are specified over an arbitrary complete DVR and are reusable
well outside this roadmap; they are placed by subject, in
`TauCeti/RingTheory/DiscreteValuationRing/` (the Newton estimate and the uniqueness of the lifted
root), `TauCeti/LinearAlgebra/FreeModule/` (the lattice index, beside Mathlib's `ℤ` analogue
`AddSubgroup.index_eq_natAbs_det`) and `TauCeti/MeasureTheory/Group/` (the Haar scaling law).

The PR contains the normative `README.md` and target signatures in `Suggested.lean`: seven
definitions of its own, two junk-tolerant wrappers around the consumed substrate, twenty-two
milestones as `sorry`-targets, and two closed comparison theorems that type-check the adapter
contract described below.

## Ownership

This roadmap does **not** own the arithmetic of the individual extension. Total ramification,
Eisenstein generators, monogenicity, the different, the discriminant and the tame criterion belong
to **Local Fields and Ramification (#189)**, whose Layer 3 specifies all of them; this roadmap
consumes that contract and is, as far as I can tell, its first counting consumer. The README lists
the consumed items explicitly and says that restating any of them in `TauCeti/` is a duplicate to
be deleted rather than a contribution.

That contract is now consumed **by name**. `Suggested.lean` imports
`TauCetiRoadmap.LocalFieldsRamification.Suggested` and uses #189's declarations directly:
`IsTotallyRamified` with `isTotallyRamified_iff_inertiaDegree_eq_one` and
`isTotallyRamified_iff_exists_eisenstein_generator`, `exists_integerRing_adjoin_eq_top`,
`localDiscriminantIdeal`, `differentExponent` and `discriminantExponent` with
`differentExponent_eq_of_algEquiv` and `discriminantExponent_eq_of_algEquiv`, and
`addVal_sum_eisenstein_powerBasis`. The README carries the full consumed-item → supplier-
declaration table.

Where this roadmap's carrier is an arbitrary `L : IntermediateField K (SeparableClosure K)` rather
than a finite extension as such, Layer 0 defines two junk-tolerant wrappers,
`intermediateFieldIsTotallyRamified` and `intermediateFieldDiscriminantExponent`. Each installs
#189's `finiteIntermediateField*` structure adapters and reduces to the canonical invariant
through a comparison theorem with a **closed proof**, exactly as #189's *Consumer contract:
Counting Totally Ramified Extensions* section prescribes. No ramification-theoretic notion —
index, different, discriminant ideal, or exponent — is defined independently here.

The two items an earlier revision of this description listed as unclaimed by the contract —
invariance of `d` (hence `c`) under `K`-isomorphism, and orthogonality of a power basis at an
Eisenstein generator — are now #189 exports (`discriminantExponent_eq_of_algEquiv` and
`addVal_sum_eisenstein_powerBasis`). This roadmap owns only their family-level consequences:
invariance of `wildExponent`, along which Theorem 2 regroups, and the coordinate-box and
coordinate-cube descriptions that carry the volume computation.

The two roadmaps agree on their shared example: the `d = 2` and `d = 3` values this roadmap's
`ℚ_2`, `n = 2` acceptance test needs are #189's own `ℚ_2(i)` and `ℚ_2(√2)` worked examples.
Concretely, `σ_{ℚ_2}(2)` has six members, two with `c = 1` and four with `c = 2`, and Theorem 1
reads `2·2^{-1} + 4·2^{-2} = 2`. That is the test that catches an off-by-one in `c`, and it also
checks that the consumed `d` and the `c` here are the same invariant.

The supplier chain has landed — #188, #244 and #189 are all merged — and this branch is updated
from `main`, so `build` is green and the relative link to `../LocalFieldsRamification/README.md`
resolves. The red-by-design state this description previously described is over.

## Coordination

I am the author of [0stellensatz/MassFormula](https://github.com/0stellensatz/MassFormula)
(Apache-2.0), where every layer is formalized `sorry`-free, so there is no third-party
coordination outstanding. The roadmap treats it as a cited source rather than a specification: it
states the mathematics intrinsically, hands the ramification half to #189, and asks for the two
general layers at a generality the source did not need. Its provenance section records where the
roadmap departs from the source — including dropping the source's own proof of
`IsAdicComplete 𝓂[K] 𝒪[K]`, which Mathlib now carries as an instance, and dropping the source's
comparator-pair device, which does not transfer to a repository where no `sorry` may land.

## Human review priorities

- **the two Layer 0 wrappers** — whether junk-tolerant totalization over
  `IntermediateField K (SeparableClosure K)` is the right carrier choice, given that the
  alternative is to carry a finiteness hypothesis in every statement downstream;
- the decision to state quantitative Newton lifting over an arbitrary complete DVR and the
  lattice index over an arbitrary DVR, rather than over a local field, and their suggested homes
  outside `NumberTheory/LocalField/`;
- the `ℝ≥0∞` convention for both mass formulas, with the convergence remark kept separate over
  `ℝ` as `Summable`, rather than a single real-valued statement carrying a summability
  hypothesis;
- representative sets as a predicate quantified over, rather than a quotient;
- whether the Eisenstein-region measure `q^{-n}(1 - q^{-1})` and the root-count integral
  `q^{-(d+1)}(1 - q^{-1})` are the right pair of acceptance constants to pin, given that a
  misnormalized Haar measure is otherwise invisible until the summit.

## Checks

`lake build TauCetiRoadmap.MassFormula.Suggested` passes on `lean-toolchain` v4.34.0-rc2 against
the pinned Mathlib, on this branch merged with `main`: rc=0, 8868 jobs, no errors, 22 `sorry`
warnings. The two comparison theorems are closed and do not count as milestones. The count rose
from sixteen in response to review: two targets for the countability and measurability behind
Theorem 1, and four more that the review showed were being assumed in prose rather than stated —
`adjoin_mem_totallyRamifiedOfDegree`, `le_addVal_sum_eisenstein_powerBasis_iff`,
`eq_of_isRoot_of_addVal_lt` and `mem_eisensteinSet_iff`. On GitHub the `build` check is green.

The pin files are untouched, and so are the issue-template `area` dropdowns — the sync bot
regenerates those on merge, so this PR leaves them alone.

## Follow-up

Serre's formula is a *mass* formula and does not by itself count extensions. A follow-up PR will
extend this roadmap to the actual counts: Krasner's count of totally ramified extensions of given
degree and given discriminant, Monge's determination of the number of isomorphism classes, and
Bhargava's mass formula over étale algebras. Kedlaya 2007 and Wood–Yasuda 2015/2017 stay out of
scope and remain named in `## Long horizon`.

## :robot: AI assistance

Drafted with Claude Opus 5, under my direction. I have read the roadmap carefully myself. The
consumption commit is David Roe's, adopted from his suggestion branch with authorship preserved.

### DIFF
--- a/TauCetiRoadmap/TotallyRamified/README.md
+++ b/TauCetiRoadmap/TotallyRamified/README.md
@@ -1,0 +1,353 @@
+# Roadmap: counting totally ramified extensions of a local field, and Serre's mass formula
+
+The [local fields and ramification roadmap](../LocalFieldsRamification/README.md) builds the
+arithmetic of a finite extension of a non-archimedean local field: total ramification, Eisenstein
+generators, monogenicity, the different and the discriminant, and the tame criterion. This
+roadmap is that theory's first counting consumer. It asks how *many* such extensions there are,
+and proves the answer — the **mass formula** of [Serre 1978]: inside a fixed separable closure,
+the totally ramified extensions of degree `n` of a local field `K` satisfy
+
+```text
+∑_{L ∈ σ_K(n)} q^{-c(L)} = n,          c(L) = d(L) − n + 1
+```
+
+where `q` is the residue cardinality and `d(L)` the discriminant exponent, together with its
+isomorphism-class form `∑_{M} 1 / (#Aut_K(M) · q^{c(M)}) = 1`. It is the local counting statement
+behind mass heuristics for global field counting, and nothing like it is upstream.
+
+The machinery the count needs is not ramification theory, and that is what makes this a separate
+roadmap: a **quantitative Newton estimate** over a complete discrete valuation ring (Mathlib's
+`HenselianLocalRing` lifts a simple root but gives no distance bound, hence no count), the
+**lattice-index scaling law** for Haar measure over such a ring (Mathlib's
+`Measure.addHaar_image_linearMap` is real-vector-space only), and the parametrization of `σ_K(n)`
+by the **Eisenstein region** of coefficient space. The first two are stated over an arbitrary
+complete discrete valuation ring and are reusable well outside this roadmap.
+
+Suggested home: `TauCeti/NumberTheory/LocalField/MassFormula/`, with the two general layers
+placed by subject instead: quantitative Newton lifting in
+`TauCeti/RingTheory/DiscreteValuationRing/`, and the lattice scaling law in
+`TauCeti/MeasureTheory/Group/`.
+
+## Scope
+
+This roadmap owns the **counting** of totally ramified extensions of a local field: the set
+`σ_K(n)` as an object, the wild exponent `c` as its weight, the analytic and measure-theoretic
+machinery that evaluates the count, and the mass formulas with the finiteness, convergence and
+orbit statements that accompany them.
+
+It does not own the arithmetic of the individual extension. **Total ramification, Eisenstein
+generators, monogenicity, the different, the discriminant, and the tame criterion belong to the
+[local fields and ramification roadmap](../LocalFieldsRamification/README.md)**, which is where
+they are specified and where their API accrues; the contract this roadmap consumes from it is
+listed below. It also does not own higher ramification groups, Herbrand functions or Hasse–Arf
+(same roadmap); local class field theory; adic or Huber-theoretic geometry over `K`, which is the
+[adic spaces roadmap](../AdicSpaces/README.md); or archimedean local fields, which
+`IsNonarchimedeanLocalField` excludes and which never enter.
+
+## Standing conventions
+
+Pinning these matters more than usual: the objects are junk-tolerant by design, and an
+implementor who picks a different model will not be able to state the summit.
+
+- **The base field.** `K` carries
+  `[Field K] [ValuativeRel K] [UniformSpace K] [IsUniformAddGroup K] [IsNonarchimedeanLocalField K]`.
+  Spell the hypotheses out; do not bundle them. Mathlib's instances then supply
+  `IsDiscreteValuationRing 𝒪[K]`, `Finite 𝓀[K]`, `CompactSpace 𝒪[K]`, `CompleteSpace K` and
+  `IsAdicComplete 𝓂[K] 𝒪[K]` by `inferInstance`; consume them, never re-prove them.
+- **The residue cardinality is `q K := Nat.card 𝓀[K]`**, and the residue characteristic is written
+  inline as `ringChar 𝓀[K]`. Do not introduce a `p` abbreviation.
+- **Extensions live inside a separable closure.** An extension is a term of
+  `IntermediateField K (SeparableClosure K)`. This is Serre's setting and it is load-bearing: in
+  equal characteristic `p` the inseparable Eisenstein polynomials (`X^p − π`) must not be counted,
+  and working inside `SeparableClosure K` excludes them by construction rather than by a side
+  hypothesis. ⚠ The consumed roadmap states its ramification theory for a finite extension `L/K`
+  as such; the bridge to subextensions of `SeparableClosure K` is this roadmap's Layer 0.
+- **The discriminant exponent is the consumed one.** `d L` is the invariant the local fields and
+  ramification roadmap defines; for a totally ramified extension its different exponent `v_L(𝔡)`
+  and the valuation of its discriminant ideal in `K` agree, since the residue degree is `1`. Do
+  not introduce a second discriminant.
+- **Junk is tolerated, `0 < n` is not optional.** For `L` infinite over `K`, or for `n = 0`, these
+  definitions take junk values. Every statement about `σ_K n` therefore carries `0 < n`, and
+  membership in `σ_K n` is what makes a statement honest. Do not add finiteness hypotheses to
+  definitions to avoid junk; add them to theorems.
+- **Sums are in `ℝ≥0∞`.** Theorems 1 and 2 are stated with `∑'` valued in `ℝ≥0∞`, where the
+  possibly infinite sum needs no convergence side condition and equality with `n` (resp. `1`)
+  already encodes convergence. The convergence remark is a *separate* statement over `ℝ` as
+  `Summable`; state both and relate them, and do not merge them.
+- **Representative sets are a predicate, not a quotient.** `IsRepresentativeSet n R` says that
+  `R ⊆ σ_K n` and every `L ∈ σ_K n` is `K`-isomorphic to exactly one member of `R`; Theorem 2
+  quantifies over every such `R`. This is Serre's phrase verbatim and avoids `Quotient.lift`
+  well-definedness obligations for `c` and `#Aut`.
+- **The coefficient space and its measure.** A monic degree-`n` polynomial is its coefficient
+  vector `a : Fin n → K`, with `a i` the coefficient of `X ^ i`. The measure is the additive Haar
+  measure of `Fin n → K` normalized so that the integer box `Fin n → 𝒪[K]` has volume `1`; by
+  uniqueness of Haar measure this is the product of the normalized coordinate measures, but the
+  product structure is needed only inside proofs, so neither `Measure.pi` nor σ-finiteness appears
+  in any statement.
+- **Valuations in an extension are `IsDiscreteValuationRing.addVal`**, `ℕ∞`-valued. No topology on
+  `L` is used anywhere: completeness of the ring of integers of `L` is the algebraic
+  `IsAdicComplete`, and all estimates are `addVal` estimates. This keeps the count free of a
+  second uniform structure.
+
+## What this roadmap consumes
+
+### From the local fields and ramification roadmap
+
+Every item below is a target *there*, not here. This roadmap's milestones rest on them and add no
+independent development of the same material; if one of them is restated in `TauCeti/` by this
+roadmap's implementors, that is a duplicate to be deleted rather than a contribution.
+
+- **Total ramification** as a predicate, with `e · f = [L:K]` and the equivalence with residue
+  degree `1`.
+- **Totally ramified is equivalent to Eisenstein**, in both directions, with the uniformizer the
+  Eisenstein polynomial produces.
+- **Local monogenicity**, and in particular that a root of an Eisenstein polynomial generates the
+  ring of integers on its power basis.
+- **The different and the discriminant**, the exponent `d(L/K)`, and the comparison with Mathlib's
+  `differentIdeal`.
+- **The tame criterion** `d = e − 1 ⟺ tamely ramified`, and the wild lower bound `e ≤ d`. Together
+  these are exactly what makes `c(L) = d(L) − n + 1` a nonnegative integer, and what makes
+  `c(L) = 0` equivalent to `¬ (p ∣ n)`; both are consumed here rather than reproved.
+
+### From Mathlib
+
+- **Local fields:** `Mathlib/NumberTheory/LocalField/Basic.lean` — `IsNonarchimedeanLocalField`
+  and the instance package above, plus `valueGroupWithZeroIsoInt`.
+  ⚠ `IsAdicComplete 𝓂[K] 𝒪[K]` **is an instance there**; it is not a gap and must not be reproved.
+- **Notation and valuations:** `Mathlib/Topology/Algebra/Valued/ValuativeRel.lean`,
+  `Mathlib/RingTheory/Valuation/ValuativeRel/*`, and
+  `Mathlib/RingTheory/DiscreteValuationRing/Basic.lean` (`IsDiscreteValuationRing.addVal`,
+  `exists_irreducible`, `Irreducible.maximalIdeal_eq`).
+- **Eisenstein polynomials:** `Mathlib/RingTheory/Polynomial/Eisenstein/*`
+  (`Polynomial.IsEisensteinAt`, `IsEisensteinAt.irreducible`).
+- **Hensel's lemma, qualitatively:** `Mathlib/RingTheory/Henselian.lean` — `HenselianLocalRing`,
+  `HenselianRing`. This lifts a simple root *modulo the maximal ideal*; it gives no distance
+  estimate and no count, which is why Layer 1 exists.
+- **Separable degree and embeddings:** `Mathlib/FieldTheory/SeparableDegree.lean` —
+  `Field.embEquivOfAdjoinSplits`, `Field.finSepDegree_eq_finrank_of_isSeparable`.
+- **Smith normal form:** `Mathlib/LinearAlgebra/FreeModule/IdealQuotient.lean` and
+  `…/Finite/Quotient.lean` — `Submodule.quotientEquivPiSpan`, the input to the lattice index.
+- **Haar measure:** `Mathlib/MeasureTheory/Group/Measure.lean` — `IsAddHaarMeasure` and the
+  existence of Haar measure on a locally compact group. ⚠ The determinant scaling law
+  `Measure.addHaar_image_linearMap` (`Mathlib/MeasureTheory/Measure/Lebesgue/EqHaar.lean`) is
+  stated for **real** vector spaces only; the non-archimedean analogue is Layer 2.
+
+## What is missing (build here)
+
+The set `σ_K(n)` and the wild exponent as a counting weight, with its invariance under
+`K`-isomorphism; the orthogonality of a power basis at an Eisenstein generator; quantitative
+Newton lifting over a complete discrete valuation ring and the local constancy of root counts it
+yields; the lattice-index formula and the Haar scaling law over such a ring with finite residue
+field; the measure of the Eisenstein region and the parametrization it carries; and the two mass
+formulas with their finiteness, convergence and orbit-counting companions. None of this is
+upstream, and none of it is claimed by the roadmap this one consumes.
+
+---
+
+## The build, in layers
+
+The ordering is the dependency order. As each layer makes the next layer's *types* expressible in
+`TauCeti/`, its milestones go into `Suggested.lean` (with `sorry`).
+
+### Layer 0: the counting invariants
+
+- **`σ_K(n)`**, the set of subextensions of `SeparableClosure K` that are totally ramified of
+  degree `n` over `K`, built on the consumed total-ramification predicate. With the bridge lemmas
+  the count needs: `σ_K(1) = {K}`; membership is preserved by the image of any `K`-embedding into
+  `SeparableClosure K`; and every member is `K⟮x⟯` for `x` a root of an Eisenstein polynomial of
+  degree `n`, which is the consumed equivalence transported to this setting.
+- **The wild exponent `c L := d L + 1 − n`** in truncated `ℕ`-subtraction, where `d` is the
+  consumed discriminant exponent. The consumed bounds make the truncation faithful, so `c` is
+  Serre's nonnegative integer and `c L = 0 ⟺ ¬ (ringChar 𝓀[K] ∣ n)`; state both here as the
+  corollaries of the consumed facts that they are, not as fresh developments.
+- **Invariance of `c` under `K`-isomorphism.** A `K`-isomorphism carries integral bases to
+  integral bases with the same discriminant, so `d`, hence `c`, is an invariant of the class.
+  ⚠ Not covered by the consumed contract, which states `d` for a fixed extension; Theorem 2 is
+  exactly a regrouping along this invariance, so it is a target here.
+- **Orthogonality of the power basis.** In the ring of integers generated by an Eisenstein
+  generator `ξ` of degree `n`, the terms of `∑_{i<n} c_i ξ^i` have pairwise distinct valuations —
+  the `i`-th is `n · addVal(c_i) + i`, distinct modulo `n` — so no cancellation is possible and
+  the valuation of the sum is the minimum of the terms'. Consequences: a ball of the ring of
+  integers is a box in power-basis coordinates, and at a radius divisible by `n` it is a cube.
+  ⚠ Also not covered by the consumed contract, which gives the monogenic basis but not the
+  valuations of a combination in it. This is the workhorse of Layers 2 and 3 and deserves its own
+  name and API.
+
+### Layer 1: quantitative Newton lifting over a complete discrete valuation ring
+
+Stated for an arbitrary complete (`IsAdicComplete`) discrete valuation ring, since nothing here
+uses the local field; the local-field case is an instance.
+
+- **Newton iteration with an estimate.** If `addVal (F y₀) > 2 · addVal (F' y₀)`, the iteration
+  `y ↦ y − F y / F' y` converges to a root `z` of `F` with
+  `addVal (z − y₀) ≥ addVal (F y₀) − addVal (F' y₀)`, and `z` is the unique root in that ball.
+  ⚠ This strictly refines Mathlib's `HenselianLocalRing`, which lifts a simple root modulo the
+  maximal ideal and yields no distance bound; the estimate, not the existence, is what Layer 3
+  consumes.
+- **Local constancy of the root count.** For a monic `f` over `𝒪[K]`, separable over `K`, there is
+  a threshold `T` such that every monic `g` with `g.coeff i − f.coeff i ∈ 𝓂[K]^T` for all `i` has
+  exactly as many roots in `L` as `f` does, for every `L ∈ σ_K n`. The route is a Bézout
+  certificate `U · f + V · f' = β` over `𝒪[K]` bounding the order of `f'` at every near-root
+  uniformly, then Newton lifting in both directions.
+- ⚠ **Not Krasner's lemma.** Mathlib's Krasner (`Analysis/Normed/Field/Krasner.lean`) is a
+  normed-field statement about a *single* root generating an extension; what Layer 3 needs is a
+  *count* that is uniform on a coefficient ball, which Krasner does not give. Keep the two apart
+  and do not route this layer through the normed setting.
+
+### Layer 2: lattices, index, and the Haar scaling law
+
+Stated for a complete discrete valuation ring with finite residue field of cardinality `q`, for
+the same reason.
+
+- **The index of an image lattice.** For a matrix `M` over `𝒪[K]` with `Associated M.det (π^k)`,
+  the quotient of the integer box `Fin n → 𝒪[K]` by its image under `M` has exactly `q^k`
+  elements. Via `Submodule.quotientEquivPiSpan` (Smith normal form over the principal ideal ring
+  `𝒪[K]`), the quotient splits into residue rings of the diagonal coefficients, whose product is
+  associated to `M.det`.
+- **The scaling law.** For the Haar measure of `Fin n → K` normalized on the integer box,
+  `μ (M · S) = q^{−k} · μ S` for `S` the box or any ball — the non-archimedean analogue of
+  `Measure.addHaar_image_linearMap`. ⚠ Prove it through lattice indices, not through uniqueness
+  of Haar measure: the index route needs no second-countability or regularity side conditions, and
+  the general-set form is not needed.
+- **Boxes and balls.** The image lattice of a diagonal matrix is the coordinate box with `i`-th
+  factor `π^{e i} · 𝒪[K]`, of volume `q^{−∑ e i}`; a ball is a translate of the lattice of a
+  constant-radius box. With Layer 0's orthogonality this computes the volume of a ball of the ring
+  of integers of `L` in power-basis coordinates, which is the only other volume the summit needs.
+
+### Layer 3: the Eisenstein region and the parametrization
+
+- **The Eisenstein region** `E_n ⊆ (Fin n → K)`: every coefficient in `𝓂[K]`, with the constant
+  term of valuation exactly that of a uniformizer. Its measure is `q^{−n} · (1 − q^{−1})`.
+- **Almost every Eisenstein polynomial is separable** ([Serre 1978, eq. (3)]): the non-separable
+  locus is null. In equal characteristic this is where the inseparable polynomials are discarded,
+  and it is the reason the count is over subextensions of `SeparableClosure K`.
+- **The root count.** For `L ∈ σ_K n`, `rootCount L a` is the number of roots of the monic
+  polynomial with coefficient vector `a` lying in `L`, counted with multiplicity. On the
+  full-measure separable locus all multiplicities are `1`. Almost everywhere on `E_n`,
+  `∑_{L ∈ σ_K n} rootCount L a = n`: an Eisenstein polynomial is irreducible, and each of its `n`
+  roots generates exactly one member of `σ_K n`.
+- **The local fibre count** ([Serre 1978, Lemma 1]): the cube of radius `π^ρ` around an Eisenstein
+  generator `ξ` contains exactly one root of each nearby Eisenstein polynomial — Layer 1 applied
+  on the region.
+- **The integral of the root count** ([Serre 1978, Lemmas 2 and 3, eqs. (5)–(13)]): for each
+  `L ∈ σ_K n`, `∫_{E_n} rootCount L = q^{−(d L + 1)} · (1 − q^{−1})`. The change of variables runs
+  in power-basis coordinates: a ball of the ring of integers is the ideal generated by a suitable
+  element, in coordinates the image lattice of the matrix of multiplication by it, whose
+  determinant is that element's norm — so Layer 2 supplies the factor `q^{−d L}` with no Jacobian,
+  no conjugates and no Vandermonde.
+
+### Layer 4: the mass formulas
+
+- **Theorem 1** ([Serre 1978, Thm. 1]): for `0 < n`, `∑' L : σ_K n, 1 / (q K) ^ c L = n` in
+  `ℝ≥0∞`. Integrate the root-count identity of Layer 3 over `E_n` and divide by the measure of the
+  region.
+- **The finiteness dichotomy** ([Serre 1978, Rmk. 1°]): `σ_K n` is infinite if and only if `K` has
+  equal characteristic `p` and `p ∣ n`. Both directions are targets. The forward direction follows
+  from Theorem 1 together with the uniform bound `c L ≤ n · addVal(n)` outside the asserted case;
+  the converse is an explicit family — the Eisenstein polynomials `X^n + π^m X + π` for `m ≥ 1`
+  have `d = n · m`, pairwise distinct, hence generate infinitely many distinct members.
+- **Convergence** ([Serre 1978, Rmk. 1°]): `Summable fun L : σ_K n => 1 / (q K : ℝ) ^ c L`, the
+  real-valued restatement, meaningful precisely in the infinite case.
+- **The orbit count** ([Serre 1978, Rmk. 3°]): for `L ∈ σ_K n`, the number of `M ∈ σ_K n` that are
+  `K`-isomorphic to `L`, times `#Aut_K(L)`, is `n`. `L / K` is separable of degree `n`, so it has
+  exactly `n` embeddings into `SeparableClosure K`; each image lies in `σ_K n` by Layer 0; and the
+  embeddings with a given image form a torsor under `Aut_K(L)`.
+- **Theorem 2** ([Serre 1978, Thm. 2]): for every `R` with `IsRepresentativeSet n R`,
+  `∑' M : R, 1 / (#Aut_K(M) · (q K) ^ c M) = 1` in `ℝ≥0∞`. Theorem 1 regrouped along isomorphism
+  classes, using the invariance of `c` from Layer 0 and the orbit count.
+- **The tame count**, as a corollary worth stating: when `¬ (ringChar 𝓀[K] ∣ n)`, every `c L = 0`,
+  so `σ_K n` is finite with exactly `n` elements.
+
+## Worked examples (acceptance criteria, keeping the definitions honest)
+
+Discharge these alongside the layers; they catch a vacuous `σ`, a wrong normalization of the
+measure, and an off-by-one in `c`.
+
+- **`n = 1`:** `σ_K 1 = {K}` with `d = c = 0`, and Theorem 1 reads `1 = 1`.
+- **Tame:** for `K = ℚ_p` with `p` odd and `n = 2`, `σ_K 2` has exactly two elements (`ℚ_p(√p)`
+  and `ℚ_p(√(u p))` for `u` a non-residue unit), both with `c = 0`, so the sum is `2`. This is the
+  tame count above in its smallest instance.
+- **Wild, and the sharpest test of `c`:** for `K = ℚ_2` and `n = 2`, `σ_K 2` has six elements —
+  `ℚ_2(√−1)` and `ℚ_2(√−5)` with `d = 2`, hence `c = 1`; and `ℚ_2(√±2)`, `ℚ_2(√±10)` with `d = 3`,
+  hence `c = 2`. Theorem 1 reads `2 · 2^{−1} + 4 · 2^{−2} = 2`. An implementation that gets `c`
+  off by one fails here and nowhere earlier. ⚠ Those two discriminant values are the `ℚ_2(i)` and
+  `ℚ_2(√2)` cases of the consumed roadmap's own worked examples; agreeing with them is the check
+  that the consumed `d` and the `c` here are the same invariant.
+- **Equal characteristic:** for `K = 𝔽_p((t))` and `n = p`, `σ_K n` is infinite, yet the real
+  series converges — the two halves of Remark 1° on the same example.
+- **The measure of the Eisenstein region** is `q^{−n}(1 − q^{−1})`, and the integral of
+  `rootCount L` over it is `q^{−(d L + 1)}(1 − q^{−1})`; summing over `L` recovers `n` times the
+  measure of the region. Checking the two constants against each other catches a misnormalized
+  Haar measure, which is otherwise invisible until the summit.
+- **Galois case of the orbit count:** for `L / K` Galois in `σ_K n`, `#Aut_K(L) = n` and the
+  isomorphism class of `L` inside `σ_K n` is `{L}`.
+
+## Ordering
+
+Layer 0 comes first and rests on the consumed contract. Layers 1 and 2 are independent of Layer 0
+and of each other — quantitative Newton lifting and the lattice scaling law are general statements
+about complete discrete valuation rings — so all three are parallel lanes, and the two general
+ones can be built before any of the consumed material lands. Layer 3 needs 0–2; Layer 4 needs
+Layer 3, except for the orbit count, which needs only Layer 0 and can be built early.
+
+## Long horizon (a roadmap for a future roadmap; do not attempt it here)
+
+Serre's formula was refined by Krasner into a count of the totally ramified extensions of given
+degree *and* given discriminant, and generalized by Bhargava into a mass formula over all étale
+extensions of a local field, which is the local input to the conjectured densities of number-field
+discriminants. Both are natural continuations and neither is specified here; they are named to
+mark the direction, not to be worked on under this roadmap.
+
+## References
+
+- J-P. Serre, *Une «formule de masse» pour les extensions totalement ramifiées de degré donné
+  d'un corps local*, C. R. Acad. Sci. Paris **286** (1978), Série A, 1031–1036 — the summit:
+  Thm. 1, Thm. 2, Remarks 1° and 3°, Lemmas 1–3 and the change of variables of §3.
+- J-P. Serre, *Local Fields*, Graduate Texts in Mathematics **67**, Springer (1979) — Chap. I §6
+  (Eisenstein polynomials and monogenicity, Prop. 17), Chap. III §3 (the different and the
+  discriminant), Chap. III §6 (the valuation of the different, tameness, Prop. 13). Consumed
+  through the local fields and ramification roadmap rather than developed here.
+- M. Krasner, *Nombre des extensions d'un degré donné d'un corps p-adique*, in *Les tendances
+  géométriques en algèbre et théorie des nombres*, CNRS (1966), 143–169 — the
+  discriminant-refined count that Serre's formula summarizes.
+- M. Bhargava, *Mass formulae for extensions of local fields, and conjectures on the density of
+  number field discriminants*, IMRN (2007) — the étale generalization.
+
+## Acknowledgements
+
+The mathematics of every layer has been formalized, `sorry`-free, in
+[0stellensatz/MassFormula](https://github.com/0stellensatz/MassFormula) (Apache-2.0), by the
+proposer of this roadmap; the provenance map below records where each layer's proof can be read.
+That project is a **cited source, not the specification**: this roadmap states the mathematics
+intrinsically, hands the ramification-theoretic half to the roadmap that owns it, and asks for the
+two general layers at a generality the source did not need.
+
+### Provenance (secondary; the layers above are definitive)
+
+File map, relative to that project's `MassFormula/`:
+
+| Layer | Files |
+| --- | --- |
+| 0 | `Defs.lean`, `UniformizerParam.lean` (orthogonality), `Second.lean` (invariance of `c`) |
+| 1 | `RootLifting.lean` |
+| 2 | `HaarScaling.lean` |
+| 3 | `First.lean`, `UniformizerParam.lean` (the fibre count) |
+| 4 | `First.lean`, `Second.lean`, `Finiteness.lean`, `Convergence.lean`, `Orbit.lean` |
+
+The source's `EisensteinMonogenic.lean`, `Discriminant.lean` and `Tame.lean` prove material this
+roadmap consumes rather than owns; they are evidence that the consumed contract is provable in
+this setting, and belong with the roadmap that specifies it, not here.
+
+Where this roadmap departs from the source:
+
+- The source proves `IsAdicComplete 𝓂[K] 𝒪[K]` itself; that instance is now in Mathlib and is to
+  be consumed, not carried.
+- The source wraps `integralClosure ↥𝒪[K] ↥L` and its maximal ideal in project abbreviations; use
+  Mathlib's spelling, and the consumed roadmap's, so the lemmas land where the rest of the
+  integral-closure API lives.
+- Layers 1 and 2 are stated in the source only for the local-field case they were needed in; this
+  roadmap asks for them over an arbitrary complete discrete valuation ring (with finite residue
+  field, for Layer 2), which is the generality their proofs already have.
+- The source is organized as a frozen specification file paired with a development file that
+  discharges it. That device does not transfer: in Tau Ceti the roadmap is what commits to a
+  statement before its proof exists, and no `sorry` may land in `TauCeti/`.
--- a/TauCetiRoadmap/TotallyRamified/README.md
+++ b/TauCetiRoadmap/TotallyRamified/README.md
@@ -61,10 +61,13 @@ implementor who picks a different model will not be able to state the summit.
   equal characteristic `p` the inseparable Eisenstein polynomials (`X^p − π`) must not be counted,
   and working inside `SeparableClosure K` excludes them by construction rather than by a side
   hypothesis. ⚠ The consumed roadmap states its ramification theory for a finite extension `L/K`
-  as such; the bridge to subextensions of `SeparableClosure K` is this roadmap's Layer 0.
-- **The discriminant exponent is the consumed one.** `d L` is the invariant the local fields and
-  ramification roadmap defines; for a totally ramified extension its different exponent `v_L(𝔡)`
-  and the valuation of its discriminant ideal in `K` agree, since the residue degree is `1`. Do
+  as such; the bridge to subextensions of `SeparableClosure K` is this roadmap's Layer 0, which
+  installs the consumed `finiteIntermediateField*` adapters.
+- **The discriminant exponent is the consumed one.** `d L` is the consumed
+  `discriminantExponent`, totalized over arbitrary subextensions by the wrapper
+  `intermediateFieldDiscriminantExponent`; for a totally ramified extension its different
+  exponent `v_L(𝔡)` and the valuation of its discriminant ideal in `K` agree, since the residue
+  degree is `1` (the consumed `discriminantExponent_eq_inertiaDegree_mul_differentExponent`). Do
   not introduce a second discriminant.
 - **Junk is tolerated, `0 < n` is not optional.** For `L` infinite over `K`, or for `n = 0`, these
   definitions take junk values. Every statement about `σ_K n` therefore carries `0 < n`, and
@@ -85,17 +88,21 @@ implementor who picks a different model will not be able to state the summit.
   product structure is needed only inside proofs, so neither `Measure.pi` nor σ-finiteness appears
   in any statement.
 - **Valuations in an extension are `IsDiscreteValuationRing.addVal`**, `ℕ∞`-valued. No topology on
-  `L` is used anywhere: completeness of the ring of integers of `L` is the algebraic
-  `IsAdicComplete`, and all estimates are `addVal` estimates. This keeps the count free of a
-  second uniform structure.
+  `L` enters any counting statement or estimate: completeness of the ring of integers of `L` is
+  the algebraic `IsAdicComplete`, and all estimates are `addVal` estimates — the spelling the
+  consumed `addVal_sum_eisenstein_powerBasis` exports. The consumed structure adapters put the
+  canonical topology and valuation on a subextension carrier only inside the Layer 0 wrappers;
+  no statement here hypothesizes a second uniform structure.
 
 ## What this roadmap consumes
 
 ### From the local fields and ramification roadmap
 
-Every item below is a target *there*, not here. This roadmap's milestones rest on them and add no
-independent development of the same material; if one of them is restated in `TauCeti/` by this
-roadmap's implementors, that is a duplicate to be deleted rather than a contribution.
+Every item below is specified there and is imported here from that roadmap's `Suggested.lean` by
+exact name; the table at the end of this section is the map. This roadmap's milestones rest on
+them and add no independent development of the same material; if one of them is restated in
+`TauCeti/` by this roadmap's implementors, that is a duplicate to be deleted rather than a
+contribution.
 
 - **Total ramification** as a predicate, with `e · f = [L:K]` and the equivalence with residue
   degree `1`.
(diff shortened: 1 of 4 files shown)


roed314:
The mass formula doesn't actually count the extensions; this would be a nice roadmap to also implement the formulas in
> Maurizio Monge, Determination of the number of isomorphism classes of extensions of a p-adic field, J. Number Theory 131 (2011), no. 8, 1429–1434.

roed314:
This comment was mostly written by GPT 5.6, but I edited it a bit.

Here are some thoughts on the division between #189 and #226.  I think there is a clean one-way boundary:

> **#189 owns the arithmetic of one finite local extension.  
> #226 owns the construction and counting of families of those extensions.**

So the dependency should be:

```text
Local Fields and Ramification (#189) → Counting Totally Ramified Extensions (#226)
```

with no dependency in the other direction.

### What #189 should own

For a finite extension `L/K`, #189 owns:

- the canonical induced norm/topology/`ValuativeRel`/local-field structure, including its
  `finiteIntermediateField*` adapters;
- `ramificationIndex`, `inertiaDegree`, `IsTotallyRamified`, and the tame/wild predicates;
- the total-ramification/Eisenstein equivalence and local monogenicity;
- `differentExponent`, `localDiscriminantIdeal`, and `discriminantExponent`;
- the tame equality and wild bounds;
- `differentExponent_eq_of_algEquiv` and `discriminantExponent_eq_of_algEquiv`;
- `addVal_sum_eisenstein_powerBasis`.

The last two bullets are the two points on which the current #226 description is now stale. They
are reusable arithmetic statements about an individual extension, so they belong in #189:

```text
#189: discriminantExponent_eq_of_algEquiv
#226: wildExponent_eq_of_algEquiv, derived from it

#189: addVal_sum_eisenstein_powerBasis
#226: coordinate-box/cube and measure consequences, derived from it
```

### What #226 should own

#226 should own:

- `σ_K(n)` / `totallyRamifiedOfDegree`;
- `IsRepresentativeSet`;
- the derived mass-formula weight `wildExponent L = d(L) + 1 - [L:K]`;
- invariance of that weight;
- quantitative Newton lifting and local constancy of root counts;
- lattice-index and nonarchimedean Haar-scaling results;
- the coefficient space and Eisenstein region;
- root-count integrals, finiteness, orbit counting, and both mass formulas.

### Intermediate-field wrappers

There is one legitimate adapter layer in #226. Its objects are arbitrary terms

```lean
L : IntermediateField K (SeparableClosure K),
```

whereas #189's canonical invariants require a finite local extension.

#226 may therefore define total or junk-tolerant wrappers, but they should:

1. be named as intermediate-field adapters rather than as a second canonical local API;
2. install #189's `finiteIntermediateFieldNormedField`,
   `finiteIntermediateFieldValuativeRel`, and `finiteIntermediateFieldTopology`;
3. come with comparison theorems to #189's `IsTotallyRamified` and
   `discriminantExponent` on finite members of `σ_K(n)`;
4. contain no independent definition of the different, discriminant, ramification index, or
   total ramification.

After #189 lands, I suggest replacing the present consumed stand-ins as follows:

| Current #226 item | Final treatment |
|---|---|
| `maximalIdealAbove` | delete unless needed internally for an adapter proof |
| `ramificationIdx` | use an intermediate-field wrapper around #189's `ramificationIndex` |
| `IsTotallyRamified` | use/wrap #189's `LocalFieldsRamification.IsTotallyRamified` |
| `discriminantIdeal` | delete |
| `discExponent` | rename as an intermediate-field wrapper reducing to #189's `discriminantExponent` |
| `discExponent_eq_of_algEquiv` | delete; consume #189's theorem |
| power-basis orthogonality | delete as a target; consume `addVal_sum_eisenstein_powerBasis` |
| `wildExponent` | keep; this is the counting weight owned here |

The README and PR body should also remove the paragraph saying that discriminant invariance and power-basis orthogonality are absent from #189's contract: the current #189 head now explicitly exports both.

roed-math:
:robot: **Claude Opus 5**, posting on David Roe's behalf.

The [#189](https://github.com/TauCetiProject/TauCetiRoadmap/pull/189) side of the boundary between these two roadmaps is final, so here is the consumption change on this side, prepared as a **one-commit suggestion branch** you can adopt or adapt rather than a push to your branch:

**Branch:** [`roed-math:pr226-consume-189`](https://github.com/roed-math/TauCetiRoadmap/tree/pr226-consume-189) — one commit, [`1b593c5`](https://github.com/roed-math/TauCetiRoadmap/commit/1b593c5af6d862eaab682cb1a39139fb80e9db6e), on top of your current head `a1c0b56`, touching only `TauCetiRoadmap/TotallyRamified/`. Adopt with

```bash
git pull https://github.com/roed-math/TauCetiRoadmap.git pr226-consume-189
```

**What it does** — exactly what your own file already promises ("when that roadmap lands they are deleted here and its definitions are used instead"): adds `import TauCetiRoadmap.LocalFieldsRamification.Suggested` and replaces each *consumed*-marked local shape by the supplier's exact declaration. The mapping is recorded as a table in the README; the highlights:

- `IsTotallyRamified`, `isTotallyRamified_iff_inertiaDegree_eq_one`, `isTotallyRamified_iff_exists_eisenstein_generator`, `localDiscriminantIdeal`, `discriminantExponent`, `differentExponent_eq_of_algEquiv` / `discriminantExponent_eq_of_algEquiv`, `addVal_sum_eisenstein_powerBasis`, `exists_integerRing_adjoin_eq_top` — all consumed by name;
- for your `IntermediateField K (SeparableClosure K)` carrier, two junk-tolerant wrappers (`intermediateFieldIsTotallyRamified`, `intermediateFieldDiscriminantExponent`) install #189's `finiteIntermediateField*` adapters, exactly as #189's *Consumer contract: Counting Totally Ramified Extensions* permits — and **both required comparison theorems with the canonical invariants are closed proofs**, so the adapter contract is type-checked rather than promised;
- everything this roadmap owns is untouched: `σ_K(n)`, `wildExponent` as the derived weight, `IsRepresentativeSet`, the Newton/root-counting, coefficient-space/Haar and mass-formula milestones.

**Verification:** `lake build TauCetiRoadmap.TotallyRamified.Suggested` is green (`rc=0`, 8705 jobs) in a local stack with #189 (and its supplier #244) merged in. One consequence to expect: once the import is in, this PR's GitHub `build` check goes red on the missing supplier until #244 → #189 land and the branch is updated from `main` — the same red-by-design state every consumer PR in the portfolio is in, and the merge order the reviews ask for.

If you adopt it, the PR description's ownership sentence should say the same thing #189's now does: #189 owns the arithmetic of one finite extension; this roadmap owns the family construction and counting; the dependency is one-way, #189 → #226. And if you'd rather draw any part of the boundary differently, say so here or on #189 — the supplier side moves too, nothing here is meant to pre-empt your design.

0stellensatz:
:robot: Drafted with Claude Opus 5; I have read it and it says what I mean.

Thanks — all three taken.

## The suggestion branch (adopted)

Pulled at `1b593c5`, your authorship preserved. I checked it against #189's *Consumer
contract* clause by clause before adopting: it installs the `finiteIntermediateField*` adapters,
both comparison theorems are closed, and no ramification-theoretic notion is defined on this side
any more. All twenty declarations it consumes are present in #189's head.

I reproduced your build rather than take it on trust — `main` + #244 + #189 + this branch:

```
lake build TauCetiRoadmap.TotallyRamified.Suggested
Build completed successfully (8710 jobs).
rc=0 — no errors; 16 `sorry` warnings in TotallyRamified/Suggested.lean
```

The milestone count is unchanged at sixteen, since the two comparison theorems are closed.

One note on phrasing rather than on code: those proofs are closed, but the structures they are
stated at come through `finiteIntermediateField_isNonarchimedeanLocalField`, which is itself still
a `sorry`-target in #189. So what is type-checked is the *shape* of the adapter contract, not the
mathematics under it. That is exactly right for a roadmap — I mention it only so that neither of
us later reads "type-checked rather than promised" as more than it is.

## The boundary

Agreed, no counter-proposal. The one-way `#189 → #226` split is what the PR body now states, in
the same terms as your contract section. The stale paragraph claiming that discriminant invariance
and power-basis orthogonality were absent from your contract is gone; this roadmap now claims only
their family-level consequences — invariance of `wildExponent`, and the box and cube corollaries.

## Monge, and counting proper

You are right that the mass formula does not count anything, and I would like to go further than
the suggestion. A follow-up PR will add these as **targets**, not as long-horizon pointers:

- **Krasner's** count of totally ramified extensions of given degree *and* given discriminant —
  Monge's formulas take it as input, and it is currently sitting in this roadmap's
  `## Long horizon`;
- **Monge 2011**, the determination of the number of isomorphism classes;
- **Bhargava's** mass formula over étale algebras.

Kedlaya 2007 and Wood–Yasuda 2015/2017 stay out, and remain named in `## Long horizon` as
direction only.

Two things I would value your read on before I write it, since both touch your side:

1. **Bhargava needs unramified extensions as objects.** #189's Layer 2 exports `frobeniusAlgEquiv`
   and `normGroup`, but no `IsUnramified` predicate. My plan is to spell it
   `inertiaDegree K L = Module.finrank K L` using your `inertiaDegree` — consuming rather than
   redefining, so the contract holds — but if you would rather own the predicate, say so and I
   will consume it instead.
2. **The product decomposition looks like it comes straight from Mathlib**
   (`Algebra.Etale.iff_exists_algEquiv_prod`, `Mathlib/RingTheory/Etale/Field.lean`), so I do not
   think that half needs anything from #189.

Keeping that as a second PR so this one stays a small diff.

Back to `awaiting-review`.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+  in power-basis coordinates: a ball of the ring of integers is the ideal generated by a suitable
+  element, in coordinates the image lattice of the matrix of multiplication by it, whose
+  determinant is that element's norm — so Layer 2 supplies the factor `q^{−d L}` with no Jacobian,
+  no conjugates and no Vandermonde.
+
+### Layer 4: the mass formulas
+
+- **Theorem 1** ([Serre 1978, Thm. 1]): for `0 < n`, `∑' L : σ_K n, 1 / (q K) ^ c L = n` in
+  `ℝ≥0∞`. Integrate the root-count identity of Layer 3 over `E_n` and divide by the measure of the
+  region.
```
kim-em, line 284: :robot: Claude+Codex say:

**Theorem 1 needs countability of `σ_K(n)` and measurability of `rootCount`, and neither is a target.**

This derivation integrates `tsum_rootCount` over `E_n` and sums `lintegral_rootCount` over `L`. That interchange is `lintegral_tsum`, which needs a countable index type and almost-everywhere measurable summands. `totallyRamifiedOfDegree K n` is an arbitrary set of intermediate fields; nothing in the roadmap says it is countable, and `rootCount L` is a bare function `(Fin n → K) → ℕ` with no measurability statement. Both are genuine theorems: countability comes from the fact that each member owns a piece of `E_n` of positive volume (by local constancy at the coefficient vector of an Eisenstein generator) and finitely many such pieces total at most `n · μ(E_n)`; measurability comes from local constancy too. The cited source has both as named theorems (`countable_sigma`, `aemeasurable_rootCount` in `First.lean`). State `(totallyRamifiedOfDegree K n).Countable` and `AEMeasurable (fun a => (rootCount L a : ℝ≥0∞))` as Layer 3 targets, with the routes above.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+  adapters and carries the comparison theorem the consumer contract requires. With the bridge
+  lemmas the count needs: `σ_K(1) = {K}`; membership is preserved by the image of any
+  `K`-embedding into `SeparableClosure K`; and every member is `K⟮x⟯` for `x` a root of an
+  Eisenstein polynomial of degree `n`, which is the consumed
+  `isTotallyRamified_iff_exists_eisenstein_generator` transported to this setting.
+- **The wild exponent `c L := d L + 1 − n`** in truncated `ℕ`-subtraction, where `d` is the
+  consumed `discriminantExponent` through its wrapper `intermediateFieldDiscriminantExponent`.
+  The consumed bounds make the truncation faithful, so `c` is Serre's nonnegative integer and
+  `c L = 0 ⟺ ¬ (ringChar 𝓀[K] ∣ n)`; state both here as the corollaries of the consumed facts
+  that they are, not as fresh developments.
```
kim-em, line 202: :robot: Claude+Codex say:

**The equal-characteristic wild lower bound is cited to a supplier theorem that does not give it.**

This bullet (and L117, "consumed here rather than reproved") tells implementors to derive `n − 1 ≤ d` and `c = 0 ⟺ ¬ (p ∣ n)` from `differentExponent_eq_ramificationIndex_sub_one_iff` and `differentExponent_bounds_of_wild`. The second one takes

```lean
(he : (ramificationIndex K L : L) ≠ 0)
```

and `IsWildlyRamified K L` is `ringChar 𝓀[K] ∣ ramificationIndex K L`. In equal characteristic `p`, wild means `p ∣ e`, so `(e : L) = 0` and the theorem says nothing. That is the case this roadmap is most interested in (`σ_K n` infinite). So there `n − 1 ≤ d` is not a corollary of the two theorems named, and neither is the `→` direction of `c = 0 ⟺ ¬ (p ∣ n)`.

The bound itself is available: Mathlib's `pow_sub_one_dvd_differentIdeal` (`Mathlib/RingTheory/DedekindDomain/Different.lean`) gives `P ^ (e − 1) ∣ 𝔇` unconditionally for a separable extension, which with the supplier's definition of `differentExponent` and `discriminantExponent_eq_inertiaDegree_mul_differentExponent` is exactly `n − 1 ≤ d`. And the supplier's README (its Layer 3 different item) asks for `e ≤ d` in the wild case with no side condition; only its Lean signature bundles the lower bound with `he`. Fix: cite the Mathlib lemma as the source of the lower bound here and at L117, and drop the claim that it is a corollary of `differentExponent_bounds_of_wild`. A one-line split of that supplier signature would be welcome but is not this PR's job.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+separable over `K`, coefficientwise closeness forces equality of root counts in every member of
+`totallyRamifiedOfDegree K n`. -/
+theorem card_aroots_eq (n : ℕ) (hn : 0 < n) (L : IntermediateField K (SeparableClosure K))
+    (hL : L ∈ totallyRamifiedOfDegree K n) {f : Polynomial ↥𝒪[K]} (hfm : f.Monic)
+    (hfsep : (f.map (algebraMap ↥𝒪[K] K)).Separable) :
+    ∃ T : ℕ, 0 < T ∧ ∀ g : Polynomial ↥𝒪[K], g.Monic →
+      (∀ i, g.coeff i - f.coeff i ∈ 𝓂[K] ^ T) →
+      ((g.map (algebraMap ↥𝒪[K] K)).aroots ↥L).card =
+        ((f.map (algebraMap ↥𝒪[K] K)).aroots ↥L).card :=
+  sorry
```
kim-em, line 271: :robot: Claude+Codex say:

**README and Lean disagree on quantifier order and on generality.**

README L228–232 says one threshold `T` works "for every `L ∈ σ_K n`". Here `∃ T` comes after `L`, so `T` may depend on `L`. The Bézout route the README describes gives the uniform version (the bound `v_L(β) = n · v_K(β)` depends only on `n`), and the uniform version is what the README promises. State it as `∃ T, 0 < T ∧ ∀ L ∈ totallyRamifiedOfDegree K n, ∀ g, …`.

Separately, README L219 and the header above (L48–51) say Layer 1 is "stated for an arbitrary complete discrete valuation ring", and README L328–330 says Layers 1 and 2 are independent of Layer 0. This theorem is stated over a local field `K` and quantifies over Layer 0's `totallyRamifiedOfDegree K n`. Either give it a general core (a monic separable `f` over a complete DVR `A`, root counts in a finite extension with a fixed ramification index) and derive the local-field form as a corollary, or narrow the generality claim and the ordering to `exists_isRoot` alone.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+them and add no independent development of the same material; if one of them is restated in
+`TauCeti/` by this roadmap's implementors, that is a duplicate to be deleted rather than a
+contribution.
+
+- **Total ramification** as a predicate, with `e · f = [L:K]` and the equivalence with residue
+  degree `1`.
+- **Totally ramified is equivalent to Eisenstein**, in both directions, with the uniformizer the
+  Eisenstein polynomial produces.
+- **Local monogenicity**, and in particular that a root of an Eisenstein polynomial generates the
+  ring of integers on its power basis.
```
kim-em, line 112: :robot: Claude+Codex say:

**The Eisenstein-to-membership direction is used everywhere, and its consumed input is misattributed.**

`tsum_rootCount` needs that each root `x` of a degree-`n` Eisenstein polynomial gives `K⟮x⟯ ∈ totallyRamifiedOfDegree K n`; the orbit count needs the same for the image of a `K`-embedding; L194 lists "membership is preserved by the image of any `K`-embedding" as a Layer 0 lemma. Only the forward direction, `exists_eisenstein_generator`, is in `Suggested.lean`.

The `←` direction of the consumed `isTotallyRamified_iff_exists_eisenstein_generator` takes `Algebra.adjoin 𝒪[K] {ξ} = ⊤` as an input. Knowing that `ξ` is an Eisenstein root and generates the field does not give that; it is Serre LF I §6 Prop. 17, "an Eisenstein root generates the ring of integers". This bullet and the table row at L134 say that fact is consumed as `exists_integerRing_adjoin_eq_top`, but that theorem produces *some* generator, not this one. Cite the supplier's README item (its Layer 3 "Eisenstein polynomial generates a totally ramified extension in which its root is a uniformizer") rather than a declaration that does not say it, and state the transported converse here as a Layer 0 target.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+  locus is null. In equal characteristic this is where the inseparable polynomials are discarded,
+  and it is the reason the count is over subextensions of `SeparableClosure K`.
+- **The root count.** For `L ∈ σ_K n`, `rootCount L a` is the number of roots of the monic
+  polynomial with coefficient vector `a` lying in `L`, counted with multiplicity. On the
+  full-measure separable locus all multiplicities are `1`. Almost everywhere on `E_n`,
+  `∑_{L ∈ σ_K n} rootCount L a = n`: an Eisenstein polynomial is irreducible, and each of its `n`
+  roots generates exactly one member of `σ_K n`.
+- **The local fibre count** ([Serre 1978, Lemma 1]): the cube of radius `π^ρ` around an Eisenstein
+  generator `ξ` contains exactly one root of each nearby Eisenstein polynomial — Layer 1 applied
+  on the region.
```
kim-em, line 272: :robot: Claude+Codex say:

**The local fibre count leaves `ρ` and "nearby" unspecified, and as written it is false for small `ρ`.** For `K = ℚ_2`, `ξ = √2`, both roots of `X² − 2` lie in `ξ + 2𝒪[L]`, so `ρ = 1` gives two roots in the cube. The condition is `nρ ≥ d + n` or similar, and it is exactly what determines the lattice whose volume Layer 3 computes. Pin it, with `d` identified as the consumed discriminant exponent.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+  the wrapper to `wildExponent`, junk case included, which is the regrouping Theorem 2 runs
+  along.
+- **The box and cube corollaries of the power-basis orthogonality.** The consumed
+  `addVal_sum_eisenstein_powerBasis` gives, at an Eisenstein generator `ξ` of degree `n`, that
+  the terms of `∑_{i<n} c_i ξ^i` have pairwise distinct valuations — the `i`-th is
+  `n · addVal(c_i) + i`, distinct modulo `n` — so no cancellation is possible and the valuation
+  of the sum is the minimum of the terms'. The targets here are its family-level consequences: a
+  ball of the ring of integers is a box in power-basis coordinates, and at a radius divisible by
+  `n` it is a cube. These are the workhorse of Layers 2 and 3.
+
```
kim-em, line 216: :robot: Claude+Codex say:

The box and cube descriptions are prose only, and the one thing a reader has to reconstruct is that the `i`-th factor of the box for the ball `v_L ≥ r` is `π ^ ⌈(r − i)/n⌉ · 𝒪[K]`. Write that (a Lean form would be better).

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+theorem exists_isRoot {A : Type*} [CommRing A] [IsDomain A] [IsDiscreteValuationRing A]
+    [IsAdicComplete (IsLocalRing.maximalIdeal A) A] (F : Polynomial A) (y₀ : A)
+    (hlt : IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ (Polynomial.derivative F)) +
+        IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ (Polynomial.derivative F)) <
+      IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ F)) :
+    ∃ z : A, F.IsRoot z ∧
+      IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ F) ≤
+        IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ (Polynomial.derivative F)) +
+          IsDiscreteValuationRing.addVal A (z - y₀) :=
+  sorry
```
kim-em, line 259: :robot: Claude+Codex say:

**Cite Mathlib's quantitative Hensel, and state uniqueness in Lean.**

The header (L18–19) and README L20 say Mathlib has "no quantitative Newton estimate (only qualitative Hensel)". `Mathlib/NumberTheory/Padics/Hensel.lean` has `hensels_lemma`: under `‖F.aeval a‖ < ‖F.derivative.aeval a‖ ^ 2` it gives a root `z` with `‖z − a‖ < ‖F'(a)‖`, `‖F'(z)‖ = ‖F'(a)‖`, and uniqueness in that ball, over `ℤ_[p]` (and it proves the exact ratio `‖z − a‖ = ‖F(a)‖ / ‖F'(a)‖` privately). Layer 1 is that theorem generalized to a complete DVR in `addVal` spelling, which is a fair thing to ask for; say so and cite it instead of claiming Mathlib has nothing. Uniqueness is a prose target (README L224) and a docstring aside above, but has no Lean form; add it next to this theorem, since it is what local constancy uses.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+- **The index of an image lattice.** For a matrix `M` over `𝒪[K]` with `Associated M.det (π^k)`,
+  the quotient of the integer box `Fin n → 𝒪[K]` by its image under `M` has exactly `q^k`
+  elements. Via `Submodule.quotientEquivPiSpan` (Smith normal form over the principal ideal ring
+  `𝒪[K]`), the quotient splits into residue rings of the diagonal coefficients, whose product is
+  associated to `M.det`.
+- **The scaling law.** For the Haar measure of `Fin n → K` normalized on the integer box,
+  `μ (M · S) = q^{−k} · μ S` for `S` the box or any ball — the non-archimedean analogue of
+  `Measure.addHaar_image_linearMap`. ⚠ Prove it through lattice indices, not through uniqueness
+  of Haar measure: the index route needs no second-countability or regularity side conditions, and
+  the general-set form is not needed.
```
kim-em, line 252: :robot: Claude+Codex say:

**Layer 2 stops at the box.** The reusable statement, and the honest analogue of the `Measure.addHaar_image_linearMap` cited here, is `μ (M '' S) = q ^ (−k) * μ S` for every measurable `S`. The index route gives it on the box; `μ ∘ (M '' ·)` is again a Haar measure, so the general case is uniqueness of Haar measure with the constant read off the box, at the price of naming the topological facts about `K` that uniqueness needs. This is a should, not a must, but the box-only form is a thin API for a layer the README advertises as "reusable well outside this roadmap". Also cite `AddSubgroup.index_eq_natAbs_det` (`Mathlib/LinearAlgebra/FreeModule/Finite/CardQuotient.lean`) as the `ℤ` analogue of `card_quotient_range`.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
+# Roadmap: counting totally ramified extensions of a local field, and Serre's mass formula
+
+The [local fields and ramification roadmap](../LocalFieldsRamification/README.md) builds the
+arithmetic of a finite extension of a non-archimedean local field: total ramification, Eisenstein
+generators, monogenicity, the different and the discriminant, and the tame criterion. This
+roadmap is that theory's first counting consumer. It asks how *many* such extensions there are,
```
kim-em, line 6: :robot: Claude+Codex say:

Timeless: drop "first".

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+  Haar measure, which is otherwise invisible until the summit.
+- **Galois case of the orbit count:** for `L / K` Galois in `σ_K n`, `#Aut_K(L) = n` and the
+  isomorphism class of `L` inside `σ_K n` is `{L}`.
+
+## Ordering
+
+Layer 0 comes first and rests on the consumed contract. Layers 1 and 2 are independent of Layer 0
+and of each other — quantitative Newton lifting and the lattice scaling law are general statements
+about complete discrete valuation rings — so all three are parallel lanes, and the two general
+ones can be built before any of the consumed material lands. Layer 3 needs 0–2; Layer 4 needs
```
kim-em, line 331: :robot: Claude+Codex say:

Timeless: "can be built before any of the consumed material lands" is timing relative to another roadmap's implementation. Delete.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+      LocalFieldsRamification.finiteIntermediateField_isNonarchimedeanLocalField K
+        (SeparableClosure K) L
+    haveI : ValuativeExtension K ↥L :=
+      LocalFieldsRamification.finiteIntermediateField_valuativeExtension K (SeparableClosure K) L
+    LocalFieldsRamification.IsTotallyRamified K ↥L
+
+omit [IsUniformAddGroup K] in
+/-- The comparison theorem the consumer contract requires: on a finite subextension the wrapper
+is exactly the consumed predicate at the consumed adapter structures. A closed proof, so the
+contract is type-checked rather than promised. -/
```
kim-em, line 103: :robot: Claude+Codex say:

Process narrative (here and at L154–156): "type-checked rather than promised". As noted on the thread, what is checked is the shape of the adapter contract, not the mathematics. Say what the theorem states and stop.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+  sorry
+
+/-- **The finiteness dichotomy** ([Serre 1978, Rmk. 1°]): infinite exactly in equal characteristic
+`p` with `p ∣ n`. -/
+theorem totallyRamifiedOfDegree_infinite_iff (n : ℕ) (hn : 0 < n) :
+    (totallyRamifiedOfDegree K n).Infinite ↔ ringChar K = ringChar 𝓀[K] ∧ ringChar 𝓀[K] ∣ n :=
+  sorry
+
+/-- **Convergence** ([Serre 1978, Rmk. 1°]): the real-valued restatement, meaningful precisely in
+the infinite case, where the `ℝ≥0∞` sum asserts nothing about summability. -/
```
kim-em, line 366: :robot: Claude+Codex say:

This is wrong: equality of the `ℝ≥0∞` sum with a finite value already gives real summability (`ENNReal.summable_toReal`), which is what README L76–79 says. Describe this as the corollary of Theorem 1 that it is.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+discriminant exponent are `LocalFieldsRamification.IsTotallyRamified` and
+`LocalFieldsRamification.discriminantExponent`. The wrappers
+`intermediateFieldIsTotallyRamified` and `intermediateFieldDiscriminantExponent` totalize them
+over arbitrary subextensions by installing the consumed `finiteIntermediateField*` adapters, and
+each carries the comparison theorem the consumer contract requires. `residueCard`,
+`totallyRamifiedOfDegree`, `wildExponent` and `IsRepresentativeSet` are this roadmap's own. -/
+
+/-- The residue cardinality of a non-archimedean local field — Serre's `q`. -/
+noncomputable def residueCard : ℕ :=
+  Nat.card 𝓀[K]
```
kim-em, line 75: :robot: Claude+Codex say:

`def` wrapping `Nat.card 𝓀[K]`. The supplier writes `Nat.card 𝓀[K]` inline throughout, and README L58 bans a `p` abbreviation on the same grounds. Use `Nat.card 𝓀[K]`, or at least `abbrev`.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+are deliberately not restated here. -/
+
+/-- The **bridge**, and the target that turns a statement about `totallyRamifiedOfDegree K n` into
+a computation: the consumed equivalence `isTotallyRamified_iff_exists_eisenstein_generator`,
+transported into `IntermediateField K (SeparableClosure K)` and packaged with the degree, which is
+the form every later layer applies. -/
+theorem exists_eisenstein_generator (n : ℕ) (hn : 0 < n)
+    (L : IntermediateField K (SeparableClosure K)) (hL : L ∈ totallyRamifiedOfDegree K n) :
+    ∃ (x : SeparableClosure K) (π : ↥𝒪[K]), Irreducible π ∧ IsIntegral 𝒪[K] x ∧
+      (minpoly 𝒪[K] x).IsEisensteinAt (Submodule.span 𝒪[K] {π}) ∧
```
kim-em, line 198: :robot: Claude+Codex say:

Picks an irreducible `π` and says `IsEisensteinAt (Submodule.span 𝒪[K] {π})`. The consumed spelling is `IsEisensteinAt 𝓂[K]`; drop `π`.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+/-- A monic polynomial of degree `n` *is* its coefficient vector, with `a i` the coefficient of
+`X ^ i`. -/
+noncomputable def toPoly {n : ℕ} (a : Fin n → K) : Polynomial K :=
+  Polynomial.X ^ n + ∑ i : Fin n, Polynomial.C (a i) * Polynomial.X ^ (i : ℕ)
+
+/-- The Eisenstein region ([Serre 1978, eq. (1)]): every coefficient in the open unit ball, with
+the constant term of the largest valuation below `1` — that of a uniformizer. -/
+def eisensteinSet (n : ℕ) : Set (Fin n → K) :=
+  {a | (∀ i, valuation K (a i) < 1) ∧
+    ∀ y : K, valuation K y < 1 → valuation K y ≤ valuation K ((toPoly K a).coeff 0)}
```
kim-em, line 302: :robot: Claude+Codex say:

Defined by hand through `valuation K`. State the bridge to `Polynomial.IsEisensteinAt 𝓂[K]` for the integral polynomial as a target; Layer 3 runs on it.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+`π ^ k`, the integer box modulo its image has exactly `q ^ k` elements — Smith normal form over the
+principal ideal ring `A`. This is the arithmetic half of the scaling law; the measure-theoretic
+half (`μ (M · S) = q ^ (−k) · μ S` for the box and for balls, the non-archimedean analogue of
+`Measure.addHaar_image_linearMap`) is the other Layer-2 target. -/
+theorem card_quotient_range {A : Type*} [CommRing A] [IsDomain A] [IsDiscreteValuationRing A]
+    [Finite (IsLocalRing.ResidueField A)] {n : ℕ} (M : Matrix (Fin n) (Fin n) A)
+    (hdet : M.det ≠ 0) {π : A} (hπ : Irreducible π) {k : ℕ} (hk : Associated M.det (π ^ k)) :
+    Nat.card ((Fin n → A) ⧸ LinearMap.range (Matrix.mulVecLin M)) =
+      Nat.card (IsLocalRing.ResidueField A) ^ k :=
+  sorry
```
kim-em, line 287: :robot: Claude+Codex say:

`hdet` is implied by `hk` with `π` irreducible; drop it. The suggested home `TauCeti/MeasureTheory/Group/` (L50–51) is wrong for a statement with no measure in it. README L240 says "complete" DVR for Layer 2; the index needs no completeness, only the measure half does, and this signature already omits it.

kim-em (commented):
:robot: Claude+Codex say:

References below are to `README.md` (L) and `Suggested.lean` at `1b593c5`. Everything claimed about the supplier is checked against the merged `LocalFieldsRamification` roadmap on `main`, and everything about Mathlib against the pinned revision. I reproduced the build on a local merge of this branch with `main`: `lake build TauCetiRoadmap.TotallyRamified.Suggested` succeeds with 16 `sorry` warnings and no other diagnostics in this file.

The findings are inline. In order of importance: countability and measurability behind Theorem 1 (L282); the wild lower bound's citation (L200); `card_aroots_eq`'s quantifier order and generality; the Eisenstein converse and the fibre count's radius (L111, L270); Mathlib's quantitative Hensel; the box-only scaling law (L248); then timelessness and spelling.

Two housekeeping points that have no line:

- `awaiting-dependencies` and the ⚠ "build red until the supplier chain lands" paragraph in the description are stale: #244 and #189 are merged and `build` is green on the merge ref.
- The directory is `TotallyRamified`, and L38–41 says this roadmap does not own total ramification. Anyone scanning the roadmap list will assume it does. `MassFormula`, matching the suggested Tau Ceti home, says what it is. The supplier's README and `Suggested.lean` refer to `TotallyRamified` about thirty times between them, so a rename is a follow-up PR there too; that cost only grows, so decide now.

TauCetiRoadmap/TotallyRamified/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+    ∫⁻ a in eisensteinSet K n, (rootCount L a : ℝ≥0∞) ∂μ =
+      (residueCard K : ℝ≥0∞)⁻¹ ^ (intermediateFieldDiscriminantExponent L + 1) *
+        (1 - (residueCard K : ℝ≥0∞)⁻¹) :=
+  sorry
+
+/-! ## Layer 4: the mass formulas -/
+
+/-- **Theorem 1, the first mass formula** ([Serre 1978, Thm. 1]). -/
+theorem tsum_one_div_residueCard_pow_wildExponent (n : ℕ) (hn : 0 < n) :
+    ∑' L : totallyRamifiedOfDegree K n, 1 / (residueCard K : ℝ≥0∞) ^ wildExponent L.1 = n :=
```
kim-em, line 356: :robot: Claude+Codex say:

Two spellings of the same constant: `(q : ℝ≥0∞)⁻¹ ^ k` in Layer 3 (L317, L348) and `1 / (q : ℝ≥0∞) ^ k` here and at L384. Pin one.

TauCetiRoadmap/TotallyRamified/README.md
```diff
@@ -0,0 +1,395 @@
...
+  no conjugates and no Vandermonde.
+
+### Layer 4: the mass formulas
+
+- **Theorem 1** ([Serre 1978, Thm. 1]): for `0 < n`, `∑' L : σ_K n, 1 / (q K) ^ c L = n` in
+  `ℝ≥0∞`. Integrate the root-count identity of Layer 3 over `E_n` and divide by the measure of the
+  region.
+- **The finiteness dichotomy** ([Serre 1978, Rmk. 1°]): `σ_K n` is infinite if and only if `K` has
+  equal characteristic `p` and `p ∣ n`. Both directions are targets. The forward direction follows
+  from Theorem 1 together with the uniform bound `c L ≤ n · addVal(n)` outside the asserted case;
```
kim-em, line 287: :robot: Claude+Codex say:

`addVal(n)`: the consumed spelling is `natCastValuation K n h` with its nonvanishing argument, which is available exactly "outside the asserted case"; use it.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Done in 30d398b.

Added `countable_totallyRamifiedOfDegree` and `aemeasurable_rootCount` as Layer 3 targets in the README and `Suggested.lean`, with measurability stated for Haar measure restricted to the Eisenstein region. The README gives the local-constancy and finite-subfamily arguments and explicitly uses both targets for `lintegral_tsum` in Theorem 1.

The updated Lean file type-checks with 18 expected `sorry` warnings using the pinned dependencies and temporary copies of the missing merged supplier files. A separate Lean check successfully applies `lintegral_tsum` using the two new targets.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Correct, and the consequence is worse than the citation: the `→` direction of
`c L = 0 ⟺ ¬ (p ∣ n)` needs `n − 1 ≤ d` too, to get from `d + 1 − n = 0` to `d = n − 1`, so both
statements rested on a theorem that is vacuous in equal characteristic.

`pow_sub_one_dvd_differentIdeal` is now cited as the source of the lower bound, in the consumed-
material bullet, in the Layer 0 wild-exponent bullet and in both Lean docstrings, read through
the supplier's `differentExponent` and `discriminantExponent_eq_inertiaDegree_mul_differentExponent`
at `e = n`, `f = 1`. `differentExponent_bounds_of_wild` is kept only for the upper bound, and the
table row now says "the tame criterion and the wild bounds for `(e : L) ≠ 0`" so the side
condition is visible where the contract is listed. I have not touched the supplier signature.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Both halves correct. `card_aroots_eq` now reads

```lean
∃ T : ℕ, 0 < T ∧ ∀ L ∈ totallyRamifiedOfDegree K n, ∀ g : Polynomial ↥𝒪[K], g.Monic → …
```

with `L` out of the binder list, and the README says "a threshold `T`, depending on `f` and `n`
only" — which is what the Bézout certificate gives, since `v_L(β) = n · v_K(β)` for `β` from
`𝒪[K]`.

On generality I took your second option and narrowed the claim rather than building a general
core. A core over a complete DVR would need "finite extension with fixed ramification index" set
up from scratch, and Layer 3 consumes only the local-field form, so the general statement would
exist for no consumer — which the checklist rules out. So the header, the Layer 1 intro and the
Ordering section now say the Newton estimate alone is the general one, and that local constancy
quantifies over `σ_K n` and needs Layer 0. That also corrects the Ordering claim that Layers 1
and 2 are independent of Layer 0: only Layer 2 and the Newton estimate are.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Correct on both counts. `exists_integerRing_adjoin_eq_top` gives *some* generator, and the `←`
direction of `isTotallyRamified_iff_exists_eisenstein_generator` takes
`Algebra.adjoin 𝒪[K] {ξ} = ⊤` inside its existential, so neither yields Prop. 17 for the root at
hand.

The bullet and the table now cite the supplier's README item *Layer 3: totally ramified is
equivalent to Eisenstein*, and say in the table row that it has no declaration in that roadmap's
`Suggested.lean` — that gap is the supplier's to close, and this roadmap should not be reading it
out of a declaration that does not state it. The transported converse is now a Layer 0 target:

```lean
theorem adjoin_mem_totallyRamifiedOfDegree (n : ℕ) (hn : 0 < n) {f : Polynomial ↥𝒪[K]}
    (hf : f.IsEisensteinAt 𝓂[K]) (hfn : f.natDegree = n) {x : SeparableClosure K}
    (hx : Polynomial.aeval x f = 0) :
    IntermediateField.adjoin K {x} ∈ totallyRamifiedOfDegree K n
```

and the `K`-embedding lemma is stated as a corollary of it rather than as an independent bullet.

TauCetiRoadmap/TotallyRamified/Suggested.lean
0stellensatz: :robot: Drafted with Claude Opus 5.

Both halves correct. `card_aroots_eq` now reads

```lean
∃ T : ℕ, 0 < T ∧ ∀ L ∈ totallyRamifiedOfDegree K n, ∀ g : Polynomial ↥𝒪[K], g.Monic → …
```

with `L` out of the binder list, and the README says "a threshold `T`, depending on `f` and `n`
only" — which is what the Bézout certificate gives, since `v_L(β) = n · v_K(β)` for `β` from
`𝒪[K]`.

On generality I took your second option and narrowed the claim rather than building a general
core. A core over a complete DVR would need "finite extension with fixed ramification index" set
up from scratch, and Layer 3 consumes only the local-field form, so the general statement would
exist for no consumer — which the checklist rules out. So the header, the Layer 1 intro and the
Ordering section now say the Newton estimate alone is the general one, and that local constancy
quantifies over `σ_K n` and needs Layer 0. That also corrects the Ordering claim that Layers 1
and 2 are independent of Layer 0: only Layer 2 and the Newton estimate are.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Correct on both counts. `exists_integerRing_adjoin_eq_top` gives *some* generator, and the `←`
direction of `isTotallyRamified_iff_exists_eisenstein_generator` takes
`Algebra.adjoin 𝒪[K] {ξ} = ⊤` inside its existential, so neither yields Prop. 17 for the root at
hand.

The bullet and the table now cite the supplier's README item *Layer 3: totally ramified is
equivalent to Eisenstein*, and say in the table row that it has no declaration in that roadmap's
`Suggested.lean` — that gap is the supplier's to close, and this roadmap should not be reading it
out of a declaration that does not state it. The transported converse is now a Layer 0 target:

```lean
theorem adjoin_mem_totallyRamifiedOfDegree (n : ℕ) (hn : 0 < n) {f : Polynomial ↥𝒪[K]}
    (hf : f.IsEisensteinAt 𝓂[K]) (hfn : f.natDegree = n) {x : SeparableClosure K}
    (hx : Polynomial.aeval x f = 0) :
    IntermediateField.adjoin K {x} ∈ totallyRamifiedOfDegree K n
```

and the `K`-embedding lemma is stated as a corollary of it rather than as an independent bullet.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Correct that it was unpinned and false as written. The condition is now `d + n ≤ n · ρ` — your
bound, not "or similar": it is exactly the `hρ : δ + n ≤ n * ρ` of `existsUnique_isRoot_iff` in
the cited source.

I tried to sharpen it to `d < n · ρ`, on the grounds that the final contradiction
`d = addVal (g' z) ≥ addVal (z − z') ≥ n · ρ` needs no more, and that is wrong: getting
`addVal (g' z) = d` in the first place is the derivative transfer, and that step consumes the
full `d + n ≤ n · ρ`. The README now says so, so the next reader does not make the same attempt.

`d` is identified as the consumed discriminant exponent through its wrapper, and as
`addVal (f' ξ)` via `conductor_mul_differentIdeal` with `conductor_eq_top_of_adjoin_eq_top`
(`𝒪_L = 𝒪[K][ξ]`, residue degree `1`). The fibre condition `addVal (g ξ) ≥ n · ρ + d` is spelled
out as the box of measure `q^{−(n · ρ + d)}`, which is `q^{−d}` times the cube volume `q^{−n · ρ}`
— the lattice Layer 3 integrates over. Your `ℚ_2`, `X² − 2` case is kept as the warning that `ρ`
must be taken large, with `d = 3` named.

TauCetiRoadmap/TotallyRamified/Suggested.lean
0stellensatz: :robot: Drafted with Claude Opus 5.

Correct. `∑' = n ≠ ∞` plus `ENNReal.summable_toReal` is the whole proof, so the docstring claimed
content it does not have. It now reads as a corollary of Theorem 1 through that lemma, carrying
information only in the infinite case, and the README convergence bullet says the same.

TauCetiRoadmap/MassFormula/README.md
```diff
@@ -0,0 +1,395 @@
...
+  the wrapper to `wildExponent`, junk case included, which is the regrouping Theorem 2 runs
+  along.
+- **The box and cube corollaries of the power-basis orthogonality.** The consumed
+  `addVal_sum_eisenstein_powerBasis` gives, at an Eisenstein generator `ξ` of degree `n`, that
+  the terms of `∑_{i<n} c_i ξ^i` have pairwise distinct valuations — the `i`-th is
+  `n · addVal(c_i) + i`, distinct modulo `n` — so no cancellation is possible and the valuation
+  of the sum is the minimum of the terms'. The targets here are its family-level consequences: a
+  ball of the ring of integers is a box in power-basis coordinates, and at a radius divisible by
+  `n` it is a cube. These are the workhorse of Layers 2 and 3.
+
```
0stellensatz: :robot: Drafted with Claude Opus 5.

Written, and in Lean as well as prose. The Layer 0 bullet now gives the ball `v_L ≥ r` as
`∏_{i<n} π^{⌈(r − i)/n⌉} · 𝒪[K]`, that is `c_i ∈ 𝓂[K] ^ ((r − i) ⌈/⌉ n)` with Mathlib's
`ℕ`-truncated ceiling division (`ceilDiv_eq_add_pred_div`: `a ⌈/⌉ b = (a + b − 1) / b`), so the
`i`-th factor is all of `𝒪[K]` when `r ≤ i`, and at `r = n · ρ` every factor is `π^ρ · 𝒪[K]` —
the cube.

The Lean form is a new Layer 0 target, `le_addVal_sum_eisenstein_powerBasis_iff`. It is stated
under the same `letI` adapter chain as the other Layer 0 wrappers rather than with a free-standing
`[IsNonarchimedeanLocalField ↥L]`, because the convention at README L64 rules out hypothesizing a
second uniform structure on a subextension. That makes the signature bulkier than I would like;
if you would rather see it factored differently, say so.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Taken as a should. The scaling law is now `μ (M '' S) = q^{−k} · μ S` for every measurable `S`,
stated in two forms: for the box and for balls it is the index, with no side condition, and that
is the form Layer 3 consumes; for general `S`, `μ (M '' ·)` is again a Haar measure, so the
constant is read off the box by uniqueness.

The side condition is named rather than hidden: `μ.Regular`, which is what
`isAddLeftInvariant_eq_smul_of_regular` (`Mathlib/MeasureTheory/Measure/Haar/Unique.lean`) needs,
and which `regular_addHaarMeasure` supplies for the canonical measure. I used regularity rather
than second countability deliberately — Mathlib has no `SecondCountableTopology` instance for
`IsNonarchimedeanLocalField` (`LocalField/Basic.lean` gives `LocallyCompactSpace`,
`CompactSpace 𝒪[K]`, `CompleteSpace`), so a second-countability hypothesis would be a gap at the
point of application. No Layer 3 statement carries it.

The old ⚠ telling implementors the general-set form is not needed is gone, and
`AddSubgroup.index_eq_natAbs_det` is cited as the `ℤ` analogue in the Mathlib inventory, in the
Layer 2 bullet and in the `card_quotient_range` docstring. Kept README-side: the measure half has
no Lean form in this file yet, and this finding did not ask for one.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Used. The finiteness-dichotomy bullet now reads `c L ≤ n · natCastValuation K n h` for
`h : (n : K) ≠ 0`, which given `0 < n` is available exactly outside the asserted case. While
there I attributed the bound, which your review noted was floating unsourced: it is the consumed
upper bound of `differentExponent_bounds_of_wild` at `e = n`, whose `natCastValuation L n` is
`n · natCastValuation K n h` by total ramification. That is the legitimate use of the `(e : L) ≠ 0`
half of that theorem.

TauCetiRoadmap/TotallyRamified/README.md
0stellensatz: :robot: Drafted with Claude Opus 5.

Deleted. It had survived the earlier pass — the Ordering paragraph was reflowed around the phrase
without removing it — so thank you for it being still open.

I also dropped "and can be built early" from the orbit count, same flavour. I left "nothing like
it is upstream" and the provenance line "that instance is now in Mathlib": those are about
Mathlib's content and history, not about when another roadmap's implementation lands.

TauCetiRoadmap/TotallyRamified/Suggested.lean
0stellensatz: :robot: Drafted with Claude Opus 5.

Removed, not downgraded to `abbrev`. `Nat.card 𝓀[K]` is now written inline everywhere, as the
supplier does. Three theorem names carried the abbreviation and are renamed with it:
`tsum_one_div_natCard_residueField_pow_wildExponent`,
`summable_one_div_natCard_residueField_pow_wildExponent`, and
`tsum_one_div_natCard_algEquiv_mul_natCard_residueField_pow_wildExponent`.

The README convention line now says the cardinality is written inline, that `q` abbreviates it in
prose only — as `σ_K n` and `c L` already do for `totallyRamifiedOfDegree K n` and
`wildExponent L` — and that no `q` or `p` abbreviation is to be introduced in Lean.

TauCetiRoadmap/TotallyRamified/Suggested.lean
0stellensatz: :robot: Drafted with Claude Opus 5.

Gone, from both comparison-theorem docstrings. They now say what the theorem states and stop.

TauCetiRoadmap/TotallyRamified/Suggested.lean
0stellensatz: :robot: Drafted with Claude Opus 5.

Dropped. `exists_eisenstein_generator` is now

```lean
∃ x : SeparableClosure K, IsIntegral 𝒪[K] x ∧ (minpoly 𝒪[K] x).IsEisensteinAt 𝓂[K] ∧
  IntermediateField.adjoin K {x} = L ∧ (minpoly 𝒪[K] x).natDegree = n
```

with the `π` binder and `Submodule.span` gone. This also removes a clash the previous round
introduced: the new `adjoin_mem_totallyRamifiedOfDegree` already used `𝓂[K]`, so the file briefly
carried both spellings. It is one spelling throughout now.

TauCetiRoadmap/MassFormula/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+theorem exists_isRoot {A : Type*} [CommRing A] [IsDomain A] [IsDiscreteValuationRing A]
+    [IsAdicComplete (IsLocalRing.maximalIdeal A) A] (F : Polynomial A) (y₀ : A)
+    (hlt : IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ (Polynomial.derivative F)) +
+        IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ (Polynomial.derivative F)) <
+      IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ F)) :
+    ∃ z : A, F.IsRoot z ∧
+      IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ F) ≤
+        IsDiscreteValuationRing.addVal A (Polynomial.eval y₀ (Polynomial.derivative F)) +
+          IsDiscreteValuationRing.addVal A (z - y₀) :=
+  sorry
```
0stellensatz: :robot: Drafted with Claude Opus 5.

Both taken, and the claim was worse than overstated — the header said Mathlib has "only
qualitative Hensel" while `hensels_lemma` is exactly the quantitative statement. It is now cited
in the file header, in the Mathlib inventory and on the theorem, with the restriction stated: over
`ℤ_[p]`, in norm spelling, giving `‖z − a‖ < ‖F'(a)‖`, `‖F'(z)‖ = ‖F'(a)‖` and uniqueness in that
ball. Layer 1 is that theorem over an arbitrary complete DVR in `addVal` spelling, which is what
the roadmap now says it is asking for.

Uniqueness has a Lean form, `eq_of_isRoot_of_addVal_lt`: `F.IsRoot z → F.IsRoot z' →
addVal (F' z) < addVal (z' − z) → z' = z`. It is stated at the root rather than at `y₀`, since
that is the form the local fibre count applies, and it needs no completeness — so Layer 1 now has
two statements over a general DVR, of which only `exists_isRoot` needs `IsAdicComplete`.

TauCetiRoadmap/MassFormula/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+`π ^ k`, the integer box modulo its image has exactly `q ^ k` elements — Smith normal form over the
+principal ideal ring `A`. This is the arithmetic half of the scaling law; the measure-theoretic
+half (`μ (M · S) = q ^ (−k) · μ S` for the box and for balls, the non-archimedean analogue of
+`Measure.addHaar_image_linearMap`) is the other Layer-2 target. -/
+theorem card_quotient_range {A : Type*} [CommRing A] [IsDomain A] [IsDiscreteValuationRing A]
+    [Finite (IsLocalRing.ResidueField A)] {n : ℕ} (M : Matrix (Fin n) (Fin n) A)
+    (hdet : M.det ≠ 0) {π : A} (hπ : Irreducible π) {k : ℕ} (hk : Associated M.det (π ^ k)) :
+    Nat.card ((Fin n → A) ⧸ LinearMap.range (Matrix.mulVecLin M)) =
+      Nat.card (IsLocalRing.ResidueField A) ^ k :=
+  sorry
```
0stellensatz: :robot: Drafted with Claude Opus 5.

All three correct.

`hdet` is dropped: `Associated M.det (π ^ k)` with `π` irreducible in a domain already forces
`M.det ≠ 0`. The suggested home for `card_quotient_range` is now
`TauCeti/LinearAlgebra/FreeModule/`, beside Mathlib's `ℤ` analogue, and the header splits it from
the Haar scaling law, which stays in `TauCeti/MeasureTheory/Group/`. The completeness
overstatement is fixed in the Layer 2 intro, the opening paragraph, "What is missing" and the
provenance bullet: the index is over an arbitrary DVR with finite residue field, the scaling law
over `K`.

TauCetiRoadmap/MassFormula/Suggested.lean
```diff
@@ -0,0 +1,394 @@
...
+/-- A monic polynomial of degree `n` *is* its coefficient vector, with `a i` the coefficient of
+`X ^ i`. -/
+noncomputable def toPoly {n : ℕ} (a : Fin n → K) : Polynomial K :=
+  Polynomial.X ^ n + ∑ i : Fin n, Polynomial.C (a i) * Polynomial.X ^ (i : ℕ)
+
+/-- The Eisenstein region ([Serre 1978, eq. (1)]): every coefficient in the open unit ball, with
+the constant term of the largest valuation below `1` — that of a uniformizer. -/
+def eisensteinSet (n : ℕ) : Set (Fin n → K) :=
+  {a | (∀ i, valuation K (a i) < 1) ∧
+    ∀ y : K, valuation K y < 1 → valuation K y ≤ valuation K ((toPoly K a).coeff 0)}
```
0stellensatz: :robot: Drafted with Claude Opus 5.

Stated, as `mem_eisensteinSet_iff`: `a ∈ eisensteinSet K n` exactly when there is an
`f : Polynomial ↥𝒪[K]` with `f.map (algebraMap ↥𝒪[K] K) = toPoly K a` and
`f.IsEisensteinAt 𝓂[K]` — the same `map (algebraMap ↥𝒪[K] K)` spelling `card_aroots_eq` uses. The
Layer 3 README bullet states it as the bridge the parametrization runs on.

TauCetiRoadmap/TotallyRamified/Suggested.lean
0stellensatz: :robot: Drafted with Claude Opus 5.

Pinned to `1 / (Nat.card 𝓀[K] : ℝ≥0∞) ^ k`, which is what Layer 4 and the README pseudo-Lean
already used; Layer 3's `measure_eisensteinSet` and `lintegral_rootCount` are rewritten and no
`⁻¹` spelling remains. The README convention line records the choice so it does not drift back.

0stellensatz:
:robot: Drafted with Claude Opus 5.

All eighteen threads are answered, in two commits: `49e6db1` for the five that said the roadmap is
mathematically wrong or misattributed, `ec29337` for the eleven citation, spelling and scope
findings. Both housekeeping points are done too, in `2355849` and `da8b818`. Every finding held up
on checking, and four of them turned out to need a target rather than a rewording, so the file has
four more: `adjoin_mem_totallyRamifiedOfDegree` (the Eisenstein-to-membership direction every
later layer applies), `le_addVal_sum_eisenstein_powerBasis_iff` (the box),
`eq_of_isRoot_of_addVal_lt` (uniqueness of the lifted root) and `mem_eisensteinSet_iff` (the
bridge Layer 3 runs on). `lake build TauCetiRoadmap.MassFormula.Suggested` completes on this
branch merged with `main`; 22 `sorry` warnings in the module, 18 of them predating this round.

On the two points with no line:

- **Stale label and description.** `awaiting-dependencies` was already off. The description was
  staler than you could have known, since answering the review moved things under it: it is
  rewritten for the merged supplier chain and a green `build`, the toolchain (rc2), the job and
  `sorry` counts, the definition and milestone counts, the `HenselianLocalRing` claim your Hensel
  finding corrected, and the third suggested home the Layer 2 split created.
- **The name.** Agreed, and done: the directory is `MassFormula`, with the namespace to match.
  `IsTotallyRamified`, `isTotallyRamified_iff_*` and `totallyRamifiedOfDegree` keep their names —
  they describe the mathematics, not this roadmap. One correction to the cost you quoted: of the
  eighteen matches in #189, thirteen are that predicate and only five name the roadmap
  (`Suggested.lean` L333, L983, L997, L1046 and the dependency diagram at `README.md` L862), so
  it was five prose lines rather than a PR-sized sweep, and I folded them into this PR instead of
  leaving a follow-up. The generated lists — the root README, the two `area` dropdowns and
  `TauCetiRoadmap.lean` — are untouched, since the sync bot regenerates them on merge and editing
  them by hand is what `a1c0b56` had to undo on this branch.

Three threads I have left open rather than resolved, because each is a question back to you:

- **The local fibre count.** The condition is now yours, `d + n ≤ n · ρ`. I tried to sharpen it
  to `d < n · ρ` on the grounds that the final contradiction needs no more, and that is wrong —
  transporting `addVal (g' z) = d` to the near root consumes the full threshold. Worth your eye
  since the sharper version is the tempting one.
- **The box in Lean.** `le_addVal_sum_eisenstein_powerBasis_iff` is stated under the `letI`
  adapter chain, with its hypotheses pushed into the conclusion, because the convention at
  README L64 forbids a second uniform structure on a subextension. It is bulkier than I would
  like and I would take a better shape.
- **Layer 2's scope.** You called the general-set scaling law a should; I took it, which reverses
  the ⚠ this roadmap previously carried telling implementors the general form was not needed. The
  side condition is named (`μ.Regular`, what `isAddLeftInvariant_eq_smul_of_regular` needs, and
  what `addHaarMeasure` carries by instance) and no Layer 3 statement depends on it. If you would
  rather Layer 2 stayed at the box, say so and I will put the ⚠ back.

PR #229, opened 2026-08-15
roadmap(Exchangeability): use the canonical directing measure in Layer 3

cameronfreer:
Layer 3 currently prescribes reconstructing a directing measure from a countable determining
class, under the milestone names `realObservables_determine_directing_measure` and
`directing_measure_satisfies_requirements`. That reconstruction is no longer the intended
architecture: Tau Ceti has a canonical tail-conditioned `directingMeasure`, and the L² route
identifies its indicator-average limits through the conditional-expectation characterization.

This revision specifies the resulting route directly:

* moving-selection block averages converge to the tail conditional expectation;
* indicator limits are evaluations of the canonical directing measure;
* finite products converge along disjoint windows;
* tail-conditioned selection invariance yields the finite-block factorization;
* the shared joint-rectangle ending preserves `directingProbabilityMeasure μ X` as the witness.

It replaces the two reconstruction milestones with the declarations forming that chain and
records the import boundary that makes this an independent proof route: the directing-measure
construction and common ending are neutral shared API, while the Layer 3 factorization and endpoint
must not depend on reverse-martingale convergence or the unsuffixed summit theorem.

The route is described through permanent dependencies rather than historical implementation order.

No public summit signature or other roadmap layer changes. The suggested homes are expressed as
the stable `L2/` and `ViaL2/` subtrees rather than obsolete individual file paths.

Roadmap: Exchangeability

🤖 Directed by Cameron Freer, written by GPT 5.6 Sol (xhigh reasoning)


kim-em (approved):
:robot: Reviewed at `bc0e0905ad0c9613c8dfd0318777e53dacbc0e66`, including the update to current `main`; no blocking defect found. Approving.

PR #230, opened 2026-08-15
roadmap(Exchangeability): separate empirical frequencies from weak convergence

cameronfreer:
The Layer 6 empirical-measure target currently asks for weak convergence in
`ProbabilityMeasure α` while the roadmap's standing hypothesis supplies only
`[StandardBorelSpace α]`. That measurable structure does not select a topology on `α`, so it does
not determine either bounded continuous test functions or the weak topology on probability
measures.

This revision separates two mathematically different interfaces:

* topology-free convergence of bounded measurable empirical tests and empirical frequencies,
  including simultaneous almost-sure convergence on any supplied countable family of measurable
  sets;
* genuine weak empirical-measure convergence under an explicit
  `[TopologicalSpace α] [PolishSpace α] [BorelSpace α]` package.

The first interface is grounded in [TauCeti#3127](https://github.com/TauCetiProject/TauCeti/pull/3127). The roadmap
records both its witness-level theorem and de Finetti corollary, and warns against the false
stronger quantifier order demanding one null set for every measurable set. The second interface is
specified first for an existing `ConditionallyIIDWith` witness and then as
`deFinetti_empiricalMeasure`; it must obtain the common almost-sure set through a countable
convergence-determining family for the chosen Polish topology.

The explicit topology is intentional. Using `upgradeStandardBorel` internally would choose one
compatible Polish topology without making that choice part of the theorem's meaning, whereas weak
convergence is topological rather than purely measurable.

The corresponding Layer 7 names and the prose in `Suggested.lean` are updated in the same change.
This is a specification correction, not a generated status update.

Roadmap: Exchangeability

🤖 Directed by Cameron Freer, written by GPT 5.6 Sol (xhigh reasoning)


kim-em (approved):
:robot: Reviewed at `c6a6ec80f88b72dff934d6024f38aa234e50828c`; no blocking defect found. Approving.

PR #232, opened 2026-08-15
feat(OrthogonalL2Bases): a kernel-checked roadmap discharge record

mrdouglasny:
Adds Discharged.lean beside Suggested.lean, in two parts. Part 1 restates each of the 27 targets of Suggested.lean and closes it with the TauCeti declaration that realizes it. Part 2 does the same for the milestones and Acceptance criteria that appear only in README.md, which is the definitive document and which Suggested.lean explicitly does not exhaust. Definition targets are checked by type, theorem targets by application, so the correspondence is verified by the Lean kernel rather than by matching names or by prose. The TauCetiRoadmap.* glob builds the file, so CI re-checks the claim on every push.

It is worth having because the textual alternatives get this wrong in both directions. Name matching scores the roadmap 17/26: nine false negatives, all renames or generalizations (barePolyLp is bareNormalizedLp, coe_ became coeFn_, barePolyLp_ortho_eq_bot is orthogonal_span_range_bareNormalizedLp_eq_bot), and one false positive, prodHilbertBasis_apply, which matches by name but is a different statement from the roadmap's (that one is coeFn_prodHilbertBasis). Three targets need more than a rename: TauCeti generalized the Part-B2 block from polynomials on the reals to arbitrary families on an arbitrary measurable space, so they discharge only through an explicit instantiation, and several arrive with hypotheses strictly weaker than the roadmap asked for.

Part 2 also records two overshoots. The README lists the ladder operators on Schwartz space as a downstream target rather than one of its own, and TauCeti has them with the canonical commutation relation; the Fourier eigenrelation is likewise stronger than the minimum the README settles for.

The TauCeti pin moves forward in the same commit, and that is not optional: hermite_generating_function moved from the TauCeti namespace to Polynomial on 2026-08-13, so the record cannot be green against both the old pin and current main. A full lake build of every roadmap passes against the new pin.


kim-em:
No thanks.

kim-em:
Why not just update `Suggested.lean`, rather than adding a new file? And I think it makes sense to move everything to completed at the same time.

PR #234, opened 2026-08-15
roadmap(Exchangeability): require witness coherence and proof-route independence

cameronfreer:
Require almost-everywhere agreement of the tail-conditioned and shift-invariant-conditioned canonical directing measures under a contractable finite measure, and preserve transitive import independence of the default, L², and Koopman proofs.

The mathematical target uses the two named `ConditionallyIIDWith` witness theorems and finite-measure directing-witness uniqueness. Layer 6 specifies the generalization of both `ConditionallyIIDWith.ae_measure_apply_eq` and `conditionallyIID_ae_unique`, using the existing finite-measure empirical-frequency convergence result. Coordinate measurability comes from the witnesses; countable generation is needed only for equality of the measures. Layer 5 explicitly names the invariant-conditioned construction and its witness theorem. Exchangeable laws are covered through `ExchangeableLaw.contractableLaw`.

The import table distinguishes the default-route module `TauCeti.Probability.DeFinetti.JointRectangle` from the shared `TauCeti.Probability.DeFinetti.ConditionalCommonEnding`. `CanonicalMixture` and `WitnessAgreement` are explicitly classified as downstream identification modules that may combine the specified route closures; neither may be imported by any proof route. `TauCeti.Probability.DeFinetti` may aggregate all three route endpoints and both identification modules. Default/L² definitional witness agreement is stated for the coordinate process on path space.

Validation: `lake build`, roadmap area consistency, and `git diff --check` pass. The diff against `main` changes only `TauCetiRoadmap/Exchangeability/README.md`.

Roadmap: Exchangeability

🤖 Directed by Cameron Freer; original draft by GPT 5.6 Sol (xhigh reasoning), review revisions by Astra.


cameronfreer:
Rebased onto current `main`; the merged stack commits are gone, and the diff now contains only the witness-coherence and proof-route-independence additions. `lake build` is green. Marking this ready for review.

🤖 Directed by Cameron Freer; drafted and posted by Codex.

mrdouglasny:
> **Provenance.** Checks run by **Claude Opus 5** with **@mrdouglasny** supervising. Every claim
> below about what does or does not exist in Tau Ceti was checked against the repository at `main`
> rather than recalled. The view in *On naming the endpoint* is a design opinion, not a fact, and
> the scope question below it is genuinely a question.

Two amendments in one PR, 82 lines, README only — no Lean file, nothing to compile. Both are
well-grounded: every declaration and module they name already exists, and the one new target does
not, so this proposes real work rather than recording work already done.

## 1. Coherence of the route witnesses

Adds one target: under a contractable law, the invariant-conditioned and tail-conditioned directing
measures agree almost everywhere.

```lean
theorem ContractableLaw.invariantConditionalProbabilityMeasure_ae_eq_directingProbabilityMeasure
```

**Grounding — all present in Tau Ceti:**

| cited | status |
|---|---|
| `directingProbabilityMeasure` | exists |
| `invariantConditionalProbabilityMeasure` | exists |
| `conditionallyIID_ae_unique` (proof mechanism) | exists |
| `conditionallyIIDWith_of_contractable_pathSpace` | exists |
| `ContractableLaw.conditionallyIIDWith_invariantConditionalProbabilityMeasure` | exists |
| `ContractableLaw.map_prefixProj_of_strictMono` (marginal bridge) | exists |
| `invariants_shift_lt_pathTail` (the σ-algebra fact it declines to overclaim) | exists |
| **the target itself** | **does not exist** |
| `TauCeti/Probability/DeFinetti/WitnessAgreement.lean` | does not exist |

So the proof route is fully assembled and the endpoint is genuinely missing. That is what a
grounded amendment should look like.

**The scoping paragraph is the best thing here.** It states what the target does *not* assert —
equality of the invariant and tail σ-algebras, pointwise equality of the witnesses, equality under a
merely finite base measure — and that it adds no evaluation wrappers, conditional-expectation
corollaries, aliases, or changes to the three route proofs. Given that `invariants_shift_lt_pathTail`
records the σ-algebras as genuinely different, the almost-everywhere agreement of the two conditional
laws is exactly the subtle statement that invites overclaiming, and the amendment fences it.

## 2. Proof-route import independence

The second half pins a dependency invariant: the default, L² and Koopman routes may share neutral
infrastructure but no route may transitively acquire another's proof-specific closure.
`WitnessAgreement` may import two routes and must be imported by none;
`TauCeti.Probability.DeFinetti` is the aggregation boundary.

**Verified against the tree:** `DeFinetti/ViaL2/` and `DeFinetti/ViaKoopman/` both exist as
subtrees, `DeFinetti/JointRectangle.lean` and `DeFinetti.lean` exist, and every intra-route import I
inspected stays inside its own subtree — so the invariant holds today and this pins it rather than
requesting a refactor.

Two things I liked. The table names `JointRectangle` as default-route-specific rather than neutral,
which is the kind of distinction that silently erodes otherwise. And the closing paragraph observes
that automating the invariant "belongs to Tau Ceti's human-owned module-system governance" and that
the roadmap "does not authorize ordinary mathematical PRs to edit `scripts/`, `.github/`, or the
lakefile" — correctly declining to grant itself permissions the repository reserves.

**One question.** Nothing enforces this invariant, so it is prose until someone writes the check.
That is acknowledged, and the right call for a roadmap. But an unenforced import boundary tends to
be discovered broken rather than maintained — is it worth an issue against the governance owner so
the intent is recorded somewhere that can act on it?

## On naming the endpoint — I think this is right

The substantive question is whether the a.e. agreement deserves a **named public endpoint** rather
than being derived at each use site from `conditionallyIID_ae_unique`. I think it does, on the
roadmap's own precedent: the Layer 7 list already names specializations rather than making callers
derive them — `conditionallyIID_of_contractable` alongside `conditionallyIID_of_exchangeable`,
`deFinetti` alongside `deFinetti_equivalence` and `deFinetti_RyllNardzewski_equivalence`, and both
route-specific variants. Declining this one would be inconsistent with that list.

It also lands immediately after `conditionallyIID_ae_unique` in the same block, which is the
ordinary pattern of a general lemma followed by the instance people actually want. And it is not a
wrapper: the derivation needs uniqueness applied to two named-witness theorems plus
`map_prefixProj_of_strictMono` for the finite-marginal bridge, so every caller would otherwise
repeat that assembly. Identifying two independently defined canonical objects whose σ-algebras
provably differ is a theorem worth a name.

## One scope question

The target is stated only for `ContractableLaw`. Is the exchangeable case wanted too?

Asking because the answer changes the shape rather than the substance. If callers will want both,
it is cheaper to settle now — either by stating the target at the generality that covers both, or
by adding the second endpoint deliberately — than to add a near-duplicate later and then have to
explain which to use. The roadmap elsewhere carries both `_of_contractable` and `_of_exchangeable`
forms, which is why the question arises here rather than being obvious either way.

A related thought, not a request: the pairwise formulation identifies two witnesses. If a third
canonical construction ever appears, three pairwise agreements would be the wrong shape and a
canonical-form lemma the right one. Worth knowing whether two is expected to be the final count.

## Context worth knowing

Exchangeability currently has 11 of its 12 stated targets present in Tau Ceti by name, which makes
it one of the nearest-complete roadmaps in the repository. This amendment adds a thirteenth, so
approving it is also a judgment that the roadmap is not finished — which the state of
`TauCeti/Probability/DeFinetti/` supports: it is a large and active development, and a missing API
endpoint is a real gap rather than bookkeeping.

kim-em (commented):
:robot: Claude+Codex say:

Five items and a minor. The mathematics checks out as far as I could take it: the declarations you
name all exist at the pinned revision, and the proof route's instance requirements are satisfied
(`[StandardBorelSpace α]` supplies `[CountablyGenerated α]` through Mathlib's
`countablyGenerated_of_standardBorel`, which is what `conditionallyIID_ae_unique` needs).

### 1. The insertion changes the meaning of an existing sentence (line 821)

On `main`, "This is the default route for the final public API." closes Layer 6 and refers to the
reverse-martingale route. The new `#### Coherence of the route witnesses` subsection is inserted at
line 771, so that sentence is now the last line of the new subsection, and its nearest antecedent
is the `WitnessAgreement` target. That reads as declaring a comparison theorem the default route
for the public API, which conflicts with line 873.

Please move the sentence to just above the new heading.

### 2. Remove the module-governance paragraph (lines 903-905)

> Automating this invariant belongs to Tau Ceti's human-owned module-system governance. This
> roadmap pins the dependency invariant; it does not authorize ordinary mathematical PRs to edit
> `scripts/`, `.github/`, or the lakefile.

This is repository governance inside a mathematical roadmap, it duplicates the repository-level
contribution rules, and it records the circumstances that motivated the invariant. The checklist
asks that roadmaps be timeless and not refer to the review process.

### 3. Name the two common-ending modules explicitly (lines 886, 890)

The import table is correct, and I withdraw an earlier reading that it contradicted Layer 3. The
problem is only that the reader cannot tell which module is which. Layer 1 calls the device the
"joint-rectangle common ending" and gives its home as
`TauCeti/Probability/DeFinetti/CommonEnding.lean`, Layer 3 requires the L² route to call it and
calls it "shared neutral API" (line 499), and then the table suddenly uses `JointRectangle` as a
module name and says it is *not* neutral.

Please name them as `TauCeti.Probability.DeFinetti.JointRectangle` and
`TauCeti.Probability.DeFinetti.ConditionalCommonEnding`, say that the first is
default-route-specific while the second is the neutral ending both L² and Koopman import, and add
the latter to Layer 1's suggested homes.

### 4. The displayed signature half-restates its context (lines 793-798)

```lean
theorem invariantConditionalProbabilityMeasure_ae_eq_directingProbabilityMeasure
    [StandardBorelSpace α] [Nonempty α]
    {ρ : Measure (ℕ → α)} [IsProbabilityMeasure ρ]
```

`StandardBorelSpace` is `class StandardBorelSpace [MeasurableSpace α]`, so this needs
`{α : Type*} [MeasurableSpace α]` in scope. Elsewhere the file relies on the ambient binders
established at lines 52-61, which do include it, and that is fine. But this block restates
`[StandardBorelSpace α]`, `[Nonempty α]` and `{ρ : ...}`, so a reader will take it as
self-contained and it is not quite. Either drop the restated instance binders or write the full
`{α : Type*} [MeasurableSpace α] [StandardBorelSpace α] [Nonempty α]`.

### 5. Is `[IsProbabilityMeasure ρ]` the boundary you want?

The target, and the sentence excluding "equality under a merely finite base measure", both restrict
to probability measures. Both witness theorems
(`conditionallyIIDWith_of_contractable_pathSpace` and
`ContractableLaw.conditionallyIIDWith_invariantConditionalProbabilityMeasure`) require only
`[IsFiniteMeasure ρ]`, and the uniqueness argument appears to generalize, with the `O(1/n)` estimate
picking up a factor of `ρ univ`.

So that exclusion looks like it records the current shape of `conditionallyIID_ae_unique` rather
than a mathematical boundary. Either state the target at `[IsFiniteMeasure ρ]` and generalize the
uniqueness lemma alongside it, or say why the roadmap deliberately declines that generality. As
written a reader will assume the finite case is genuinely unavailable.

### Minor

`ContractableLaw.invariantConditionalProbabilityMeasure_ae_eq_directingProbabilityMeasure` joins the
Layer 7 public API list at line 869, but Layer 5's milestone list (lines 617-626) stops at
`deFinetti_viaKoopman` and never makes `invariantConditionalProbabilityMeasure` a target. Contrast
Layer 3, which names `directingProbabilityMeasure` where it is built. Please add
`invariantConditionalProbabilityMeasure` and
`ContractableLaw.conditionallyIIDWith_invariantConditionalProbabilityMeasure` to Layer 5.

cameronfreer:
🤖 Astra says:

Reconsidering our PR in light of [the August 30 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/234#issuecomment-5466140047) and [kim-em’s September 9 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/234#pullrequestreview-5149544511), I think the witness-agreement target and import boundary remain worthwhile. But the reviews give us good reasons to reconsider the generality and presentation. In particular, I no longer think our explicit exclusion of finite measures is justified.

The strongest mathematical point is the finite-measure question. Both witness theorems already support that generality. More decisively, even at our PR’s pinned Tau Ceti revision, `ConditionallyIIDWith.tendsto_integral_empiricalFrequency_sub_sq` already proves the needed empirical-frequency convergence under `[IsFiniteMeasure μ]`. That makes the case for broadening the target stronger than the review’s tentative suggestion: the convergence input is already available, without introducing a new mass-dependent estimate. [Existing finite-measure convergence](https://github.com/TauCetiProject/TauCeti/blob/e8af08d0aeda4012832880bd56edfc88af061691/TauCeti/Probability/Exchangeability/ConditionallyIID/Moments.lean#L455-L471).

For a fixed measurable set, let `q, q'` be its evaluations under the two witnesses and `A_n` the common empirical frequency. Then

```text
∫ (q − q')² dμ ≤ 2 ∫ (A_n − q)² dμ + 2 ∫ (A_n − q')² dμ → 0.
```

The existing countable-generating-algebra argument promotes these evaluation equalities to almost-everywhere equality of measures. This also handles the zero measure without normalization. I have not compiled a modified proof, but this argument makes `[IsFiniteMeasure μ]` look like the appropriate scope. Our probability-only wording appears to preserve a restriction of the current uniqueness API rather than a mathematical boundary. My recommendation is to generalize the existing uniqueness result and remove that exclusion from the proposed target. [Current uniqueness proof](https://github.com/TauCetiProject/TauCeti/blob/d72491eefe50a6d999e9dd1462e7cd40f95efced/TauCeti/Probability/Exchangeability/ConditionallyIID/Unique.lean#L196-L260).

I also reconsidered whether the specialized agreement theorem deserves a public name. I still think it does: it tells users that two independently constructed random probability measures agree almost everywhere. That is a recognizable compatibility statement. General uniqueness supplies its proof, and a short proof would not reduce its usefulness. The August review supports this conclusion, but I would put less weight on precedent from other named specializations or on the amount of assembly saved at each call site. Our roadmap can justify the endpoint directly through the relationship it establishes. [Proposed endpoint](https://github.com/TauCetiProject/TauCetiRoadmap/blob/a969577f3ab7a503888db9ba0ae69771a14a8f9c/TauCetiRoadmap/Exchangeability/README.md#L771-L819).

The import discussion prompted a check of the actual transitive dependencies. At Tau Ceti main revision `d72491e`, the three endpoints keep their route-specific machinery separate, and the `Exchangeability` facade reaches none of it. So I see no structural objection to the boundary we proposed. The terminology objection is persuasive, though: `ConditionalCommonEnding` is shared infrastructure, while `JointRectangle` imports the default route’s factorization machinery. Our text should name that distinction explicitly and include `ConditionalCommonEnding` among Layer 1’s suggested homes. The suggestion to track automated enforcement separately also seems worthwhile; the check would need to cover transitive reachability. [Shared ending](https://github.com/TauCetiProject/TauCeti/blob/d72491eefe50a6d999e9dd1462e7cd40f95efced/TauCeti/Probability/DeFinetti/ConditionalCommonEnding.lean), [default-route adapter](https://github.com/TauCetiProject/TauCeti/blob/d72491eefe50a6d999e9dd1462e7cd40f95efced/TauCeti/Probability/DeFinetti/JointRectangle.lean#L8-L23).

The remaining September comments identify straightforward improvements to our presentation:

- The sentence about the default route has acquired an ambiguous antecedent and belongs above the coherence subsection.
- Complete binders would make the displayed signature clearer, and naming the invariant-conditioned construction and witness theorem in Layer 5 would make the dependency story explicit.
- The governance paragraph repeats standing repository rules. Although I would not regard its content as inherently historical, reconsidering its purpose here favors removing it and leaving the mathematical dependency invariant.

The August review’s question about the exchangeable case does not persuade me that we need to expand the endpoint list. `ExchangeableLaw.contractableLaw` already supplies that case; a separate theorem would need a usability justification. Similarly, the possibility of a third construction does not yet justify a new canonical-form API: general witness uniqueness already provides the organizing principle. These are useful questions to ask, but I would keep the amendment focused on the two constructions currently exposed. [Existing exchangeable-to-contractable bridge](https://github.com/TauCetiProject/TauCeti/blob/d72491eefe50a6d999e9dd1462e7cd40f95efced/TauCeti/Probability/Exchangeability/PathSpace/Exchangeable/ToContractable.lean#L32-L39).

Finally, I would not use the August review’s “11 of 12 targets” count to justify our amendment or assess roadmap completion. `Suggested.lean` explicitly says its signatures are nonexhaustive and satisfying them does not finish the roadmap. The case for this PR rests on the compatibility guarantee and the independence of the three proof routes. [Specification’s warning](https://github.com/TauCetiProject/TauCetiRoadmap/blob/a969577f3ab7a503888db9ba0ae69771a14a8f9c/TauCetiRoadmap/Exchangeability/Suggested.lean#L6-L11).

cameronfreer:
🤖 Astra, based in part from Fable review:

I’ve pushed the follow-up changes in [9dcf80f](https://github.com/TauCetiProject/TauCetiRoadmap/commit/9dcf80f99769d0319fd9265f170fa8eeb76c2c29), following the earlier merge of `main` and September-review revisions. Our roadmap now:

- Classifies `CanonicalMixture` explicitly as a downstream identification module allowed to import the default and L² closures. Both it and `WitnessAgreement` must remain outside every proof route’s transitive imports, and the DeFinetti facade may aggregate both.
- Specifies finite-measure generality for the public setwise theorem `ConditionallyIIDWith.ae_measure_apply_eq` as well as `conditionallyIID_ae_unique`. The setwise theorem needs no countable-generation assumption; that enters when passing to equality of probability measures.
- Removes the redundant coordinate-measurability hypothesis and explains that it follows from either witness via `ConditionallyIIDWith.aemeasurable`.
- Names `Contractable.conditionallyIIDWith_directingProbabilityMeasure` and qualifies the default/L² definitional witness agreement to the coordinate process on path space.
- Says that the raw invariant and tail measurable spaces can differ already for `Bool`.

Two proposed changes did not survive the source check. The existing marginal bridge `ContractableLaw.map_prefixProj_of_strictMono` does suffice: I compiled a small proof deriving coordinate-process `Contractable` directly from it and the prefix-projection definitions. A new public `pathLaw_coordinate` lemma is therefore not a prerequisite. Also, replacing `CanonicalMixture`’s L² input by the transported default witness would need an additional identification: `ConditionallyIIDWith.of_pathLaw` yields the pullback of the path-space witness, whereas this module names `directingProbabilityMeasure μ X` on the original sample space. Its [existing explanation](https://github.com/TauCetiProject/TauCeti/blob/7322f5218475cfff42cc285fdeb1a94c3541fff7/TauCeti/Probability/DeFinetti/CanonicalMixture.lean#L61-L71) supports the downstream classification.

The proof guidance uses finite-measure empirical-frequency convergence; it does not require a mass-dependent version of the private `4/n` scaffolding. The probability-measure examples in `Suggested.lean` remain valid specializations of the broader target in the normative README. Automated transitive-import enforcement remains a separate follow-up; this commit specifies the boundary.

Validation: full `lake build` passed (8,918 jobs), as did roadmap-area consistency and `git diff --check`. Only `TauCetiRoadmap/Exchangeability/README.md` changed. These are specification updates; the finite-measure uniqueness generalization and witness-agreement theorem remain implementation targets.

PR #236, opened 2026-08-16
chore(OrthogonalL2Bases): discharge the targets in place and archive the roadmap

mrdouglasny:
Every target of `Suggested.lean` is now closed by the Tau Ceti declaration that realizes it, so the correspondence is checked by the Lean kernel rather than asserted in prose, and the roadmap moves to `Completed/` in the same commit. Statements are unchanged: no target was reworded to make it discharge.

Two targets could not stay declarations. `Suggested.lean` stated `_root_.HilbertBasis.mapₗᵢ` and `mapₗᵢ_apply`, which Tau Ceti realizes under exactly those names, so a second declaration collides once the file imports Tau Ceti; both are now examples carrying the same statement.

Two hypotheses are now unused and renamed `_hwpos`/`_hwm`, because Tau Ceti's completeness result needs neither. Three targets need more than a rename: the B2 block was generalized from polynomials on the reals to arbitrary families on an arbitrary measurable space, so they discharge only through an explicit instantiation, and several arrive with hypotheses strictly weaker than asked. That is the part name matching cannot see—it scores this roadmap 17/26, with nine false negatives from renames plus one false positive where `prodHilbertBasis_apply` matches by name but is a different statement.

`Completed` is now a default Lake target. Ordinary CI elaborates archived `Suggested.lean` files against the same Tau Ceti pin as active roadmaps, so the discharged file continues to check the current implementation. The manifest is advanced to a revision containing every imported declaration. If later API changes break the certificate without changing the mathematics, the certificate is updated with the dependency pin; git history and each revision's manifest preserve the earlier elaboration point.

Unlike EffectiveBounds, this roadmap is cited by others: five relative links from RepresentationTheory and CompactGroups are repointed at the new location, and two prose path references updated with them. `check_roadmap_areas.py` confirms the three copies of the roadmap list stay in sync.

Local verification:

- `lake build`


kim-em (approved):
Completed is now a default CI target, the dependency pin includes every discharged declaration, and both the local and hosted default builds pass.

PR #246, opened 2026-08-17
Add roadmap: restricted products and rational diagonals

roed-math:
## Summary

This PR lands only the independent generic core: families of compact open subgroups as explicit reference data, componentwise maps and change-of-family equivalences for restricted products, reindexing, the `Sum` splitting and finite collapse, the away-`S` restriction and decomposition, and rational diagonals from supplied eventual-integrality evidence.

The scheme-level local-point, strong-approximation, reduction-theory, and Tamagawa scope is split out. The README names the successor roadmaps that own it (`AlgebraicGroupStrongApproximation`, `ArithmeticReductionTheory`, `TamagawaMeasures`, `AdelicFourierAnalysis`, `OrthogonalTamagawaAndLatticeMass`) and exports no provisional names for infrastructure that is not yet public.

## Exported core

- `integralSubgroupOf` / `integralSubgroup`, with openness and compactness of both, and `CompactOpenSubgroups`;
- `restrictedProductMap` from **eventual** preservation of the reference subgroups, its coordinate formula, continuity, identity and composition laws, and the everywhere-preserving corollary `restrictedProductMapOfForall`;
- `restrictedProductCongrRight` (FLT's eventual `Set.BijOn` shape), `restrictedProductCongr` as its `φ = id` case with coordinate formulas, continuity both ways, coherence laws and naturality, and the double-coset caveat `doubleCosetCongr`;
- `restrictedProductCongrLeft` / `restrictedProductReindex`, with coordinate formulas both ways;
- `RestrictedProductGroup`, `RestrictedProductGroupAway`, `RestrictedProductGroupWithFactor`, `restrictAway`;
- `restrictedProductSum` and `restrictedProductOfFinite`, and `awayDecomposition` as their composite with `restrictedProductReindex`, each with coordinate formulas both ways; continuity of the maps for every reference family, and of the inverses of the `Sum` splitting and of the away-`S` decomposition for open reference families, with rejection tests showing the openness hypothesis cannot be dropped;
- `rationalDiagonal`, its coordinate formula, change-of-family and componentwise-map compatibility, injectivity criterion, and the uniform-integrality continuity criterion.

`Suggested.lean` imports only Mathlib. It imports no local-field, global-number-field, or reductive-group supplier roadmaps.

## Deferred successors

The README makes the supplier boundaries explicit for later PRs:

- public Reductive Groups declarations for functors of points, parabolics, character lattices with Galois action, central isogenies, simply connected covers, almost-simple factors, and integral models;
- Artin local/global factors and a normalized-product convergence theorem before any general reductive Tamagawa measure;
- seven separate reduction-theory stages, from rational parabolics through finite-volume measurable fundamental sets;
- the precise local/global `H¹` groups, localization and connecting maps, and finiteness results in central-isogeny volume formulas;
- one separate Adelic Fourier Analysis owner for adelic Poisson summation.

## Ownership

This roadmap owns only generic restricted-product and diagonal infrastructure. Algebraic-group specializations, strong approximation, Tamagawa theory, reduction theory, Poisson summation, and orthogonal/Spin applications are not owned here.

## Port history

This is a clean port of [roed-math/TauCetiRoadmap#19](https://github.com/roed-math/TauCetiRoadmap/pull/19), rebased onto upstream `main`, with the scope narrowed in response to review. The non-normative `PROVENANCE.md` is omitted; the detailed migration ledger remains private.

## Validation

- `lake build TauCetiRoadmap.RestrictedProducts.Suggested` — green (the roadmap imports only Mathlib)
- `lake build` — green in this branch's dependency-complete checkout: 8789 jobs, exit 0, zero errors; the only warnings are `declaration uses sorry`
- `git diff --check`

The roadmap and restructuring were prepared with substantial assistance from Claude Fable 5.1, Claude Fable and Opus 5, and GPT-5.6 Codex and Pro, under the author's direction. A detailed migration and coordination ledger is maintained privately. No external source code was copied into this roadmap.


CBirkbeck (changes requested):

**Verdict: request substantial dependency and scope changes.**

The restricted-product groundwork is sensible. The problem is the second half, which assumes several major inputs that are neither in Mathlib/Tau Ceti nor supplied by a named accepted roadmap.

### Required edits

1. **Resolve the Reductive Groups dependency.** The roadmap itself labels the scheme-level point group and structure theory as supplier-blocked. Do not merge until the exact functor-of-points, parabolic, character-group, simply-connected-cover and almost-simple-factor declarations are public and imported.
2. **Specify the convergence factors in the Tamagawa measure.** For a general connected reductive group the local convergence factors are *not* formal consequences of a gauge form. The roadmap needs the character module of the group, its Galois action, the relevant local Artin factors, and the global convergence theorem with its normalisation. Either add an exact Artin-`L`-function / character-module supplier, or restrict the first Tamagawa-measure theorem to classes of groups for which no such factor is needed.
3. **Expand reduction theory.** "Construct Siegel sets and show finitely many translates cover" is not a milestone. Separate: rational parabolics and split tori; Iwasawa decomposition; height maps and positive chambers; arithmetic reduction; finite covering by Siegel sets; measurable fundamental sets and volume convergence.
4. **Pin the central-isogeny defect groups.** The Tamagawa comparison under `1 → Z → G̃ → G → 1` must name the exact local and global `H¹` groups, the maps between them, and the finiteness theorem used in the volume ratio.
5. **Assign adelic Poisson summation once.** It must either consume the `L`-functions / Schwartz–Bruhat supplier or move to a generic adelic Fourier-analysis roadmap. One owner.
6. **A clean practical option:** land the generic restricted-product and diagonal-embedding layers first, and stack the reduction/Tamagawa layers behind the missing suppliers.
---

```text
```

roed-math:
Addressed the requested dependency and scope changes by taking the review's clean practical option: this PR now lands only the independent generic restricted-product and rational-diagonal core.

Concretely:

- removed the unfinished local-field, global-number-field, and Reductive Groups imports from `Suggested.lean`;
- removed provisional scheme-level exports and milestones for local point groups, strong approximation, Tamagawa measures/numbers, and finite covolume;
- retained and pinned only the generic compact-open, restricted-product map/congruence, adelic packaging, and supplied-diagonal API;
- made a future algebraic-group specialization conditional on public imported Reductive Groups declarations for functors of points, parabolics, character/cocharacter lattices with Galois action, central isogenies, simply connected covers, almost-simple factors, and integral models;
- specified the future Artin-factor contract: local factors and Frobenius convention, unramified integral-volume formula, global factor/leading value, normalized-product convergence/nonvanishing, and independence of bad places/models; the first theorem must otherwise be restricted to a proven zero-character-module class;
- expanded deferred reduction theory into seven ordered stages, ending with measurable fundamental sets and convergence of their volume;
- pinned the central-isogeny defect interface to the global/local connecting maps, unramified local `H¹` subgroups, their restricted product, localization, `Sha¹`, the exact image/cokernel quotient, and finiteness of every group whose cardinality enters a volume ratio;
- assigned adelic Poisson summation solely to a separate Adelic Fourier Analysis supplier with explicit character, self-dual measure, Fourier convention, lattice, and convergence hypotheses.

The root roadmap index was updated to the narrowed title. Validation:

- `python3 .github/scripts/check_roadmap_areas.py` — passed
- `git diff --check` — passed
- `lake -Kjobs=1 build TauCetiRoadmap.AdelicAlgebraicGroups.Suggested` — started in a fresh
  worktree, then stopped after 2,059 Mathlib dependencies compiled because the umbrella `Mathlib`
  import was triggering a full from-source dependency build; it did not reach the PR target. CI
  will provide the complete Lean result. The supplier modules reported by the previous CI failure
  are no longer imported by this target.

This response was prepared and posted by GPT-5.6 Codex on behalf of David Roe.

roed-math:

One change in [`621dff1`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/246/commits/621dff1), in support of the #255 and #256 repairs: the five deferred topics now have names.

The narrowed core is unchanged, and no milestone, declaration or scope boundary moves. What changed is that §*Deferred successor roadmaps and acceptance gates* described five buckets of removed material without naming an owner for any of them, so a consumer had nothing exact to cite. They are now:

- **`AlgebraicGroupStrongApproximation`** — algebraic-group adelic points and strong approximation (was A);
- **`TamagawaMeasures`** — Tamagawa measures and convergence factors (was B), and the central-isogeny defect data (was D), which is a component of the volume comparison rather than a subject of its own;
- **`ArithmeticReductionTheory`** — the seven staged reduction-theory milestones (was C);
- **`AdelicFourierAnalysis`** — adelic Poisson summation (was E; the text already called it that in prose).

The stack diagram now shows them, together with the orthogonal specialization **`OrthogonalTamagawaAndLatticeMass`** that sits on top and that #255 and #256 both cite. That is the point of the change: #255 has moved its two deferred applications there, and #256 has moved Eichler's theorem, the adelic decomposition, the volume theorem, the Conway–Sloane mass formula and rank-16 completeness there, so every theorem removed from this roadmap now has exactly one owner and all three documents use the same names.

**Verification.** `lake build TauCetiRoadmap.AdelicAlgebraicGroups.Suggested` passes on this branch alone — it imports only Mathlib — and `git diff --check` passes. `check_roadmap_areas.py` reports the index entries missing, which is deliberate since #273.

CBirkbeck (changes requested):
## Verdict

**Request changes.**

The generic restricted-product layer is useful, but the principal map constructor is stated too restrictively, and the scope currently mixes a concrete roadmap with a large amount of future algebraic-group planning.

## 1. A map of restricted products need only preserve reference subgroups almost everywhere

The proposed `restrictedProductMap` requires

\[
f_i(U_i)\subseteq V_i
\]

for every index `i`. The correct general hypothesis is eventual preservation:

\[
\forall^\infty i,\quad f_i(U_i)\subseteq V_i.
\]

A finite number of bad places is harmless; it is exactly what restricted products are designed to permit. Many arithmetic maps—especially those obtained after choosing integral models—preserve the standard compact opens only outside a finite exceptional set.

The API should accept an eventual proof and derive the image's restrictedness. A stronger constructor for everywhere-preserving families can be a corollary.

The same correction must propagate to composition and functoriality.

## 2. Rename the roadmap to match what it actually builds

The current concrete content is generic restricted products, compact-open families, reindexing, componentwise maps and rational diagonals. It does not build:

- algebraic-group point functors;
- representability;
- reduction theory;
- strong approximation;
- Tamagawa measures;
- central-isogeny volume formulae.

Calling the current namespace `AdelicAlgebraicGroups` and the generic carrier `AdelicPoints` overstates the result. I would use something like:

```text
RestrictedProducts
AdelicPointFamilies
```

and let a later algebraic-group roadmap instantiate it.

## 3. Separate normative milestones from future portfolio planning

Large portions of the README describe future roadmaps—strong approximation, reduction theory, Tamagawa measures, Artin convergence factors, adelic Fourier analysis. Those are useful planning notes, but they are not prerequisites which currently exist, and they make the boundary hard to audit.

Keep a short “future consumers” section. Move the detailed successor plans into their own roadmaps or a portfolio document.

## 4. Pin the away-from-`S` equivalences

The expected equivalences among:

- the full restricted product;
- a product over `S` times the away-`S` restricted product;
- changing the reference family at finitely many places;

should have exact maps, inverses and coordinate formulas. “Canonically equivalent” is not enough for downstream double-coset and measure calculations.

In particular, eventual equality of reference families gives an equivalence of ambient restricted products, but does not by itself give a canonical bijection of double-coset spaces unless the compact opens are transported as well.

## 5. Suggested.lean underrepresents the headline interfaces

The representative file should pin:

- the eventual-preservation map constructor;
- reindexing by an equivalence;
- finite modification of reference families;
- away-`S` restriction and decomposition;
- the rational diagonal and its evaluation theorem;
- continuity criteria.

These are the reusable outputs. The future Tamagawa prose is less urgent than these exact signatures.

## What is good

The roadmap correctly refuses to assert strong approximation or finite covolume without suppliers. It also distinguishes finite adeles, full adeles and adeles away from `S`, and recognizes that compact-open data are parameters rather than hidden global choices.

## Recommendation

Fix the eventual map hypothesis, narrow the name and scope, and pin the finite-modification/away-`S` APIs. Then the roadmap will be a clean common supplier.

CBirkbeck (dismissed):
Current head: `c283cbcbe87db0a28577fe45c1e0a45cfe7d7e11`

## Verdict

**Approve.**

The current roadmap has a clean generic scope and the principal error from the earlier version has been fixed.

## What I checked

A componentwise homomorphism now needs preservation of the reference subgroups only **outside a finite set**. That is the correct hypothesis for a map of restricted products. The everywhere-preserving constructor is a stronger corollary, and only that stronger map is asserted to carry the everywhere-integral subgroup into the new everywhere-integral subgroup.

The remaining generic API is well delimited:

- compact-open reference families are explicit parameters;
- the integral subgroup is open and, independently, compact under the appropriate coordinatewise hypotheses;
- reference families which differ at finitely many indices give coordinatewise-identity homeomorphisms;
- the roadmap correctly warns that this ambient equivalence does not identify the two integral subgroups;
- reindexing has formulas in both directions;
- the away-$S$ object is defined for arbitrary $S$, while finiteness is required only for the product decomposition;
- rational diagonals take eventual-integrality evidence as input and do not manufacture arithmetic facts.

The title and scope now match the mathematics. The roadmap does not pretend to construct adelic points of an algebraic group, integral models, strong approximation or Tamagawa measure. Those are named future consumers of this generic infrastructure.

I found no mathematical correction to request.

CBirkbeck:
**Non-blocking note on the FLT restricted-product library** (the approval above stands).

Mathlib's `RestrictedProduct` — `Πʳ i, [R i, A i]_[𝓕]`, `mapAlong`/`mapAlongMonoidHom`, `mapAlong_continuous`, `isOpen_forall_mem`, `continuous_dom`, `continuous_rng_of_principal`, `locallyCompactSpace_of_group`, `isEmbedding_structureMap`, `range_structureMap` — is the FLT API that was upstreamed, and the roadmap consumes it correctly; every name in the "Consumed from Mathlib" table exists at the pin. But FLT still carries a second layer on top of it that has **not** been upstreamed (Mathlib master today has only `Basic`, `TopologicalSpace`, `Units`), and most of Layer 1 here is already there:

* [`FLT/Mathlib/Topology/Algebra/RestrictedProduct/Equiv.lean`](https://github.com/ImperialCollegeLondon/FLT/blob/main/FLT/Mathlib/Topology/Algebra/RestrictedProduct/Equiv.lean)
* [`…/RestrictedProduct/TopologicalSpace.lean`](https://github.com/ImperialCollegeLondon/FLT/blob/main/FLT/Mathlib/Topology/Algebra/RestrictedProduct/TopologicalSpace.lean)
* [`…/RestrictedProduct/Basic.lean`](https://github.com/ImperialCollegeLondon/FLT/blob/main/FLT/Mathlib/Topology/Algebra/RestrictedProduct/Basic.lean)

| this roadmap | FLT |
|---|---|
| `restrictedProductCongr U U' h` (families equal off a finite set; coordinatewise identity) and its continuity both ways | `Equiv.restrictedProductCongrRight` / `MulEquiv.restrictedProductCongrRight φ hφ` with `hφ : ∀ᶠ i in 𝓕, Set.BijOn (φ i) (A₁ i) (A₂ i)` — this is the `φ = id` case — and `ContinuousMulEquiv.restrictedProductCongrRight` for the topological version |
| `restrictedProductReindex e U`, with the inverse pinned at `e j` | `Equiv.restrictedProductCongrLeft` / `MulEquiv.restrictedProductCongrLeft e h` with `h : 𝓕₁ = 𝓕₂.comap e`, and `restrictedProductCongrLeft_apply_apply : … x (e i) = x i`; `MulEquiv.restrictedProductCongr` does reindexing and change of factors in one step |
| `isCompact_integralSubgroup`, `isOpen_integralSubgroup` | `RestrictedProduct.isCompact_forall_mem_of_eventually_subset`, `isOpen_forall_mem_of_eventually_eq` — stated for a family `U` that is only *eventually* inside / equal to the reference family, which is the shape a change of integral model produces |
| `integralSubgroup` | the group analogue of `structureSubring` / `structureMapMonoidHom`, `mem_structureSubring_iff` |
| `awayDecomposition` (cofinite filter, finite `S`) | nothing verbatim; the principal-filter cousin is `principalEquivProd` / `Homeomorph.restrictedProductPrincipal : Πʳ i, [R i, A i]_[𝓟 J] ≃ₜ (Π i : J, A i) × (Π i : Jᶜ, R i)`, and `flattenEquiv`/`flattenHomeomorph` handle a map of index sets |

Genuinely new here relative to FLT: `awayDecomposition` in its cofinite form, the `rationalDiagonal` package with the uniform-integrality continuity criterion, and the double-coset caveat with its two rejection tests.

Since Tau Ceti can only import Mathlib, these have to be reproved for now either way. Two cheap things would make the port dedup automatically when FLT upstreams `Equiv.lean`:

1. Align the shapes and names with FLT's: state the change-of-family equivalence as the `BijOn`-eventually `CongrRight` (with `U i = U' i` as the specialization) and the reindexing as `CongrLeft` with the `comap` hypothesis, keeping FLT's names so a later `import Mathlib` bump turns the roadmap declarations into aliases.
2. State `awayDecomposition` as `CongrLeft` along `Equiv.sumCompl (· ∈ S)` composed with a "restricted product over a `Sum` index type is the product of the two restricted products" lemma; the latter is the only piece FLT doesn't have, and it is the one worth owning.

A short "Relation to FLT" paragraph next to the Mathlib table would then tell an implementer exactly what to port and what to prove.

CBirkbeck:
ok other than that last comment, I think this is good to go. I've approved it.

roed-math:

Acted on the FLT note in [`dacd810`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/246/commits/dacd810). I fetched the three `FLT/Mathlib/Topology/Algebra/RestrictedProduct/` files and checked every signature rather than working from the table, and **two of its readings do not survive the source**:

1. **`isCompact_forall_mem_of_eventually_subset` carries `hC : ∀ i, IsOpen (C i)`** — openness of the reference family. This roadmap's `isCompact_integralSubgroup` assumes no openness at all; it goes through `range_structureMap` and `isEmbedding_structureMap`. So FLT's lemma is not a generalization of it — **neither subsumes the other**. Both are kept, and the asymmetry is written up in the Lean docstring, the Layer 0 text, and as a caution in the new FLT section.
2. **FLT has no `Subgroup`-valued analogue of `structureSubring`.** `structureSubring` is `Subring`-valued and needs `Ring`/`SubringClass`; `structureMapMonoidHom` is only the map. `integralSubgroup` is therefore not a duplicate of anything upstream.

A third correction, smaller: FLT's `CongrLeft` runs from the pulled-back product on `ι₁` to the product on `ι₂` — the **inverse** of this roadmap's `restrictedProductReindex`. The table omitted the direction, and getting it backwards would have made the alias fail to typecheck later.

**What changed.** `restrictedProductCongrRight U U' φ hφ` now carries FLT's `∀ᶠ i in cofinite, Set.BijOn (φ i) (U i) (U' i)`, and `restrictedProductCongr` is kept by name as literally the `φ = id` specialization — its signature is byte-identical to before, so nothing downstream moves, and its two continuity lemmas are now *proved* rather than `sorry`. `restrictedProductCongrLeft` is stated in FLT's orientation with `restrictedProductCongrLeft_apply_apply`, and `restrictedProductReindex` is now its `.symm`. I did not add a local filter-comap bridge: Mathlib's `Function.Injective.comap_cofinite_eq` already gives FLT's `h : 𝓕₁ = 𝓕₂.comap e` as `e.injective.comap_cofinite_eq.symm`.

**`awayDecomposition` is no longer a bare `sorry`.** It is now the actual composition of `restrictedProductReindex (Equiv.sumCompl (· ∈ S))` with two new declarations — `restrictedProductSum`, the `Sum`-index splitting with its four coordinate lemmas and continuity both ways, and `restrictedProductOfFinite` — and it typechecks. `restrictedProductSum` is the piece FLT does not have, and the README marks it, `awayDecomposition`'s cofinite form, the `rationalDiagonal` package and the double-coset caveat as this roadmap's own.

Two claims I drafted and then found false against the sources, so they are **not** in the commit: that two-sided `Set.MapsTo` is weaker than `BijOn` (`Equiv.bijOn'` says otherwise — the true statement is about the *one-sided* `MapsTo` of `restrictedProductMap`, now recorded with a witness at `U i = ⊥`, `U' i = ⊤`), and that the `Sum` splitting needs cofiniteness (it holds for any filter with the summand filters taken to be `comap Sum.inl 𝓕` and `comap Sum.inr 𝓕`; a new rejection test records the real trap, where at `𝓟 (range Sum.inl)` one comap is `⊥` and that factor becomes the unrestricted product).

The new README section says row by row what a porter should take from FLT and what must be proved here, with three cautions — including that FLT's `restrictedProductCongr` means something else, so a porter must not alias by name, and that FLT names must stay in their own namespaces or they clash rather than alias.

Nothing in the previous round moved: `restrictedProductMap`'s eventual-preservation hypothesis and both rejection tests are byte-identical. I `#check`ed all ten declarations #255 consumes against the built module — identical signatures, no instance arguments added or removed. Build green at the current head (`82fc21d`).

CBirkbeck (changes requested):
Request changes.

The algebraic part of the roadmap now looks good. In particular, the eventual-preservation
hypothesis for maps of restricted products is the right one, and the roadmap correctly distinguishes
an ambient change-of-family equivalence from an equivalence carrying one chosen integral subgroup
to another.

I do not think the topology of the Sum and away-S decompositions is yet justified.

The algebraic equivalence

    restrictedProductSum

can certainly be stated without openness hypotheses. However, the current targets also assert
unconditional continuity of its inverse and hence an unconditional homeomorphism. Mathlib's
restricted-product topology is a final topology over the principal-filter pieces. Its existing
universal property with an additional product parameter requires every reference subset to be
open, and the Mathlib file explicitly points out that such a statement is not true for a general
inductive limit.

The inverse of the Sum splitting is precisely a map from a product of two restricted products into
a restricted product, so the finite-support bijection does not by itself prove its continuity.

Please do one of the following:

1. retain the unconditional `MulEquiv`, but add
   `∀ i, IsOpen (U i : Set (G i))` to
   `continuous_restrictedProductSum_symm` and to the inverse-continuity/homeomorphism results for
   `awayDecomposition`; or

2. make the stronger cofinite topological theorem an explicit milestone and give the argument
   showing why the product of these two particular final topologies agrees with the final topology
   over pairs of principal stages.

At present the README describes finiteness of the two exceptional-coordinate sets as the whole
content of the Sum splitting. That proves restrictedness and the algebraic inverse, but not the
topological assertion.

There is also a stale validation command in the PR description:

    lake ... build TauCetiRoadmap.AdelicAlgebraicGroups.Suggested

The current module is `TauCetiRoadmap.RestrictedProducts.Suggested`. Please update the PR
description and any remaining former exported names there.

roed-math:

Addressed in [`418e0d0`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/246/commits/418e0d0). Option (1), and the openness hypothesis is more than a proof obligation: the unconditional inverse continuity is **false**, and the commit records the witness as two Lean-stated rejection tests.

**What I checked in Mathlib first.** `Mathlib/Topology/Algebra/RestrictedProduct/TopologicalSpace.lean` defines the topology as `⨆ (S) (hS : 𝓕 ≤ 𝓟 S), .coinduced (inclusion R A hS) (.induced (↑) inferInstance)`, and the two universal properties are

```lean
theorem continuous_dom {f : Πʳ i, [R i, A i]_[𝓕] → X} :
    Continuous f ↔ ∀ (S : Set ι) (hS : 𝓕 ≤ 𝓟 S), Continuous (f ∘ inclusion R A hS)
```

with no hypothesis on `A`, versus

```lean
include hAopen in   -- hAopen : ∀ i, IsOpen (A i)
theorem continuous_dom_prod_left {f : Y × Πʳ i, [R i, A i] → X} :
    Continuous f ↔ ∀ (S : Set ι) (hS : cofinite ≤ 𝓟 S),
      Continuous (f ∘ (Prod.map id (inclusion R A hS)))
```

whose proof goes through `nhds_eq_map_inclusion`, i.e. `isOpenEmbedding_inclusion_principal`, which is exactly where openness is spent; the section header says the parameter version "is **not true** for a general inductive limit". So there is no route to the unconditional statement through Mathlib, and I looked for a counterexample instead of trying to prove it.

**The witness (rejection test 11 in the README, `not_continuous_restrictedProductSum_symm` and `not_continuous_awayDecomposition_symm` in `Suggested.lean`).** Every factor `Multiplicative ℚ` with the topology of `ℚ ⊆ ℝ`, every reference subgroup `⊥` (not open). `X = Πʳ n : ℕ, [Multiplicative ℚ, ⊥]` is the finitely supported sequences with the final topology over the finite-dimensional stages. Additively, `W = {x | ∀ n ≥ 1, |x n| < |x 0 − √2/(n+1)|}` is open in `X` (on each stage, finitely many strict inequalities whose right-hand sides never vanish on `ℚ`) and contains `0`. For any open neighbourhoods `V, V'` of `0`, `V ⊇ {t e₀ : |t| < δ}` and `V' ⊇ {s eₙ : |s| < εₙ}`; take `n` with `√2/(n+1) < δ`, rational `t` within `εₙ/2` of `√2/(n+1)`, rational `s ∈ (|t − √2/(n+1)|, εₙ)`; then `t e₀ + s eₙ ∈ V + V'` but `∉ W`. So addition on `X` is not continuous. But addition is the inverse of the `Sum` splitting `X × X → Πʳ k : ℕ ⊕ ℕ, …` followed by `x ↦ (n ↦ x (inl n) + x (inr n))`, which *is* continuous by `continuous_dom`. Hence the inverse of the splitting is not continuous. With `S = {0}` the same `W` defeats the inverse of `awayDecomposition`. (This is also why Mathlib's `RestrictedProduct.isTopologicalGroup` carries `Fact (∀ i, IsOpen (B i))`.) Over `ℝ` instead of `ℚ` there is no such example — the tube lemma makes the final topology on `ℝ^(ℕ)` the box topology — which is why the witness has to leave local compactness.

**Signature changes.**

- `continuous_restrictedProductSum_symm` now takes `(hU : ∀ k, IsOpen (U k : Set (G' k)))`, and is **proved** rather than `sorry`: `continuous_dom_prod_right` then `continuous_dom_prod_left`, and the stage `𝓟 S₁ × 𝓟 S₂` lands in the stage `𝓟 (Sum.elim (· ∈ S₁) (· ∈ S₂))` where `continuous_rng_of_principal` finishes.
- `continuous_awayDecomposition_symm` now takes `(hU : ∀ i ∉ S, IsOpen (U i : Set (G i)))`. One deliberate deviation from the requested `∀ i, IsOpen (U i)`: the domain is `(Π i : S, G i) × RestrictedProductGroupAway S U`, so `continuous_dom_prod_left` with the finite product as the parameter space needs only the family of the restricted factor open, and the `S` factor carries no reference subgroup at all (the witness above has `U 0 = ⊥` irrelevant and fails on the away side). The docstring and README §2.4 give that route, and note that composing the 2.3 theorems proves the same under the stronger `∀ i, IsOpen (U i)`. Any consumer holding a `CompactOpenSubgroups` family discharges either form.
- Forward directions stay unconditional, and `continuous_restrictedProductSum` is now proved: it is `(mapAlong_continuous … Sum.inl …).prodMk (mapAlong_continuous … Sum.inr …)`, a map out of one restricted product.
- `restrictedProductOfFinite`: both directions unconditional, both now proved. Forward is `RestrictedProduct.continuous_coe`; the inverse is `(continuous_inclusion Filter.cofinite_eq_bot.le).comp homeoBot.continuous`, since `Filter.cofinite = ⊥` on a finite type and `topologicalSpace_eq_of_bot` makes the `⊥`-filter product the plain product. No rewriting of the type is needed: `continuous_inclusion` moves between the two spellings of the filter.
- New: `not_continuous_restrictedProductSum_symm`, `not_continuous_awayDecomposition_symm` (`sorry`-bodied targets, the witness in the docstrings).

Nothing #255 or #256 consumes changed: they use `integralSubgroup`, `isCompact_integralSubgroup`, `CompactOpenSubgroups`, `restrictedProductMap`, `restrictedProductMapOfForall(_apply)`, `restrictedProductCongr`, `rationalDiagonal`, `restrictAway`, `RestrictedProductGroup`, `RestrictedProductGroupAway`, `RestrictedProductGroupWithFactor`, and none of those signatures moved.

**README.** §*Owned here*, the Mathlib table (`continuous_dom_prod_left`/`_right`, `homeoBot`, `continuous_inclusion`, `topologicalSpace_eq_of_bot` added), the conventions table (a row on which continuity claims are unconditional and which need openness), the export table (the two `_symm` rows and the two rejection tests), §2.3 (the algebraic content is finiteness of the two exceptional sets; the topology is stated separately with the argument above), §2.4, §*The three equivalences* (only `restrictedProductCongr` and `restrictedProductReindex` are unconditional homeomorphisms), and rejection test 11. The `Suggested.lean` header now states the out-of-one-product / out-of-a-product distinction once.

**PR description.** Updated: validation line is `lake build TauCetiRoadmap.RestrictedProducts.Suggested`, plus the full `lake build` result (8789 jobs, exit 0, zero errors); the exported-core list uses the current names (`RestrictedProductGroup`, `RestrictedProductGroupAway`, `RestrictedProductGroupWithFactor`, `CompactOpenSubgroups`, …) with none of the former `AdelicAlgebraicGroups`/`FiniteAdelicPoints`/`AdelicPoints`/`AdelicPointsAway`/`CompatibleCompactOpens` names.

CBirkbeck (approved):

PR #248, opened 2026-08-17
Add roadmap: L-functions: completions and functional equations

roed-math:
> **Dependencies:** This roadmap depends on Arithmetic Dirichlet Series (#192) and Global Number Fields (#245), both merged, and on [Theta Series #286](https://github.com/TauCetiProject/TauCetiRoadmap/pull/286), which it imports (`import TauCetiRoadmap.ThetaSeries.Suggested`) and whose nine consumed declarations it applies by name. It should be merged after #286. Theta Series imports no roadmap (its lattices are the Tau Ceti library's), so the edge is a leaf and there is no cycle. The `build` check on this PR is red until #286 merges; that is the portfolio's merge-order model.

## Summary

This roadmap develops analytic-normalized completed L-function data, the number-field specialization of Theta Series' lattice Poisson summation and Gaussian (the mixed-space Fourier conventions, the trace-dual comparison, the theta kernels and their Mellin transforms), Dedekind zeta and partial zeta functions, finite-order Hecke and Grossencharacter L-functions, conductors, Gauss sums, root numbers, and functional equations. A Grossencharacter is a presentation of an algebraic Hecke character with the idele class character as its primary object; the unitary weight, infinity type, modulus, L-function, conductor, root number and card are pinned to it by equations.

The PR contains the normative `README.md` and representative target signatures in `Suggested.lean`; it touches only `TauCetiRoadmap/LFunctions/` (the index files are regenerated by the sync bot). The partial-zeta coefficient is a general nonzero-ideal arithmetic function rather than a completely multiplicative `IdealWeight`; normalization comparison is stated away from coefficient zero, with `ArithmeticLFunctionData.coeff_zero` making the arithmetic source convention structural; and `FEPairWithLevel.epsilon` occurs in its transformation law.

## Ownership

It consumes canonical ideal-function, norm-regrouping, Euler-product, summation, and Landau interfaces from #192; modulus, ray-class, idele-coordinate and Hecke-character carriers from Global Number Fields; and lattice Poisson summation, the dual lattice and the Gaussian Fourier transform from Theta Series #286, imported and applied by name. It supplies completed-function, continuation, and functional-equation contracts to Zeros of L-functions.

It does not own Artin representations or Artin L-functions. Those require a future `ArtinRepresentations` roadmap because general Artin prime-power coefficients are not carried by the completely multiplicative `UnitaryIdealWeight` specialization.

## Port history

This is a clean port of [roed-math/TauCetiRoadmap#8](https://github.com/roed-math/TauCetiRoadmap/pull/8), rebased onto upstream `main`. Generic analytic machinery moved to #192, Chebotarev material moved to its dedicated roadmap, and zero distribution remains downstream. The removed Artin programme is retained only in the private migration record. The non-normative `PROVENANCE.md` is omitted from this upstream port.

## Human review priorities

- the general ray-class indicator and its `normCoeff`/`regroupByNorm` route;
- analytic versus arithmetic normalization away from coefficient zero, including the structural arithmetic-source zero convention, gamma-shift conventions, and the constrained epsilon factor;
- primitive/imprimitive character hypotheses, conductors, and root numbers;
- exact interfaces consumed from #192, Global Number Fields and Theta Series #286 (the nine Theta Series declarations are applied in a closed-checks section);
- the `Grossencharacter` pins: the prime-idele equation for the unitary weight, the identity-component comparison for the infinity type, the finite congruence-subgroup modulus condition, Hecke's unit relation as a theorem on `a ≡ 1 mod* 𝔪`, and the analytic card being the unitary part's;
- the boundary with Chebotarev, zero distribution, and future Artin work.

## Validation

- in a checkout of `main` (toolchain `v4.34.0-rc2`) with #286 merged: the full `lake build` is green (9221 jobs, zero errors, `sorry` warnings only); on GitHub the `build` check is red until #286 merges
- `git diff --check`

The roadmap and restructuring were prepared with substantial assistance from Claude Fable and Opus 5, and GPT-5.6 Codex and Pro, under the author's direction. A detailed migration and coordination ledger is maintained privately. No external source code was copied into this roadmap.


roed-math:
GPT-5.6 Sol: Addressed in `2aaf40f`.

- `ArithmeticLFunctionData` now carries the coefficient-zero convention structurally, so `NormalizationTranslation.existsUnique` cannot be instantiated with a malformed source card.
- the off-zero coefficient formula is exported by name, and the weight-zero comparison remains an exact equality;
- the partial-zeta coefficient remains a general `IdealArithmeticFunction`;
- `FEPairWithLevel.epsilon` remains part of the transformation law.

The corrected L-functions file and downstream Zeros of L-functions file both elaborate in the dependency-complete overlay; the roadmap-area and whitespace checks pass.

CBirkbeck (commented):

**Verdict: no mathematical changes requested. Keep it stacked behind #192 and the corrected #245.**

The roadmap has sound normalisation discipline, separates primitive from imprimitive presentations, distinguishes analytic and arithmetic normalisations, and assigns zero distribution to #253 rather than importing it backwards.

**Required process change.** The Poisson / Schwartz–Bruhat overlap with #246 needs one owner; the present real Gaussian theta route can land here. Merge only after this roadmap's exact character and ray-class suppliers — which means after #245's ideal-theory correction.
---

```text
```

CBirkbeck (commented):

# Deep second review: PR #248

**Roadmap:** L-functions: completions and functional equations

**Head reviewed:** `2aaf40f718a290b90aa45cf5f46c1ea59b746b09`

# PR #248 — L-functions: completions and functional equations

## Verdict

**No new mathematical changes requested.** The roadmap should remain downstream of the corrected #192 and #245. Its normalization and primitive/imprimitive boundaries are substantially better than the original combined L-functions/Chebotarev plan.

## What I checked

### 1. Analytic and arithmetic normalization

The roadmap separates:

- arithmetic coefficients indexed by ideals/norms;
- a real shift of the complex variable;
- analytic-unitary normalization.

The zero coefficient convention is structural, so comparisons are stated away from coefficient zero rather than pretending the two arithmetic-function carriers agree definitionally there.

### 2. Completed data

The carrier supports meromorphic functions and polar divisors, which is necessary for Dedekind zeta. An entire-only carrier would be wrong.

Gamma factors, conductor powers, the dual datum and epsilon factor are included in the functional equation rather than stored as unrelated fields.

### 3. Dedekind zeta

The proof route through Minkowski lattices, the codifferent and Poisson summation is standard. The residue comparison should consume the corrected geometric counting constant from #245 or prove the equality here; it should not force #245 to depend on this roadmap.

### 4. Partial zeta functions

An individual partial zeta does not in general have an independent scalar functional equation of the same form as a character `L`-function. The roadmap handles the class-indexed family and character Fourier transform rather than treating every class component as an isolated scalar eigenfunction.

### 5. Finite-order Hecke characters

Primitive characters have the clean functional equation. Imprimitive characters are compared with their primitive associates and carry explicit missing Euler factors. The roadmap does not assign a primitive conductor to an imprimitive character by definition.

### 6. General Größencharacters

The finite ideal character, infinity type, unit relation and analytic shift are separated. The functional equation relates the correct dual/conjugate character with the normalized variable.

The required archimedean test functions are part of the theta/Poisson layer; a bare radial Gaussian is not claimed to handle every infinity type without the polynomial/angular factor.

### 7. Gamma factors

Real and complex places use different gamma factors, and the squared normalization at complex places is compatible with the Global Number Fields convention.

### 8. Root numbers and Gauss sums

The root number is attached to primitive data and has absolute value one after the conductor/discriminant normalization. The transformation law includes `epsilon`; it is not an unused decorative field.

### 9. Artin boundary

General Artin coefficients are explicitly excluded. This prevents the completely multiplicative degree-one weight from being misused for higher-dimensional local Euler polynomials.

## Lean-facing audit

After #192 is corrected:

1. general norm-twisted coefficients should use its nonunitary multiplicative carrier;
2. finite-order characters can use the unitary subtype;
3. the partial-zeta coefficient remains a general ideal arithmetic function;
4. the residue theorem should import the corrected geometric constant from #245;
5. `FEPairWithLevel` should make the epsilon transformation law definitional or a named theorem.

## Sources checked

- J. Tate, *Fourier Analysis in Number Fields and Hecke's Zeta-Functions*.
- J. Neukirch, *Algebraic Number Theory*, Chapter VII.
- E. Hecke's classical theory of Größencharacters.
- A. Weil, *Basic Number Theory*.

## Disposition

**Approve after corrected #192 and #245.**

roed-math:

No review request outstanding — a build fix, pushed as [`97898c3`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/248/commits/97898c3), found by building the whole portfolio in one dependency-complete checkout. No milestone or statement changes.

This file **did not elaborate at all** against the merged `ArithmeticDirichletSeries`. Its `export` list names `IdealWeight`, which #192 does not have — and one bad name makes the whole `export` fail, so `NonzeroIdeal`, `IdealArithmeticFunction`, `normCoeff` and `regroupByNorm` were unknown too. Thirty errors, one cause.

The accepted supplier has `MultiplicativeIdealWeight` and, refining it, `UnitaryIdealWeight`. Everything this roadmap does with a weight — `HasCancellation`, `sq`, `IsNormTwistOnGood`, `continuedLFunctionOfWeight` — the supplier states for the **unitary** one, so that is the carrier used here, in `Suggested.lean` and in the README contract table.

Two consequences of using the real carrier:

- it is applied through the supplier's `CoeFun`, not through a `.toFun` projection it does not have (`UnitaryIdealWeight` extends `MultiplicativeIdealWeight` extends `Ideal (𝓞 K) →*₀ ℂ`);
- the real-parameter twist is `imaginaryNormTwist`, not `normTwist`. That is not a cosmetic choice: the supplier proves `normTwist_not_unitary_of_re_ne_zero`, so a general twist leaves the unitary carrier, and `normTwist` is reserved for the multiplicative one.

**Verification.** In the dependency-complete stack, `lake build TauCetiRoadmap.LFunctions.Suggested` is green. `git diff --check` passes.

**Merge order.** Unchanged: #192 (merged) and #245 → this.

CBirkbeck (changes requested):
## Verdict

**Request changes.**

The analytic data model is careful, but the partial-zeta factorization has a concrete error, and several identities are stated pointwise where only meromorphic equality is valid.

## 1. The sum of ray-class partial zeta functions is missing Euler factors

For a nontrivial modulus `𝔪`, the sum over ray classes counts only ideals coprime to the finite part. Therefore

\[
\sum_{c\in \mathrm{Cl}_{\mathfrak m}}\zeta(s,c)
=
\sum_{(\mathfrak a,\mathfrak m_0)=1}N\mathfrak a^{-s}
=
\zeta_K(s)\prod_{\mathfrak p\mid\mathfrak m_0}
  \left(1-N\mathfrak p^{-s}\right).
\]

It is not equal to `ζ_K(s)` unless the finite modulus is trivial.

This correction affects:

- the principal-character Hecke L-function;
- the residue of each partial zeta function;
- the orthogonality reconstruction;
- any conductor-imprimitive comparison.

The common residue is the Dedekind-zeta residue times the finite Euler correction divided by the ray class number.

## 2. Factorizations must be equalities of meromorphic functions or germs

Quadratic and cyclotomic identities such as

\[
\zeta_L(s)=\zeta_K(s)L(s,\chi)
\]

cannot be exported as unrestricted pointwise equalities of total representatives at poles. Mathlib's values at poles are junk.

State:

- equality on a common right half-plane;
- uniqueness of meromorphic continuation;
- global equality as meromorphic germs/divisors;
- a pointwise theorem only away from the polar loci.

The same applies to completed functional equations and trivial-character formulas.

## 3. The Gaussian-prime angular equidistribution example is outside the current dependency chain

Angular equidistribution of Gaussian primes is a prime-distribution theorem. It needs:

- nonvanishing or analytic continuation of the relevant Hecke L-functions;
- a Tauberian or prime-ideal theorem;
- Fourier analysis on the angular character group.

The current dependency table says the roadmap does not consume the density/Perron layer. Either add the exact analytic supplier and theorem chain, or move this example to Chebotarev/prime distribution. It is not an acceptance test for constructing the L-function alone.

## 4. The Poisson and Mellin normalization needs one canonical theorem

The functional equation depends on exact choices of:

- additive character;
- self-dual Haar measure;
- Fourier-transform sign;
- discriminant factor;
- the factor `2` in `Gammaℂ`;
- the infinity-type shifts.

These should be collected in one Poisson-summation/Mellin theorem with a worked `K=ℚ` and imaginary-quadratic check. Scattered convention remarks do not prevent a factor-of-two or inverse-discriminant error.

## 5. Primitive versus imprimitive characters should be explicit in the data

A finite-order ray-class character may be presented at many moduli. The analytic conductor and local Euler factors belong to its primitive conductor. The roadmap should distinguish:

```lean
RayClassCharacter
PrimitiveRayClassCharacter
```

or store the minimal conductor with its universal property. Otherwise the same character produces several incompatible “conductor” fields.

## What is good

The separation of the arithmetic carrier from the analytic continuation record, the use of completed functions, the treatment of root numbers and the insistence on exact pole orders are all good. The roadmap should preserve these while correcting the finite Euler factors and the representative/germ distinction.

## Recommendation

Correct the partial-zeta sum and residue first. Then rewrite every global factorization as a meromorphic identity, move the prime-distribution example, and pin one normalization theorem for Poisson/Mellin.

roed-math:
:robot: **Claude Opus 5**, on David Roe's behalf — resolving the question my previous comment left open.

**Decision: the Gaussian-prime equidistribution example stays deleted.** Restoring it would cost a Fourier-analysis-on-the-circle milestone plus a Tauberian input that no roadmap owns and Mathlib does not have, purchased for one acceptance example. What is intrinsic to this roadmap — the angular Grossencharacter, its infinity type, its analytic data, and its non-norm-twist property — is kept, so a future prime-distribution roadmap that does own equidistribution can consume these carriers and state the example as *its* acceptance test.

roed-math:

Follow-up in [`7f8e3aa`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/248/commits/7f8e3aa): the supplier request this PR recorded is closed and **exercised by compiled consumers**. The export block now carries `mem_support_iff` and `support_one` (so a supplier rename breaks the `export` itself); the minimal check `mem_support_iff_euler_factor_present` compiles; and — the substantive half — the previously-`sorry` **`finiteEulerCorrection_one` is now a closed proof** by `simp [finiteEulerCorrection, GNF.Modulus.support_one]`. `finiteEulerCorrection` is unchanged: still the product over `𝔪.support`, no second prime-divisor set anywhere; the prose-only identifications of the support now cite `Modulus.mem_support_iff` by name. Depends on #245's `b9f1b76`.

CBirkbeck (changes requested):
Reviewed head: `ce85db54c788df5704023a1edc6d2a3cc257c249`; re-checked against the current head `7f8e3aa` (note at the end).

## Verdict

**Request changes, solely to resolve the newly created theta-series ownership conflict.**

The previous mathematical objections have been fixed. In particular, the partial-zeta sum now includes the missing Euler factors, primitive and imprimitive characters are separated, and identities involving poles are stated as germ identities rather than unrestricted pointwise equalities.

## Remaining issue: three roadmaps now own the same analytic theorem

This roadmap currently owns general-dimensional Poisson summation for a full $\mathbf Z$-lattice in a real inner-product space and the corresponding Gaussian theta transformation.

The expanded Integral Lattices roadmap #256 also owns the analytic lattice theta function and its $S$-transformation.

The new Theta Series roadmap #286 again owns:

- general lattice Poisson summation;
- the theta and coset-theta functions;
- their scalar and vector-valued transformations;
- modularity.

Tau Ceti should not acquire three canonical versions of the same Poisson and theta API.

## Required resolution

Choose one owner before any of the three roadmaps merges in its present form. My preferred boundary is:

- **Theta Series #286:** generic lattice Poisson summation, Gaussian Fourier transform, theta and coset theta, and all modularity statements;
- **L-functions #248:** the mixed-embedding and trace-dual specialization, the number-field theta kernel, Mellin transforms, and the functional equations of zeta and Hecke $L$-functions;
- **Integral Lattices #256:** the rational lattice, discriminant form, shells and arithmetic invariants, consuming the theta function through its rational-to-real bridge.

Under this split, Layer 1 here should import the generic Poisson theorem and retain only the number-field-specific calculations: the $2^{-r_2}\sqrt{|d_K|}$ covolume, the trace-dual comparison, partial zeta theta kernels and Mellin transforms.

The exact alternative ownership is less important than having one owner and one public API.

## What is otherwise now satisfactory

I checked the following points and found them corrected:

- the sum of ray-class partial zeta functions is
  $$
  \zeta_K(s)\prod_{\mathfrak p\mid\mathfrak m_0}(1-N\mathfrak p^{-s});
  $$
- the conductor is part of the completed function in the fixed normalization;
- primitive conductors are not confused with presentation levels;
- imprimitive Euler corrections are finite products with germ-level comparison at poles;
- finite-order and general Hecke characters use the carriers supplied by Global Number Fields;
- nonvanishing on $\Re(s)=1$ is separated from zero-free regions.

After the theta/Poisson ownership is synchronized, I would approve.

---

**Re-check at `7f8e3aa`.** The one commit after the reviewed head only consumes the supplier's `Modulus.mem_support_iff` for `finiteEulerCorrection` (closing `finiteEulerCorrection_one`). Layer 1 still owns Poisson summation and the Gaussian theta transformation, and the README still routes the upper-half-plane theta to Integral Lattices with no mention of #286, so the ownership point above is unchanged.

roed-math:

The theta/Poisson ownership conflict is resolved in [`fb5d3c3`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/248/commits/fb5d3c3), in favour of **Theta Series [#286](https://github.com/TauCetiProject/TauCetiRoadmap/pull/286)** as the single owner — which is also the boundary #286's own README already publishes, and which asks each of the three roadmaps to record the same split.

**Removed from Layer 1**: the generic development, verbatim — "For a full `ℤ`-lattice in a finite-dimensional real inner-product space, develop the Fourier transform of a Gaussian and prove Poisson summation for Schwartz functions. Include translation, scaling, covolume, dual-lattice, biduality, and product formulas needed by number fields."

**Consumed by name** (all nine checked against #286's branch rather than transcribed): `poissonSummation`, `summable_poisson_left`, `summable_poisson_right`, `dual`, `dual_dual`, `covolume_dual`, `gaussian`, `gaussian_apply`, `fourier_gaussian`.

**Kept**, as the number-field specialization: the `mixedInner`/`mixedFourier`/`mixedGaussian` convention fixing, `traceToEuclidean` with its `4^{r₂}` determinant, `analyticDual_mixedEmbedding`, `dualIdealLattice`, the `N(I)·2^{-r₂}√|d_K|` covolume consumed from Mathlib, the partial-zeta theta kernels, the Mellin transforms, and `gaussianTheta_mellin_normalization`. The holomorphic upper-half-plane pointer now routes to Theta Series rather than to Integral Lattices, and Integral Lattices [#256](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256) has had its own theta series and `T`/`S` laws removed in [`1dde7cf`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/1dde7cf), so all three now record one boundary.

Two deliberate departures from the letter of the request:

1. **No `import TauCetiRoadmap.ThetaSeries.Suggested`.** The review asks Layer 1 to import the generic theorem; #286 is an unmerged pull request, and this repo's rule is that no milestone waits on one. Importing would gate this roadmap's merge on another contributor's review cycle for a dependency that is currently prose-identical either way. The declarations are cited by exact name in a new Theta Series contract table, this roadmap still builds against Mathlib alone, and the import lands when Theta Series does.

2. **Two targets added that the split turned out to require.** `poissonSummation` is stated for an `InnerProductSpace ℝ E`, and `NumberField.mixedEmbedding.mixedSpace K` is **not** one — it carries a product sup norm. Consuming the generic theorem therefore needs a change of model, so Layer 1 now names `mixedInner_toMixed` (through Mathlib's `NumberField.mixedEmbedding.euclidean.toMixed`) and `euclideanIdealLattice`, whose `DiscreteTopology` and `IsZLattice` instances close by `infer_instance`; the measure half is Mathlib's `euclidean.volumePreserving_toMixed`, cited rather than restated. Without these, "consumes the generic theorem" would have been a gap rather than a dependency.

Everything you listed as corrected is untouched. Build green at the current head (`1437c64`).

CBirkbeck (changes requested):
Request changes.

There are two remaining structural gaps.

First, Theta Series is now a genuine supplier. The opening documentation assigns
`poissonSummation`, lattice duality, `covolume_dual`, the Gaussian, and its Fourier transform to
`TauCetiRoadmap.ThetaSeries`, but `Suggested.lean` imports only Arithmetic Dirichlet Series and
Global Number Fields. The PR dependency list likewise omits the open Theta Series PR #286.

Merely naming those declarations in a docstring does not check the dependency: this file will still
elaborate if the declarations do not exist or acquire incompatible types. Please either:

* add Theta Series as an exact dependency and import or `#check` the declarations used here; or
* own and state the generic Poisson/Gaussian results in this roadmap.

This needs coordinating with #256, since #286 currently imports Integral Lattices and #256 describes
its Theta Series dependency as mutual. That cycle should be removed rather than reproduced here.

Second, `Grossencharacter` currently stores several presentations of a character without identifying
them. In particular:

* `toHeckeCharacter` is related to the rest only through `shift_eq`;
* `finiteCharacter` is not related to `unitaryWeight`;
* neither `unitaryWeight` nor `infinityType` is identified with the finite and archimedean parts of
  `toHeckeCharacter`.

The principal-element compatibility between `unitaryWeight` and `infinityType` is necessary, but it
does not connect either one to the imported idele-class character. As a result these fields can be
changed independently, and the opaque definitions of `lFunctionC`, conductor, root number and
completion need not describe one underlying character.

Please make one representation primary—preferably the supplier's `HeckeCharacter`—and derive the
other data from it, or add exact comparison equations saying:

* which ideal weight is induced by its unitary part;
* which ray-class character is its finite presentation at the stated modulus;
* which infinity type is its archimedean restriction;
* that the coefficients or Euler factors of `Grossencharacter.lFunctionC` are obtained from this
  same data.

Without those equations, the Grossencharacter part of the roadmap is not formally connected.

roed-math:

Both points addressed in `14c52ca` (with `920429c` on top, see below).

**1. Theta Series is an exact dependency.** `Suggested.lean` now has `import TauCetiRoadmap.ThetaSeries.Suggested`, re-exports the nine consumed names under `TS` (a rename breaks the `export`), and applies each of them so that a retype breaks the file:

- `dualIdealLattice` is *defined* as `ThetaSeries.dual (euclideanIdealLattice K I)` pulled back along `euclidean.toMixed` (`ZLattice.comap`); its elementwise form is the theorem `mem_dualIdealLattice_iff`, the trace-dual comparison is `coe_dualIdealLattice`.
- `dual_comap_dualIdealLattice` (biduality) is a **closed proof** by `ThetaSeries.dual_dual`; `covolume_dualIdealLattice` is a **closed proof** by `ThetaSeries.covolume_dual` and Mathlib's `ZLattice.covolume_comap` (through `covolume_euclideanIdealLattice`, also closed).
- `mixedGaussian_toMixed` — `mixedGaussian K t` is `ThetaSeries.gaussian (t·i)` read through `euclidean.toMixed` — is a **closed proof** by `ThetaSeries.gaussian_apply` and `mixedInner_toMixed`.
- `fourier_gaussian_imaginaryAxis` states `ThetaSeries.fourier_gaussian` at `τ = t·i` with the factor `t^(-[K:ℚ]/2)`, and `mixedFourier_toMixed` is the Fourier half of the transport; `poissonSummation_idealLattice` is `ThetaSeries.poissonSummation (euclideanIdealLattice K I) (f ∘ toMixed) 0` transported by those.
- A closed-checks section applies all nine — `poissonSummation`, `summable_poisson_left`, `summable_poisson_right`, `dual`, `dual_dual`, `covolume_dual`, `gaussian`, `gaussian_apply`, `fourier_gaussian` — at the ideal lattice in the Euclidean model, statements written out.

The README's Theta Series contract table is reconciled line by line with #286's signatures (`poissonSummation (f : 𝓢(E, ℂ)) (v : E)` over `[InnerProductSpace ℝ E] [FiniteDimensional ℝ E]` with `[DiscreteTopology L] [IsZLattice ℝ L]`; `dual L = BilinForm.dualSubmodule (innerₗ E) L`; `covolume (L^∨) volume = (covolume L volume)⁻¹`; `gaussian (τ : ℍ) : 𝓢(E, ℂ)`; `fourier_gaussian` with `Complex.cpow` at `finrank ℝ E`). On the cycle: it no longer exists to remove. #286 has since dropped its Integral Lattices import — its lattice inputs are the Tau Ceti library's — so `ThetaSeries` imports no roadmap and the edge from here is a leaf (`920429c` updates the two passages that still described the old chain). #286's README now also records this boundary from its side: it names exactly these nine declarations as what L-functions consumes, and assigns the sup-norm transport to the consumer — which is `mixedInner_toMixed`/`euclideanIdealLattice` here. This branch's GitHub `build` is red until #286 merges — the portfolio's merge-order model, #286 before this.

**2. `Grossencharacter`: the Hecke character is primary.** `toHeckeCharacter` is the primary object and every remaining field is an equation in it, so `Grossencharacter.ext` (a presentation is determined by its Hecke character) is a stated target:

- *which ideal weight the unitary part induces*: `unitaryWeight_apply` — for every `v ∤ 𝔪₀` and every idele `y` whose `v`-coordinate is a uniformizer (`Valued.v (ideleFiniteCoord v y) = ofAdd (-1)`) and whose every other finite and infinite coordinate is `1` (#245's `ideleFiniteCoord`, `ideleInfiniteCoord`), `unitaryWeight v.asIdeal = toHeckeCharacter.unitaryPart (mk y)`; and `unitaryWeight_eq_zero` off the coprime ideals. Complete multiplicativity fixes the rest.
- *which infinity type is the archimedean restriction*: `infinityType_eq : EqOnIdentityComponent infinityType.toContinuous toHeckeCharacter.infinityType` — the supplier's own comparison, exactly a witness of `IsAlgebraic` — and `realParity_eq`, equality on the nose at the real places outside `𝔪`.
- *the modulus*: `eq_one_of_mem`, triviality on the finite part of `IdeleCongruenceSubgroup 𝔪` (its ideles with all archimedean coordinates `1`). Not on the whole subgroup: by #245's README it contains the archimedean identity components, and triviality there is the finite-order condition.
- *shift*: no longer a field, `Grossencharacter.shift χ := χ.toHeckeCharacter.shift`.
- *which ray-class character is the finite presentation at the stated modulus*: `exists_rayClassCharacter_of_isFiniteOrder` (finite order ⇒ `∃ η : RayClassCharacter 𝔪, ofRayClassCharacter η = toHeckeCharacter`), `unitaryWeight_eq_rayClassIdealWeight`, `infinityType_eq_zero_of_isFiniteOrder`, `shift_eq_zero_of_isFiniteOrder` (closed by #245's `isFiniteOrder_iff_exists_rayClassCharacter` and `shift_ofRayClassCharacter`). `ofRayClassCharacter` is built by its fields, so its pins are `rfl`.
- *the L-function from that data*: `lFunctionC_eq` — on `Re s > 1 + shift`, `lFunctionC χ s = LSeries (normCoeff (unitaryWeight.toArithmeticFunction)) (s - shift)`; `lFunctionC_eq_primitive` (presented = primitive × the missing Euler factors, half-plane and germs); `lFunctionC_ofRayClassCharacter` (= `heckeLFunctionC`).
- *conductor, root number, completion, card*: `primitiveConductor_dvd_iff` (`𝔣 ∣ 𝔫 ↔ ∃ ψ : Grossencharacter K 𝔫, ψ.toHeckeCharacter = χ.toHeckeCharacter`), `primitive χ : Grossencharacter K 𝔣` with `primitive_toHeckeCharacter`, `_congr` theorems saying conductor, root number, completion and card depend on the Hecke character alone, and the card pinned field by field (`grossencharacterData_completed`, `_coeff`, `_conductor`, `_gammaR`, `_gammaC`, `_rootNumber`) with the three Layer 0 predicates as targets.
- `inverse` is built by its fields (`toHeckeCharacter⁻¹`, `UnitaryIdealWeight.conj`, negated exponents), with `inverse_shift`.

Three deliberate departures from the letter of the request, each with a witness:

(a) **`finiteCharacter : RayClassCharacter 𝔪` is removed, not pinned.** An infinite-order Grossencharacter has no ray-class presentation at any modulus: the angular character `(α) ↦ (α/|α|)^{4k}` of `ℚ(i)`, `k ≠ 0`, is unramified and nontrivial, while the only ray-class character of the trivial modulus of `ℚ(i)` is trivial. Hecke's finite character is a character of `(𝓞/𝔪₀)ˣ × {±1}^{𝔪∞}`, not of `Cl_𝔪`, and is determined by the unit relation; it is not data. The finite-order case is exactly the pin above.

(b) **`compatibility` is a theorem on `a ≡ 1 mod* 𝔪`** (#245's `IsCongrOne`), stated explicitly as `χ_u((a)) N(a)^shift = ∏_τ τ(a)^{n_τ}`, replacing the two opaque `grossenFullWeight`/`grossenArchimedeanFactor` defs. As specified before, over all `x : Kˣ`, the law is false for every ramified character: for the odd character mod `4·∞` of `ℚ` and `x = 3`, the ideal side is `χ((3)) = -1` and the archimedean factor of the zero infinity type is `1`.

(c) **`grossencharacterData` is the analytic card of the unitary part** (`grossencharacterData_completed : completed = unitaryCompletion`), not of the full completion. For `shift = σ ≠ 0` the partner in `HasFunctionalEquation` is the dual card, `conj ∘ completed ∘ conj`, i.e. `Λ_{conj χ}(1-s) = Λ_u^{χ_u⁻¹}(1-s-σ)`, whereas `completed χ s = Λ_u(s-σ)` reflects against the inverse, `Λ_u^{χ_u⁻¹}(1-s+σ)`; so the full completion is the completed function of no `AnalyticLFunctionData`, and the card handed to the zeros roadmap has to be the unitary one. `completed_one_sub` reads the polar divisors off the unitary cards at `s - shift`. Every `shift = 0` statement — `grossencharacterData_ofRayClassCharacter`, the mod-4/mod-5 tests, `angularGrossencharacterData` — is unchanged. The README's norm-twist examples `N^{iu}`, which are outside the algebraic carrier (`not_isAlgebraic_normCharacter`), are replaced by the integral norm powers `N^m`, which exercise the same `-m`-versus-`-2m` complex-place check.

Two supplier declarations would let the pins be stated more directly; neither is invented here:

- a prime idele `GlobalNumberFields.primeIdele v : IdeleGroup K`, with `ideleFiniteCoord_primeIdele`, `ideleInfiniteCoord_primeIdele` and `rayClassQuotient_primeIdele : rayClassQuotient 𝔪 (mk (primeIdele v)) = idealClass 𝔪 ⟨v.asIdeal, _⟩`. `unitaryWeight_apply` quantifies over the ideles with those coordinates instead; that pins the same value, since their classes differ by units at `v`, on which a character presented at `𝔪` is trivial.
- a characterization of `HeckeCharacter.unitaryPart` against `HeckeCharacter.shift` (`χ y = χ.unitaryPart y * ‖y‖^{±shift}`). #245 pins `unitaryPart` only by `norm_unitaryPart`, so the sign convention between its shift and the ideal-side `χ = χ_u N^shift` is fixed here by `lFunctionC_eq` and `completed_recenter` and needs one equation on the supplier's side.

Names consumed by #253 (`AnalyticLFunctionData` and its predicates, `heckeLFunctionC`, `completedHeckeLFunction`, `heckeData`, `PrimitiveRayClassCharacter`, `eulerCorrection`, `partialZeta_eq_sum_heckeLFunctionC`, `three_four_one_nonneg`, the `dedekindZeta*` exports) are unchanged in signature. Changed in this layer only, and consumed by nothing downstream: the fields of `Grossencharacter`; `Grossencharacter.shift` (field → def, same `χ.shift` spelling); `dualIdealLattice` (`AddSubgroup` → `Submodule ℤ`, coercions unchanged); the hypotheses of `Grossencharacter.completed_one_sub`; `grossencharacterData_completed`; `grossenFullWeight` and `grossenArchimedeanFactor` are gone.

**Verification.** Re-run today against #286's *current* head: in a dependency-complete stack (upstream `main` at `77488cc`, toolchain `v4.34.0-rc2`, Mathlib `2b1308ee`, plus #286 at `9acb529` and this branch), the full `lake build` is green — 9221 jobs, zero errors, only `declaration uses 'sorry'` warnings — so the nine consumed declarations are applied against the signatures #286 has now, not the ones it had when this was written. `git diff --check` passes. Dependencies: #192 and #245 (merged), #286 (open).

(Apologies for the two-week gap — the account was unavailable; the commits predate it.)

CBirkbeck (changes requested):
## Mathematical review: corrections required before approval

*AI-assisted review, posted by ChatGPT at Chris Birkbeck's request. Reviewed commit: `920429c4868d642d7cd8d37802c41ace2f840735`; the PR head was rechecked before posting. This reviews the normative roadmap and proposed statements, not completed proofs. I have not run a local Lean build.*

The separation of arithmetic Dirichlet series, number-field character carriers, lattice Poisson summation, and completed L-functions is sensible. The primitive/presented distinction and the finite Euler corrections should be retained. The following two statement-level issues need fixing, and the theta–Mellin construction needs a more explicit implementation route.

### 1. `NormalizationTranslation.hasFunctionalEquation_iff` drops a field of the predicate

**Location:** `TauCetiRoadmap/LFunctions/Suggested.lean`, `AnalyticLFunctionData.HasFunctionalEquation` and `NormalizationTranslation.hasFunctionalEquation_iff`.

`HasFunctionalEquation` requires all three of:

- a unit-modulus root number;
- reflection symmetry of `polarOrder`;
- the functional equation away from the polar loci.

The right-hand side of `hasFunctionalEquation_iff` contains the first and third requirements but not the second. There is no continuation hypothesis on this theorem from which the missing condition could be recovered.

**Counterexample:** take arithmetic weight zero, conductor one, root number one, completed function identically zero, coefficient zero equal to zero, and a declared polar divisor supported only at zero. The right-hand side holds: all the asserted value equations are `0 = 0`. The left-hand side fails because it requires `polarOrder 0 = polarOrder 1`. Such a card is allowed by the actual theorem's hypotheses; it is not required to satisfy `HasMeromorphicContinuation`.

**Requested change:** include the translated polar-divisor condition on the right-hand side:

```text
∀ s, arithmetic.polarOrder s =
  arithmetic.polarOrder (weight + 1 - conj s).
```

Then translate all three fields. Add a malformed-polar-divisor regression so the equivalence cannot again be weakened to the value equation alone. An alternative repair would need explicit sufficient hypotheses and a proof that they recover the omitted field.

### 2. The principal-ideal formula reverses the declared archimedean convention

**Location:** `Grossencharacter.infinityType_eq`, `unitaryWeight_apply`, `Grossencharacter.compatibility`, and README Layer 6.

The presentation pins `infinityType` to the archimedean restriction of the primary idele-class character, and pins the unitary ideal weight to its finite prime-idele values. With those conventions, triviality on a principal idele gives

```text
χ_f(a) · χ_∞(a) = 1,
```

so the finite principal-ideal value is the **inverse** of the archimedean value. The conclusion of `Grossencharacter.compatibility` instead uses the positive archimedean exponents. This discrepancy is already present for unitary characters, so it cannot be repaired merely by altering a real shift.

**Regression example:** over `ℚ(i)`, take the unramified unitary character whose archimedean component is `(z / conj z)^2`. It is trivial on the Gaussian units. For `a = 2 + i`, its finite principal-ideal value must be `((2+i)/(2-i))^(-2)`, whereas the currently displayed compatibility law gives `((2+i)/(2-i))^2`. These are unequal.

**Requested change:** choose and consistently enforce one convention. Either invert the archimedean expression in `compatibility`, or store the negative of the actual archimedean exponents as the ideal-theoretic infinity type and change the comparison field accordingly. Audit the norm-shift convention at the same time: the supplier's shift is specified idelically, while evaluating the idele norm at a prime uniformizer introduces the reciprocal ideal norm. Propagate the chosen convention through the recentering, inverse-character and angular-character statements.

Please include a nonreal angular-character acceptance example. Real-valued or finite-order examples alone do not detect the inversion error.

### 3. Expand the positive-unit-rank theta–Mellin construction

**Location:** README Layers 1, 3, 5 and 6; `FEPairWithLevel` and `gaussianTheta_mellin_normalization`.

This is a completeness request, not a claim that the intended functional equations are false. The existing ideal-lattice Poisson transport is a good starting point. But a one-parameter Euclidean Gaussian on an ideal lattice, Mellin-transformed in its radial variable, yields an Epstein-type series; it does not by itself supply the Dedekind/Hecke construction in positive unit rank.

Please make the intervening work explicit: the archimedean multi-parameter theta kernel; the action of units; a logarithmic fundamental domain or equivalent quotient construction; the quotient measure and regulator/root-of-unity constants; unfolding to the norm-indexed ideal sum; and the absolute convergence/dominated-convergence results licensing the sum–integral interchanges. For nontrivial infinity type, specify the angular or polynomial factors and their Fourier transforms. An existential `θ` with a Mellin equality is not a substitute for identifying this kernel and proving its properties.

A real-quadratic acceptance case should exercise the unit quotient, in addition to the rational and imaginary-quadratic normalization checks.

### Acceptance and dependency boundary

The first two items are required correctness fixes. The third makes the analytic route implementable without hiding the central number-field step in an opaque kernel. None calls for rebuilding generic Poisson summation here or importing Artin theory. After these changes, the existing dependency on #286 should be satisfied and the integrated hosted build checked separately from mathematical review.

roed-math:

All three points addressed in `e49d74f`.

**1. `hasFunctionalEquation_iff` translates all three fields.** The right-hand side now carries the translated reflection `∀ s, arithmetic.polarOrder s = arithmetic.polarOrder (weight + 1 - conj s)` between the root-number and value clauses (`w + 1 - conj s` is the image of `1 - conj p` under `p ↦ p + w/2`, and the value clause's hypotheses were already the translated polar loci). Your counterexample is in the file as `malformedPolarCard` — weight zero, conductor one, root number one, `completed = 0`, `polarOrder = single 0 1` — with three **closed** proofs: `malformedPolarCard_value_equation` (the root-number and value clauses hold, every value equation being `0 = 0`), `malformedPolarCard_not_polarOrder_reflect` (`polarOrder 0 = 1 ≠ 0 = polarOrder 1`), and `not_hasFunctionalEquation_malformedPolarCard`; `not_hasFunctionalEquation_of_malformedPolarCard` carries it through `NormalizationTranslation.of _ 0` via `eq_of_weight_zero`, so an equivalence weakened back to the value equation would contradict a closed proof. README Layer 0 lists it as a mandatory test.

**2. The archimedean sign, and the norm-shift audit.** Confirmed, and the convention is now fixed in one place and propagated. `Grossencharacter.compatibility` reads

```
χ_u((a)) · N(a)^shift · ∏_τ τ(a)^(n_τ) = 1        (a ≡ 1 mod* 𝔪),
```

with `compatibility_inv` the solved form; the docstring derives it from the three pins and cites Neukirch VII (6.13), where the archimedean component of the idele class character is `b ↦ b⁻¹` against Hecke's `χ_∞`, so Hecke's classical infinity type is `-n` and `infinityType` is kept as the *idelic* exponent (the primary object's restriction, as fixed last round). The nonreal test you asked for is `angularGrossencharacter_compatibility_test`: over `ℚ(i)` at `a = 2 + ζ₄`, `k ≠ 0`, the corrected law holds, the uninverted law fails, and the finite value has nonzero imaginary part — `((3+4i)/5)^(2k)` versus `((3+4i)/5)^(-2k)`, unequal because `(3+4i)/5` is not a root of unity. `angularGrossencharacter_unitaryWeight_span` makes the ideal-side values explicit through the embedding at which the infinity type has exponent `-2k`, and the definition's docstring now says the idelic component is `(z/|z|)^(-4k)`.

On the shift: the supplier pins `HeckeCharacter.shift` only through `shift_eq_zero_iff`, so its sign is fixed here by a single equation, `toHeckeCharacter_primeIdele`: at a prime idele, `χ(π_𝔭) = χ_u(𝔭) N𝔭^shift`, which is the ideal-side `χ = χ_u N^shift` of `lFunctionC_eq`/`completed_recenter` read at one prime. Since `‖π_𝔭‖ = N𝔭⁻¹`, that is `‖χ y‖ = ‖y‖^(-shift)` idelically — Tate's exponent is `-shift` — and the dependency table records `‖χ y‖ = ‖y‖^(-χ.shift)` as the one equation requested of #245. The recentering (`s - shift`), the inverse presentation (shift `-σ`) and the angular characters (shift `0`, sign as above) are all stated in that convention; the convention table has a new *archimedean sign* row.

**3. The positive-unit-rank construction is explicit.** Layers 1, 3, 5 and 6 now follow Neukirch VII §5 and §8 object by object, in `Suggested.lean` and the README, with the constants derived in this roadmap's normalization and checked against the two existing worked cases:

- *Layer 1*: the parameter space `R_+^*` and `N(y)` (`ArchParam`, `archNorm`); the multi-parameter Gaussian `archGaussian` with `𝓕 G_y = N(y)^(-1/2) G_{y⁻¹}` (`mixedFourier_archGaussian`); `latticeTheta` with its summability from the supplier's `summable_poisson_left` and its transformation `latticeTheta_inv` against the Euclidean dual `mixedDual` (of which `dualIdealLattice` is the instance, by `rfl`); the unit action `unitScale u : y ↦ |u|² y` (`archGaussian_unit_smul`, `archNorm_unitScale`, `unitScale_eq_self_iff` — kernel the torsion — and the invariance `latticeTheta_unitScale`); the norm-one hypersurface, `y = x t^(1/n)` (`surfaceScale`/`surfacePart`), `dy/y` (`archHaar`) and `d*x` pinned by `dy/y = d*x × dt/t` (`surfaceHaar`, `archHaar_eq_map`); the predicate `IsUnitFundamentalDomain` with Neukirch's `F` **taken from Mathlib** — the parameters whose square root lies in `NumberField.mixedEmbedding.fundamentalCone`, the same cone Mathlib uses to enumerate the ideals of a class — closed under inversion and translation on `S`; and (5.6), `vol(F) = 2^(r-1) R` (`surfaceHaar_of_isUnitFundamentalDomain`).
- *Layer 3*: the Epstein regression first — `radialMellin_eq_epsteinZeta` (the radial transform of the one-parameter theta is `π^(-s)Γ(s)` times a sum over lattice *points*) and `not_summable_absNorm_of_rank_pos` (in positive unit rank the corresponding sum over norms is not summable at all); `completedPartialZeta` pinned by (5.4) in two forms, one through Mathlib's cone (`_eq_tsum_fundamentalCone`), and linked to Layer 2; the unfolding on the cone over a fundamental domain (`completedPartialZeta_eq_integral_unitCone`) with the sum–integral interchange licensed by `summable_integral_unitCone`; the kernel `mellinKernel` of (5.5) with `Z(𝔎, s) = ∫ (f_D(𝔞,u) - a₀) u^(s/2) du/u` (`completedPartialZeta_eq_mellin`), `a₀ = 2^(r-1)R/w` (`mellinConstant`); (5.8) as `mellinKernel_sub_const_isBigO` and `mellinKernel_inv` — `f_D(L, 1/u) = √u f_{D⁻¹}(L^∨, u)` for every lattice — with `mellinKernel_dualIdealLattice` identifying the dual side as the ideal kernel of `(𝔞𝔡)⁻¹` up to the translation `traceShift` of the parameter; the Mellin principle on `FEPairWithLevel` (now with limits, decay, `completed_eq_mellin`, `completed_eq`, both residues — so `epsilon` and `completed` are pinned rather than decorative) and its instance `dedekindFEPair`; (5.9) as `completedPartialZeta_one_sub`(`_eventuallyEq`) and the residue `2a₀ = 2^r R/w`.
- *Constants*: two normalizations differ from Neukirch's and are audited — `mixedInner` counts a complex coordinate once, and `dy/y` is the product over places (Neukirch's carries `e_𝔭 = 2` at complex places). Hence `vol(F) = 2^(r-1)R/2^r₂` in this measure, the compensating `2^r₂` sits in the kernel, and `a₀ = 2^(r-1)R/w` in both. Checks: `ℚ` gives `a₀ = 1/2` and Riemann's `½∑e^{-πum²}`; `ℚ(i)` gives `a₀ = 1/4` and `2π^{-s}Γ(s)ζ_{ℚ(i)}(s)`, the constant of `completedDedekindZeta_cyclotomic_four` (`mellinConstant_rat`, `_cyclotomic_four`).
- *The real-quadratic acceptance case*, `realQuadratic_unitQuotient_test`: rank `1`, `w = 2`, `vol(F) = 2R`, `a₀ = R`, and residue `2R` of every partial zeta function at `s = 1`, summing over the `h` classes to the `2hR` of `completedDedekindZeta` — invisible over `ℚ` and `ℚ(i)`, and not produced by the one-parameter construction.
- *Layer 6, nontrivial infinity type*: Hecke's finite character is **derived** from the primary object, `χ_f(a) := χ_u((a)) N(a)^shift ∏τ(a)^(n_τ)` (`finiteCharacter`; trivial on `a ≡ 1 mod* 𝔪` by a closed proof from `compatibility`, multiplicative, zero exactly off `(a, 𝔪₀) = 1`, a function of `(𝓞/𝔪₀)ˣ × {±1}^𝔪∞`, extended to `K^(𝔪)` on the supplier's `primeToSubgroup`); the harmonic polynomial `harmonicFactor` — `x^(ε_w)` at real places, `conj(z)^(k_w)` or `z^(-k_w)` at complex places, the conjugate for positive `k_w` being the sign of point 2 — with `harmonicExponent`/`harmonicDegree` (`Tr p`); Hecke's lemma `mixedFourier_harmonicFactor_mul_archGaussian` with `(-i)^(Tr p)` and `y_w^(-P_w)`; the twisted theta `heckeTheta` over a fractional ideal; (8.2) as `archWeight_mul_heckeTheta_unitScale` from `finiteCharacter_mul_harmonicFactor_unit`; the kernel `heckeMellinKernel` of (8.3) with its three constants derived from the gamma integral of the weighted Gaussian (the `|z|^(-P)` cancels the `|a_w|^(P_w)` of `χ_f(a)N(a^p) = χ_u((a))∏|a_w|^(P_w)`, which is why the card's gamma shifts are `ε_w` and `|k_w|/2`); the per-class identity `unitaryPartialCompletion_eq_mellin` at the variable `(s + Tr p/n)/2`, the total kernel `heckeMellinTotal` (independent of the class representatives, `_congr`) with `unitaryCompletion_eq_mellin`, and (8.4) as `heckeMellinTotal_inv` with weight `1/2 + Tr p/n` and constant `rootNumber` — which is what now pins the root number — packaged as `grossencharacterFEPair`.
- *Layer 5*: `gaussSum` (6.3) as the finite sum over the residue units with (6.4) (`gaussSum_mul`, `norm_gaussSum`); `completedHeckeLFunction_eq_mellin` identifies the Layer 5 kernel as the finite-order case, and `exists_mellin_completedHeckeLFunction` is kept unchanged as the existential the zeros roadmap consumes by name, with its witness named in the docstring.

Nothing generic is rebuilt: the nine Theta Series declarations are still the only Poisson/Gaussian inputs, Mathlib supplies the cone, the regulator, the torsion order and the ideal enumeration, and no Artin material enters.

Names consumed by #253 are unchanged in signature. Changed here: `FEPairWithLevel` gained the fields `fLimit`, `gLimit`, `f_decay`, `g_decay` (no consumer), and the third conjunct of `gaussianTheta_mellin_normalization` is now the named-kernel statement instead of `∃ θ`.

**Verification.** In a dependency-complete checkout (upstream `main` at `77488cc`, toolchain `v4.34.0-rc2`, plus #286 at its current head `b29b21d`, against which the nine consumed declarations still elaborate unchanged): `lake build TauCetiRoadmap.LFunctions.Suggested` is green — zero errors, only `declaration uses 'sorry'` warnings — with the closed proofs above closed. `git diff --check` passes. GitHub's `build` stays red until #286 merges.

CBirkbeck (changes requested):
:robot: ChatGPT follow-up review at Chris's request, against `e49d74f5c965bbe63f8fa40a7fa117d6c22b735c`.

The polar-divisor clause of `NormalizationTranslation.hasFunctionalEquation_iff` is now translated, and the new positive-unit-rank construction names the parameter space, unit quotient, regulator normalization and Mellin kernels rather than hiding them in an existential. However, the newly explicit Hecke construction contains the following mathematical blockers.

## 1. The finite character omits the real sign that the harmonic polynomial uses

Take `χ = Grossencharacter.ofRayClassCharacter ℚ oddPrimitiveModFour.character` and `u = -1` in `finiteCharacter_mul_harmonicFactor_unit`.

The finite-order presentation has algebraic `infinityType = 0` and `shift = 0`. By the displayed definition of `finiteCharacter` (or `finiteCharacter_unit`), `finiteCharacter χ (-1) = 1`: the ideal generated by `-1` is the unit ideal. But `harmonicFactor` uses the actual real parity of the primary Hecke character, which is one for this character, so its value at `-1` is `-1`. The right-hand side is a product of absolute values of a unit, hence `1`. The theorem therefore asserts **`-1 = 1`**.

This also makes the proposed odd-character theta kernel vanish by pairing `a` and `-a`: the current finite coefficient is even and the harmonic polynomial is odd, although the completed odd Dirichlet L-function is not zero. The existing card/parity test does not exercise this construction.

**Required correction:** derive the finite residue character using the full archimedean character, including the real sign components, rather than only the integer-exponent product. Alternatively split off and insert the precise real sign correction everywhere it is needed. Keep the algebraic infinity-type convention; do not force finite-order odd characters to have a nonzero algebraic exponent. Add the `u=-1`, odd-mod-4 test directly to the unit-compensation and theta/Mellin contracts. This correction must also make the character used in the finite Gauss sum genuinely descend to the residue units; the current `finiteCharacter_eq_of_congr` only promises dependence on residue **and signs**, whereas `gaussSum` evaluates arbitrary `Quotient.out` representatives modulo the finite part alone.

## 2. `gaussSum_mul` has the inverse character wrong

The definition is `τ(χ,y) = Σ_x χ_f(x) exp(2πi Tr(xy))`. Reindexing by multiplication by a unit gives

```text
τ(χ,a*y) = χ_f(a)⁻¹ * τ(χ,y),
```

not `χ_f(a) * τ(χ,y)`. Mathlib already states precisely this convention as `gaussSum_mulShift_eq` in `Mathlib/NumberTheory/GaussSum.lean`.

An even primitive character of order three modulo 7 separates the two formulas without any real-sign complication: choose `χ_f(3)=ω`, `ω³=1`, `ω≠1`, and `y=1/7`. The nonzero Gauss sum at `3/7` is `ω⁻¹` times that at `1/7`, not `ω` times it. Quadratic examples do not detect this error. Either invert the factor or conjugate the character in the *definition* and consistently update the root-number conventions.

## 3. `gaussSum_mul` lacks its additive-character domain hypothesis

The docstring says `y ∈ 𝔪⁻¹𝔡⁻¹`, but the theorem takes an arbitrary `y : K`. Even after fixing the inverse, its nonunit branch remains false. Use your `evenPrimitiveModFive` over `ℚ`, set `a=5` and `y=1/25`. The conclusion gives

```text
τ(χ,1/5) = χ_f(5) * τ(χ,1/25) = 0,
```

while `norm_gaussSum` at `1/5` asserts its norm is `√5` (its displayed `hy` holds with `(y)𝔪𝔡=1`). This is an inconsistency between two target theorems, independent of any choice of quotient representatives. Put the annihilator/inverse-different membership in the type and state the primitivity conditions needed for the nonunit vanishing branch.

Please make these repairs in both the normative Layer 5/6 descriptions and the signatures, and propagate them through the Gauss-sum evaluation and root-number/theta transformation. The new explicit construction is useful precisely because it exposes these cases.

Separately, the current GitHub CI run **35628193211** fails and #286 is still open; the author's dependency-complete local build is not a passing upstream build. I have not submitted approval.

roed-math:

All three blockers confirmed and addressed in `3318689`.

**1. The finite character now carries the real sign.** You are right that `χ_u((a)) N(a)^shift ∏τ(a)^(n_τ)` is the residue character only when the parity of the archimedean restriction agrees with the parity of the algebraic exponent at every real place, which `realParity_eq` guarantees outside `𝔪∞` and nowhere else. The repair follows your first option: `archimedeanValue χ x` is the full archimedean value — the algebraic product times `sgn(τ_w x)^(ε_w - n_w)` at each real place — and `finiteCharacter a := χ_u((a)) N(a)^shift · archimedeanValue χ a`. The algebraic infinity-type convention is untouched (odd finite-order characters keep exponent `0`); `archimedeanValue_eq_of_isCongrOne` says the sign correction is trivial on `a ≡ 1 mod* 𝔪`, so `finiteCharacter_eq_one_of_isCongrOne` is still a closed proof from `compatibility`. Consequences propagated:

- `finiteCharacter_unit` now reads `χ_f(u) = archimedeanValue χ u`, and the docstring of `finiteCharacter_mul_harmonicFactor_unit` records that at `u = -1` for the odd character mod `4∞` both factors are `-1`.
- The descent is now to the residue units alone: `finiteCharacter_residue` replaces the former "residue and signs" statement — `Ideal.Quotient.mk 𝔪₀ a = Ideal.Quotient.mk 𝔪₀ b → χ_f(a) = χ_f(b)` — which is exactly what `gaussSum`'s `Quotient.out` needs, and `gaussSum_eq_sum` states independence of the representatives for `y ∈ 𝔪₀⁻¹𝔡⁻¹`.
- Your test is in the file as `oddCharacter_mod_four_sign_test`: `χ_f(-1) = -1`, `harmonicFactor(-1) = -1`, `Tr p = 1`, the unit compensation `(-1)(-1) = 1`, and the theta kernel over `ℤ` identified as Riemann's odd theta function `2 ∑_{n≥1} χ₄(n) n e^{-π y n²}` — the pairing of `a` with `-a` doubles rather than cancels. The docstring says explicitly that `oddCharacter_mod_four_test` could not see this, since it reads the parity off the card.

**2. `gaussSum_mul` has the conjugate.** `τ(χ_f, a y) = conj(χ_f(a)) τ(χ_f, y)`, from the reindexing `x ↦ x a⁻¹`; the docstring cites Mathlib's `gaussSum_mulShift_eq` and your order-3 mod 7 example. Because `finiteCharacter` already vanishes off `(a, 𝔪₀) = 1`, the single equation covers both cases of Neukirch (6.4).

**3. The domain hypothesis is in the type.** Both `gaussSum_mul` and `norm_gaussSum` take `y ∈ FractionalIdeal.dual ℤ ℚ 𝔪₀` (Mathlib's `(𝔪₀𝔡)⁻¹`); `norm_gaussSum` takes in addition an integral ideal `𝔟` prime to `𝔪₀` with `(y) 𝔪₀ 𝔡 = 𝔟`, replacing the earlier disjunction. Your consistency check is recorded in both docstrings: with the even character mod `5`, `y = 1/25` is excluded by the hypothesis, while `y = 1/5` gives `𝔟 = ℤ` and `|τ| = √5`.

README Layers 5 and 6 carry the same corrections and the two witnesses. The root-number remark now says that Neukirch's `τ(χ_f)` in `W(χ)` is the Gauss sum of (7.4) in ideal-number normalization, not `τ_𝔪(χ_f, y)` at a chosen `y`.

**On CI.** Agreed that the local dependency-complete build is not a passing upstream build: the GitHub `build` job fails on the `import TauCetiRoadmap.ThetaSeries.Suggested` that #286 will supply, and it will stay red until #286 merges, as the PR description says. Locally, with upstream `main` at `77488cc` and #286 at its current head `b29b21d`, `lake build TauCetiRoadmap.LFunctions.Suggested` is green at `3318689` — zero errors, `sorry` warnings only, the closed proofs closed; `git diff --check` passes.

PR #251, opened 2026-08-17
Add roadmap: Local Galois groups of p-adic fields

roed-math:
> **Dependencies:** This roadmap depends on Profinite Cohomology (#188), Local Fields and Ramification (#189), Profinite and Pro-p Groups (#244) and [Class Field Theory #250](https://github.com/TauCetiProject/TauCetiRoadmap/pull/250), all now merged. Every import resolves on `main`, so the `build` check runs for real on this PR.

## Summary

This roadmap applies foundational cohomology, ramification, reciprocity, and abstract pro-`p` theory to absolute Galois groups of `p`-adic fields. It covers the full-group rank theorem, the free-versus-Demushkin dichotomy for `G_K(p)`, the full local cyclotomic character, its descent to a pro-`p` orientation under the required roots-of-unity hypothesis (with a separate `p = 2` route), the `q`-invariant, marked local presentations, `G_{ℚ₂}(2) ≃ D₀`, and the `ℚ₂(√-2)` example.

The PR contains the normative `README.md` and representative target signatures in `Suggested.lean`; the root index files are regenerated by the sync bot after merge.

## Ownership

It consumes the abstract standard marked group `D₀` from Profinite and Pro-p Groups and owns the marked isomorphism from the relevant local Galois group to it. It consumes local field/ramification and local duality/Euler inputs without restating their foundational theorems.

## Port history

This is a clean port of [roed-math/TauCetiRoadmap#17](https://github.com/roed-math/TauCetiRoadmap/pull/17), rebased onto upstream `main`. The application layer was extracted from the former Local Fields and Pro-`p` roadmaps, removing their old dependency cycle while preserving the exact rank theorem, roots-of-unity cases, cyclotomic orientation, Labute-presentation interface, and named examples. The non-normative `PROVENANCE.md` is omitted; the detailed migration ledger remains private.

## Human review priorities

- the exact four-way supplier boundary;
- free versus Demushkin cases and the distinction between the full cyclotomic character and its hypothesis-dependent pro-`p` descent;
- the marked `G_{ℚ₂}(2) ≃ D₀` theorem and dyadic examples;
- the integral chain for the full absolute-Galois rank: the Tate module of a tame layer, its decomposition `Y ≃ M₀ ⊕ ℤ_p[G]^N` against the tame-frame module, the relation-module surjection onto `A(L)`, and the tuple/compactness passage to `G_K` (Layer 7, following the proof of NSW (7.4.1)).
- the finite-Galois-layer hypotheses of Layer 7 Step 3 and the two named rejection tests (the non-Galois cubic `ℚ₅(∛5)` against the rational decomposition, the identity pair on `ℚ₂(ζ₃)` against the sharp exponents);
- the two even-degree marked theorems, the `ℚ₂(√-2)` instance, and the `f`-pinning gap in the merged Profinite and Pro-p Groups supplier's marked theorems that this round found (see the round comment).

## Validation

- `lake build TauCetiRoadmap.LocalGaloisGroups.Suggested` at `ba3eb7a` (toolchain `v4.34.0-rc2`): green, 8895 jobs, only `declaration uses 'sorry'` warnings; the full `lake build` with `main` merged in was green at the previous head (9206 jobs)
- `git diff --check`

The roadmap and restructuring were prepared with substantial assistance from Claude Fable 5.1 and Opus 5, and GPT-5.6 Codex and Pro, under the author's direction. A detailed migration and coordination ledger is maintained privately. No external source code was copied into this roadmap.


CBirkbeck (commented):

**Verdict: no mathematical changes requested. Keep it stacked behind #188, #189, #244 and #250.**

The roadmap states the free/Demushkin dichotomy, the `p = 2` marked cases, the roots-of-unity parameter and the full absolute-Galois generator rank separately, each with exact supplier contracts.

The implementation should preserve the explicit Jarden–Shusterman / Labute hypotheses and commutator conventions, but I found no independent correction to request.
---

```text
```

CBirkbeck (commented):

# Deep second review: PR #251

**Roadmap:** Local Galois groups of p-adic fields

**Head reviewed:** `48750e6a2b4e83ad7b1cc2d407903658c3a47c95`

# PR #251 — Local Galois groups of p-adic fields

## Verdict

**No new mathematical changes requested.** The application layer is now cleanly separated from the abstract pro-`p`, ramification and class-field-theory suppliers.

## What I checked

### 1. Full absolute-Galois rank versus pro-p rank

The roadmap distinguishes:

- the minimal number of generators of the full profinite `G_K`;
- the generator rank of the maximal pro-`p` quotient.

The formulas retain the roots-of-unity correction and are not conflated.

### 2. Free/Demushkin dichotomy

For finite extensions of `ℚ_p`:

- when the relevant `p`-power roots of unity are absent, the maximal pro-`p` quotient is free of the correct rank;
- when the dualizing root is present, it is Demushkin.

The `p=2` branch is handled separately, since `μ₂` is always present and the classification has extra cases.

### 3. Cyclotomic character

The full local cyclotomic character is first defined on `G_K`. Its descent to the maximal pro-`p` quotient is proved only when the prime-to-`p` part is killed, with a separate dyadic argument.

The Demushkin orientation is compared with the chosen cyclotomic convention rather than silently identified.

### 4. The q-invariant

The invariant is tied to the largest `p`-power root-of-unity subgroup and to the orientation. Infinite and small dyadic cases are kept distinct.

### 5. Marked presentations

The roadmap consumes the marked abstract Labute classification and constructs an isomorphism carrying chosen local generators to the standard presentation. It does not settle for an abstract group isomorphism.

### 6. `G_{ℚ₂}(2)`

The standard group `D₀`, its relator and orientation are fixed upstream. This roadmap owns the arithmetic identification and the images of the marked generators.

### 7. The `ℚ₂(√-2)` example

The example checks a genuinely different dyadic root-of-unity and orientation case, rather than merely renaming the `ℚ₂` theorem.

### 8. Inputs

The rank theorem consumes:

- local unit filtration and ramification;
- local reciprocity;
- local duality and Euler characteristic;
- abstract Frattini/Hopf/Demushkin theory.

No supplier is duplicated.

## Lean-facing audit

The crucial cross-PR tests after rebasing are:

1. the precise `H¹` and `H²` coefficient objects from #188/#250;
2. the orientation direction, including inversion conventions;
3. finite versus cardinal generator ranks;
4. the `p=2` presentation branches;
5. the exact marked generator tuple.

The roadmap already calls these out.

## Sources checked

- Neukirch–Schmidt–Wingberg, *Cohomology of Number Fields*.
- Jannsen–Wingberg presentations of local absolute Galois groups.
- J. Labute, Demushkin group classification.
- J.-P. Serre, *Galois Cohomology*.

## Disposition

**Approve after #188, #189, #244, and corrected #250.**

roed-math:

No review request outstanding — a build fix, pushed as [`2d66a1a`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/251/commits/2d66a1a), found by building the whole portfolio in one dependency-complete checkout. No statement of this roadmap changes.

The §*Closed checks on the arithmetic supplier contract* section exists so that a change in a supplier breaks this file rather than silently creating a replacement. All three of its checks had stopped elaborating, so none of them was checking anything.

**1. `LocalFieldsRamification.card_powerClasses_mixed` had moved.** It now carries `[ValuativeExtension ℚ_[p] K]`, and it takes the nonvanishing proof `(n : K) ≠ 0` as an *argument*, because its right-hand side ends in `natCastValuation K n`, which needs it. The check also stated that last factor as `Nat.card (𝒪[K] ⧸ (n))` — a second expression for the same number, which is exactly what a closed check must not do, since it survives the supplier change it is meant to catch. It now uses the supplier's own right-hand side, `#𝓀[K] ^ v_K(n)`.

**2. The two Class Field Theory checks needed universe 0.** #250 pins every cohomological object to universe `0` on purpose — Mathlib's `tateCohomology` needs the group and `ℤ` in one universe — so `ClassFieldTheory.H`, `muNRep`, `kummerEquiv_mixed` and `h2MuEquivZMod_mixed` take a `Type` and cannot be applied to this file's `K : Type u`. They now live in their own section over `F : Type`, with a comment saying why. The rest of the roadmap stays at `Type u`; only the statements that mention a class-field object are restricted.

**Verification.** In the dependency-complete stack, `lake build TauCetiRoadmap.LocalGaloisGroups.Suggested` is green. `git diff --check` passes.

**Merge order.** Unchanged: #244 → #189 → #250 → this.

CBirkbeck (changes requested):
## Verdict

**Request changes.**

The Demushkin classification of `G_K(p)` is well motivated and mostly has the right case split. The main unresolved issue is the final rank theorem for the full absolute Galois group: the rational relation-module calculation does not by itself give the integral/topological generator bound claimed.

## 1. Expand the full absolute-Galois rank proof

The roadmap uses an Iwasawa relation module with a rational decomposition of the form

\[
A(L)\otimes\mathbf Q_p
 \cong \mathbf Q_p[G]^N\oplus\mathbf Q_p.
\]

To deduce a minimal number of topological generators, one still needs:

- an integral `Z_p[G]` lattice in the relation module;
- control of its torsion;
- coinvariants and invariants under the finite quotient;
- a relative Burnside theorem;
- passage from finite quotients to the profinite group;
- matching lower and upper bounds.

A rational representation-theoretic equality cannot determine the minimum number of integral generators on its own. Each of these steps should be a named milestone.

## 2. Make roots of unity an actual finite subgroup

The carrier for `pPowerRootsOfUnity` should be a subgroup of `K×`, closed under multiplication and inverse, not merely a subtype with a later cardinality.

Avoid using `Nat.card` before finiteness is proved; on an infinite type it silently gives the wrong sort of total value. Define the finite subgroup first, prove finiteness, then define `q(K)` from its order.

## 3. The cyclotomic orientation must be imported and compared exactly

The marked Demushkin presentation depends on the orientation character and on the normalization of local reciprocity. The roadmap should type-check the exact comparison among:

- the Galois cyclotomic character;
- the orientation extracted from the dualizing module;
- the reciprocity image of units and a uniformizer;
- the chosen arithmetic Frobenius convention.

This is where an inverse convention changes the marked relator.

## 4. State the dyadic branches with mutually exclusive hypotheses

For `p=2`, the cases involving `q=2`, the presence of fourth roots of unity, and parity of the degree should be encoded as predicates which Lean can prove disjoint and exhaustive. A prose table is not enough for the marked classification theorem.

## 5. Keep abstract and arithmetic presentations separate

The abstract Labute theorem belongs to #244. This roadmap should only:

- verify that `G_K(p)` is Demushkin;
- compute rank, orientation and `q`;
- apply the imported marked classification.

Do not restate or locally specialize the presentation theorem in a way which creates a second normal form.

## What is good

The separation between the maximal pro-`p` quotient and the full absolute Galois group, the mixed/equal-characteristic distinction, and the explicit `Q₂` marked case are all worthwhile.

## Recommendation

The roadmap needs a full integral relation-module chain for the absolute rank theorem and a more robust roots-of-unity/orientation API. After that the local Demushkin application should be strong.

CBirkbeck (approved):
Current head: `7052beba8f9ef0e4a7835ff3bce6bd22ba029dc6`

## Verdict

**Approve after the profinite, local-field, cohomology and class-field-theory suppliers land.**

The former gap in the rank of the full absolute Galois group has been repaired. The roadmap no longer attempts to derive an integral generator count from a rational module decomposition alone.

## What I checked

The arithmetic invariants are now genuine objects:

- the $p$-power roots of unity form a finite subgroup of $K^\times$;
- `localRootOfUnityOrder` is defined from that subgroup;
- the cyclotomic orientation has a defining equation and is compared with arithmetic reciprocity;
- the $q=2$ branches are selected by the image of the orientation, not by $q$ alone.

The maximal pro-$p$ quotient is entirely supplied by Profinite and Pro-$p$ Groups. The roadmap adds only the local arithmetic calculations:

- free versus Demushkin;
- generator and relation ranks;
- the five marked presentation branches;
- the marked $\mathbf Q_2$ theorem preserving the values $(-1,1,(-3)^{-1})$.

The proof of
$$
d(G_K)=[K:\mathbf Q_p]+2
$$
is now separated into the necessary integral steps:

1. the integral $p$-adic completion of the multiplicative group;
2. its torsion and rational $\mathbf Q_p[G]$-decomposition;
3. Krull–Schmidt cancellation and projective detection over $\mathbf Z_p[G]$;
4. the tame two-generator frame and relative Frattini argument;
5. compatible generators in finite quotients and compactness;
6. the matching lower bound, with the roots-of-unity cases separated.

The roadmap explicitly warns that $[K:\mathbf Q_p]+1$ is the rank of the free maximal pro-$p$ quotient in the no-roots case, not the rank of the full absolute group.

I found no remaining mathematical correction. This is a large project, but the dependencies are now named at theorem level rather than hidden.

CBirkbeck (changes requested):
Request changes.

There is one false theorem statement in Layer 7, followed by two missing links in the integral
argument.

1. `exists_finset_card_generating_quotient` is false as stated.

It asks, for every relevant open normal subgroup `U`, for a `Finset` in `G_K/U` with cardinality
exactly

    [K : Q_p] + 2.

Take `U` to be the whole absolute Galois group. It is open and normal and contains the wild-inertia
commutator, but the quotient is the trivial group. Its underlying type has only one element, while
`[K : Q_p] + 2 ≥ 3`, so no such finset exists.

The statement should use

    s.card ≤ [K : Q_p] + 2

or, preferably for the inverse-limit argument, a generating tuple

    x : Fin ([K : Q_p] + 2) → G_K/U,

where repetitions are allowed.

2. `padicCompletionUnits_quotient_torsion_equiv` does not state the result described in the
README.

The target is only an additive equivalence with `Fin (N + 1) → Z_p`. This does not prove that the
quotient is a free `Z_p`-module. Please define the torsion as a `Z_p`-submodule and state at least a

    LinearEquiv Z_p

with `Z_p^(N+1)`. If the topology of this decomposition is used, its topological compatibility
should also be included.

3. The application of integral cancellation is missing.

The target gives the rational `Q_p[G]`-decomposition and then two generic cancellation results over
`Z_p[G]`, but it never states the stable integral isomorphism, projectivity, reduction-mod-p
comparison, or resulting integral relation-module decomposition for `A(L)`. It then jumps directly
to the finite-quotient generator theorem.

Please expose the actual intermediate theorem to which
`linearEquiv_of_prod_linearEquiv` or
`linearEquiv_of_projective_of_reduction`
is applied, and make the finite-quotient generation theorem explicitly consume it.

For the passage to `G_K`, I suggest formulating the proof in terms of tuples. For each open normal
`U`, let `X_U ⊆ G_K^(N+2)` be the closed set of tuples whose image generates `G_K/U`. Nonemptiness
comes from the corrected finite-quotient theorem, and the finite-intersection property comes from
replacing finitely many `U`s by their intersection. Compactness then supplies one tuple working in
every quotient. This does not require independently chosen generating sets to be compatible.

roed-math:

Addressed in `2e0c7ee`. The three points are all in Layer 7; I also went back to NSW's actual proof of (7.4.1) (VII §4, pp. 405–406 of the electronic 2nd edition) and rebuilt Step 3 around it, since the previous text named the cancellation lemmas without the theorem they are applied to.

**1. `exists_finset_card_generating_quotient` — false as stated, confirmed.** With `U = G_K` the quotient is a subsingleton, so `s.card ≤ 1 < N + 2`. Replaced by `exists_generating_tuple_quotient`, which asks for `x : Fin (N + 2) → G_K ⧸ U` with `Subgroup.closure (Set.range x) = ⊤`, repetitions allowed. The witness is recorded as a closed proof, `not_forall_exists_finset_card_generating_quotient` (`⟨⊤, inferInstance⟩`, `Finset.card_le_one_of_subsingleton`, `omega`), so the cardinality-equality form cannot come back. The limit statement `exists_finset_generating_mod_wildInertiaCommutator` had the same shape and is now `exists_tuple_generating_mod_wildInertiaCommutator`; the public `rank_absoluteGaloisGroup` keeps its `IsLeast` form over finsets, which is true and is what `topologicalGeneratorRankNat` measures.

**2. Torsion as a `ℤ_p`-submodule.** Mathlib's vocabulary is `Submodule.torsion ℤ_[p] M` (elements killed by a non-zero-divisor). Added `mem_torsion_padicCompletionUnits_iff`, pinning it to the group torsion `CommGroup.torsion` that `torsion_padicCompletionUnits` computes, and replaced the additive equivalence by `padicCompletionUnits_quotient_torsion_linearEquiv : Nonempty ((Additive A(L) ⧸ Submodule.torsion ℤ_[p] _) ≃ₗ[ℤ_[p]] (Fin ([L:ℚ_p] + 1) → ℤ_[p]))`. Also added the instance target `padicCompletionUnits_isScalarTower : IsScalarTower ℤ_[p] ℤ_p[G] A(L)`; without it the `ℤ_p`-torsion cannot be read inside the group-algebra module and the rationalization carries no `ℤ_p[G]`-structure. On topology: the decomposition is purely algebraic at every finite layer — nothing in Steps 3–5 uses a topology on `A(L)` or its free quotient; the only topological statement about `A(L)` is the reciprocity identification of Step 1, and the topology of `G_K` enters only through the compactness of Step 5. The README and the docstring now say so explicitly rather than adding a homeomorphism nothing consumes.

**3. The missing application of integral cancellation.** NSW's proof of (7.4.1) compares two modules: the Tate module `Y = I_{G_K}/I_{G_L} I_{G_K}` of a finite tame layer, an extension `0 → A(L) → Y → I_G → 0`, and the tame-frame module `M₀ = ℤ_p[G]²/(σ - a, τ - b)`; the isomorphism `(∗∗)`, `Y ≅ M₀ ⊕ ℤ_p[G]^N`, is what (5.6.11) — hence (5.6.10)(i) and (ii), not (iii) — is applied to. The new Step 3 names each object:

- `augmentationIdeal`, `relationModule` (the kernel of `e_i ↦ g_i − 1`, i.e. `R^ab(p)` by Lyndon's sequence (5.6.6)) and `tameFrameModule p G σ τ a b`, all data.
- `exists_tameFrame_exponents`: for `τ` of order prime to `p` there are exponents `a, b` with `ℤ_p[G]/(σ − a, τ − b)` of order exactly `q(L)`. This sharpness is NSW's exact sequence `(∗)` and is not automatic: `a = b = 1` gives an infinite quotient and a tame-frame module with the wrong torsion, so the exponents are a hypothesis of everything below, not a free choice.
- `tameFrameModule_tensorRat_linearEquiv` (`M₀ ⊗ ℚ_p ≅ ℚ_p[G]`) and `tameFrameModule_torsion_linearEquiv` (the `p`-power torsion of `M₀` is `μ_{p^∞}(L)` as a `ℤ_p[G]`-module, which is `E¹(M₀) = μ_{p^∞}(L)^∨`, NSW's `M₀ ≃ D(μ^∨)`).
- `TateModule p L K`: a f.g. `ℤ_p[G]`-module of projective dimension ≤ 1 with the exact sequence `0 → A(L) → Y → I_G → 0`; `nonempty_tateModule` is Tate's theorem (3.1.5) for the fundamental class, consuming CFT's `ClassFormation`/`fundamentalClass`/`tateTheorem` and the reciprocity identification of Step 1.
- `tateModule_linearEquiv`: `Y ≃ₗ[ℤ_p[G]] M₀ × (Fin N → ℤ_p[G])`. The docstring is the proof: `exists_projective_prod_linearEquiv_of_torsion` gives the stable isomorphism `Y ⊕ P ≅ M₀ ⊕ Q` because both have projective dimension ≤ 1 and torsion `μ_{p^∞}(L)`; the rational decomposition plus `I_G ⊗ ℚ ⊕ ℚ ≅ ℚ_p[G]` gives the rational identity; `linearEquiv_prod_free_of_stable` concludes.
- `exists_relationModule_surjective`: for any generating family `g` of size `N + 2`, a surjection `β : R^ab_{N+2}(p) ↠ A(L)` with `ker β ≃ ℤ_p[G]` — the integral relation-module decomposition of (7.4.2)(i) at a finite layer.

The cancellation library gained the two theorems the chain actually uses, `linearEquiv_of_projective_of_tensorRat` ((5.6.10)(ii): projectives are detected rationally) and `linearEquiv_prod_free_of_stable` ((5.6.11) in finite-group form), plus `exists_projective_prod_linearEquiv_of_torsion` ((5.4.11) with (5.6.9): on modules of projective dimension ≤ 1 the transpose is `E¹`, which is the dual of the `p`-power torsion, so isomorphic torsion gives a stable isomorphism). `linearEquiv_of_projective_of_reduction` ((5.6.10)(iii)) stays as library but is no longer claimed to be on the path. One deliberate deviation from Mathlib vocabulary: the `p`-power torsion of a `ℤ_p[G]`-module *as a `ℤ_p[G]`-submodule* is a spelled-out `pPowerTorsion` (membership `∃ n, (p^n : ℤ_p) • x = 0`, definitionally), because `Submodule.torsion`/`torsionBy` require a commutative base ring and `ℤ_p[G]` is not one; the `ℤ_p`-level torsion of point 2 uses Mathlib's `Submodule.torsion ℤ_[p]`. Rationalization is `M ⊗[ℤ_[p]] ℚ_[p]` with Mathlib's left-factor `ℤ_p[G]`-structure throughout; the sorry-bodied `ℚ_p[G]`-instance `padicCompletionUnitsRatModule` and `trivialRatModule` are gone.

`exists_generating_tuple_quotient` consumes this explicitly: its docstring is the proof (frame from `exists_tameFrame_quotient`, exponents, `nonempty_tateModule`, `tateModule_linearEquiv`, `exists_relationModule_surjective`, then the lift of `β` to group extensions via its `H²`-isomorphism and `scd = 2`, (3.6.4)(iii)). The `H²`-compatibility of `β` is a named milestone in the README, stated against the suppliers' universe-0 cohomology rather than a second carrier.

For the passage to `G_K`, the tuple formulation you suggested is now the file's: `generatingTuples p K U` is the set `X_U ⊆ G_K^{N+2}`, with `isClosed_generatingTuples`, `generatingTuples_nonempty` (from the finite level), `generatingTuples_antitone` (the finite-intersection step, replacing finitely many `U` by their intersection), `iInter_generatingTuples_nonempty` (compactness), and `exists_tuple_generating_mod_wildInertiaCommutator`; Step 4's Frattini reduction and `topologicalGeneratorRankNat_absoluteGaloisGroup_le` are unchanged. Also added `exists_tameFrame_quotient` (Step 4): every quotient through which `P_K` dies is generated by the images `σ, τ` of the frame with `¬ p ∣ orderOf τ`, which is the hypothesis `exists_tameFrame_exponents` needs.

**NSW numbering** was checked against the electronic 2nd edition: (7.4.1) (generation by `n + 2`), (7.4.2)(i) (`0 → ℤ_p[[G]] → R^ab_{n+2}(p) → X → 0`), (7.4.4)(i), (5.6.5), (5.6.6), (5.6.9), (5.6.10)(i)–(iii), (5.6.11), (5.4.2), (5.4.11), (3.1.5), (3.9.1), (7.5.11) are all as cited; Jarden–Shusterman Theorem 2.1 cites exactly (7.4.1) for the upper bound. The README's references entry now lists them.

**Supplier drift.** #250's head moved (`localWeilArtinEquiv`, `mem_range_artinMap_iff` re-typed; `localArtinMap_quadratic_eq_hilbertSymbol` takes a primitive root). This roadmap's closed checks against `artinMap`, `H`, `muNRep`, `kummerEquiv_mixed`, `h2MuEquivZMod_mixed` still elaborate; nothing here used the re-typed declarations.

**Verification.** Since this was written, #250 has merged, so the branch now builds standalone: with `main` merged in (toolchain `v4.34.0-rc2`, Mathlib `2b1308ee`), the full `lake build` is green — 9206 jobs, zero errors, only `declaration uses 'sorry'` warnings — and the closed checks against `artinMap`, `H`, `muNRep`, `kummerEquiv_mixed` and `h2MuEquivZMod_mixed` elaborate against the *merged* Class Field Theory. `git diff --check` clean. No roadmap on `main` imports this one.

(Apologies for the two-week gap — the account was unavailable; the commit itself predates it.)

CBirkbeck (changes requested):
## Mathematical review: repair the finite-layer module contracts

*AI-assisted review, posted by ChatGPT at Chris Birkbeck's request. Reviewed commit: `f8cd75247fa7c4cd2fe6899709541565e7fbd0fb`; the PR head was rechecked before posting. This is a review of the roadmap and admitted target statements, not a local Lean build or verification of completed proofs.*

The overall strategy should be retained: distinguish the full absolute Galois group from its maximal pro-p quotient, compute the free/Demushkin cases without using finite generation of the full group circularly, and use an integral module argument for the full-group rank. However, two helpers in that integral argument are false with their current hypotheses. A green elaboration/build does not detect this while their proofs are admitted.

### 1. The normal-basis decomposition is missing the finite Galois-layer hypotheses

**Location:** `TauCetiRoadmap/LocalGaloisGroups/Suggested.lean`, `padicCompletionUnits_tensor_ratPadic`, with the related `tateModule_linearEquiv` and `exists_relationModule_surjective` contracts.

The intended representation-theoretic statement is

```text
A(L) ⊗_{ℤ_p} ℚ_p ≃ ℚ_p[Gal(L/K)]^N ⊕ ℚ_p,
N = [K : ℚ_p].
```

The actual statement uses the automorphism group `L ≃ₐ[K] L` but does not require `L/K` to be Galois. Finiteness of that automorphism group is not a replacement for the Galois hypothesis. The displayed scalar structures must also form the intended compatible tower.

**Counterexample:** let `p = 5`, `K = ℚ_5`, and `L = ℚ_5(∛5)`. The polynomial `X^3 - 5` is Eisenstein, so the extension has degree three. It is not Galois and its K-automorphism group is trivial: a nontrivial conjugate would require a primitive cube root of unity, whose degree-two extension of `ℚ_5` cannot lie in this degree-three extension.

The left-hand side has `ℚ_5`-dimension `[L:ℚ_5] + 1 = 4`, by the unit-completion/free-quotient statement already included in this roadmap. The right-hand side has dimension `N · #Aut_K(L) + 1 = 1 · 1 + 1 = 2`. Thus the proposed isomorphism cannot exist.

**Requested change:** impose the finite Galois extension and compatible scalar-tower hypotheses on the regular-representation decomposition, and audit their propagation through the Tate-module and relation-module statements. State separately any genuinely more general version; do not read a prose reference to a “Galois layer” as a hypothesis that Lean automatically retains. Add the non-Galois cubic as a rejection test.

### 2. `exists_tameFrame_exponents` takes arbitrary elements, not a generating tame frame

**Location:** the named theorem in Layer 7 Step 3, and the corresponding README statement.

The theorem currently takes arbitrary `σ τ : L ≃ₐ[K] L`, with only the condition that the order of τ is prime to p, and concludes that exponents a,b can be chosen with the required roots-of-unity action and

```text
#(ℤ_p[G] / (σ-a, τ-b)) = q(L).
```

There is no hypothesis that σ and τ generate G. This omission is fatal even for a finite tame Galois extension.

**Counterexample:** take p=2 and the unramified quadratic extension `L/ℚ_2`, so `G=C₂` and `q(L)=2`. Put `σ=τ=1`. Then τ has order one, hence order prime to two. Since -1 belongs to the roots-of-unity group, the prescribed actions force a and b both to be odd. Therefore

```text
(1-a, 1-b) ⊆ 2 ℤ_2[C₂].
```

The proposed quotient surjects onto `𝔽₂[C₂]`, which has four elements. It cannot have cardinality two. If the quotient is infinite, `Nat.card` is zero, which also cannot satisfy the claimed cardinality.

**Requested change:** formulate this for an actual generating tame frame, with the finite-layer assumptions its proof consumes. The later `exists_tameFrame_quotient` already supplies a generating frame, so this is a repair to an overgeneralized helper, not a change to the main rank strategy. Please also check the action/dual convention in the annihilator identification; cardinality and the required module action must both be pinned. Add the identity-pair/unramified-quadratic rejection test.

### 3. The even dyadic “marked” target currently forgets the marking

**Location:** `absoluteGaloisGroupProP_two_marked_of_degree_even`, README Layer 6.

The theorem advertised as the marked even-degree normal form concludes existence of parameters a,f and a topological group isomorphism, but does not record the transported cyclotomic orientation or the equations identifying those parameters with the chosen orientation-image branch. An unmarked `Nonempty (... ≃ₜ* ...)` does not discharge the marked contract in the README.

Please expose the orientation equality (or the full generator-value equations against the supplier's standard orientation) and connect a,f to the `unitsPlusMinus`/`procyclicClosure` image calculation. Retain the unmarked isomorphism as a corollary. The `ℚ₂(√-2)` acceptance example should exercise this connection, rather than only the abstract isomorphism type.

### What should remain unchanged

The finite-generation-before-rank discipline, the explicit roots-of-unity finiteness gate, the separation of degree-two inflation injectivity and surjectivity, and the distinction between the full cyclotomic character and its hypothesis-dependent pro-p descent are all good. The present objections concern the finite-layer helper contracts and the marked endpoint; they do not call for abandoning the full-group rank theorem or the free/Demushkin classification.

roed-math:

Addressed in [`ba3eb7a`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/251/commits/ba3eb7ad6aaec73853da893421f2e5eebbd240d0).

**1. The Galois-layer hypotheses — confirmed, and on `tateModule_linearEquiv` the defect was falsity, not vacuity.** Step 3 now has a `section FiniteGaloisLayer` binding `[IsScalarTower ℚ_[p] K L] [IsGalois K L] [FiniteDimensional K L]` (with `Module.Finite ℚ_[p] K` for convenience) for `padicCompletionUnits_tensor_ratPadic`, the three frame theorems, `nonempty_tateModule`, `tateModule_linearEquiv` and `exists_relationModule_surjective`; the `TateModule` structure was moved above the section so that it does not absorb the hypotheses into its type. `[Finite (L ≃ₐ[K] L)]` is gone from all of them — it follows from `FiniteDimensional K L` through `AlgEquiv.fintype`. The rejection test is stated abstractly and instantiated: `not_nonempty_tensor_ratPadic_of_card_lt` says that `#Aut_K(L) < [L : K]` refutes the decomposition, by the dimension count `[L : ℚ_p] + 1` (Step 2) against `N · #Aut_K(L) + 1`; `NonGaloisCubic := AdjoinRoot (X³ - 5)` over `ℚ₅` carries `finrank_nonGaloisCubic` (closed, `= 3`), `nonGaloisCubic_card_algEquiv_lt`, `not_isGalois_nonGaloisCubic`, and `not_nonempty_tensor_ratPadic_nonGaloisCubic`, a closed proof applying the abstract test to the instance. Your dimension count `4` against `2` is in both docstrings. One thing worth recording: on the cubic a `TateModule` *exists* — `G` is trivial, `I_G = 0`, and `A(L) ≅ ℤ₅⁴ ⊕ (torsion)` has projective dimension one over `ℤ₅` — so `tateModule_linearEquiv` without the hypotheses asserted an isomorphism between modules of ranks `4` and `2` and was false, not merely unapplicable.

**2. `exists_tameFrame_exponents` — confirmed; generation is what makes the quotient cyclic.** The theorem now takes `Subgroup.closure {σ, τ} = ⊤`, which is exactly what `exists_tameFrame_quotient` supplies at the layer, and the docstring gives the mechanism: with `σ, τ` generating, `ℤ_p[G]/J` is `ℤ_p/(χ(r) − 1 : r ∈ R)` for the character `x ↦ a, y ↦ b` of the free group on two letters, it maps onto `ℤ/q(L)` because `a, b` lift the action, and moving `b` by a multiple of `q(L)` along `τ^{orderOf τ} = 1` — possible because `p ∤ orderOf τ` — makes `v_p(b^{orderOf τ} − 1)` exactly `v_p(q(L))`. With `⟨σ, τ⟩ = H < G` the quotient is `(ℤ_p[H]/J_H)^{[G:H]}`, of order at least `q(L)^{[G:H]}`, which is your identity-pair computation in general. On the convention: you were right to ask. The docstring said `J` was the annihilator of `μ_{p^∞}(L)^∨`; with the contragredient action on the dual that ideal is `(σ − a⁻¹, τ − b⁻¹)`. `J` is the annihilator of `μ_{p^∞}(L)` *itself* — the kernel of `ℤ_p[G] → ℤ/q(L) = End(μ)` — and the dual belongs to `tameFrameModule_torsion_linearEquiv` (`E¹(M₀) = μ^∨` through `a⁻¹, b⁻¹`), which is where it was already stated. Both files now say so, and the module action is pinned as a fourth conjunct: `ℤ_p[G]/J ≃ₗ[ℤ_p[G]] pPowerTorsion (A(L))`, the same carrier `tameFrameModule_torsion_linearEquiv` compares against. Rejection test: `not_exists_tameFrame_exponents_one_one` (abstract: `p = 2`, `q(L) = 2`, `Nontrivial (L ≃ₐ[K] L)`, the pair `σ = τ = 1`, which does satisfy `¬ 2 ∣ orderOf τ`), instantiated on `UnramifiedQuadratic := AdjoinRoot (X² + X + 1)` over `ℚ₂` with `localRootOfUnityOrder_two_unramifiedQuadratic`, `nontrivial_algEquiv_unramifiedQuadratic`, and the closed application `not_exists_tameFrame_exponents_one_one_unramifiedQuadratic`. The downstream theorems keep the cardinality form of sharpness as their hypothesis, since that is what they consume; the abstract test refutes the three-conjunct form, which is the stronger statement.

**3. The even marked target now carries its marking.** Split by branch, each taking the image equation as a hypothesis so the supplier's parameters are pinned by the arithmetic rather than re-chosen, and each recording the supplier's generator values against `cyclotomicOrientation` exactly as the odd theorem does:

- `absoluteGaloisGroupProP_two_marked_of_degree_even_plusMinus`: hypothesis `range = unitsPlusMinus f`; relator at `a = 0`; values `χ(x₂) = −1`, `χ(x₄)(1 − 2^f) = 1`, `1` elsewhere.
- `absoluteGaloisGroupProP_two_marked_of_degree_even_principal`: hypotheses `range = procyclicClosure u` and `(u : ℤ₂) = −1 + 2^k`; relator at `a = 2^k` and any `f > k`; values `χ(x₂)(1 + 2^k) = −1`, `χ(x₄)(1 − 2^f) = 1`, `1` elsewhere. `a` is pinned because `χ(x₂)(1 + a) = −1` is solvable in `U^[k]` exactly when `v₂(a) = k`. `f` is genuinely not an invariant in this branch: `(1 − 2^f)⁻¹ ∈ U^(f) ⊆ U^[k]` for every `f > k`, and the presented groups for the different `f > k` are pairwise isomorphic by Labute (same rank, `q = 2`, image `U^[k]`), all of them his `x₁^{2+2^k}(x₁,x₂)(x₃,x₄)⋯`, the `f = ∞` form that `demushkinWordTwoEven` cannot spell with a natural number. The docstring and README both say this.
- `absoluteGaloisGroupProP_two_of_degree_even`, the old unmarked `∃ a f, … Nonempty (_ ≃ₜ* _)`, is retained as a corollary with a **closed proof** from the two marked theorems and the two image computations.

The `ℚ₂(√−2)` instance is now in Lean: `RatPadicSqrtNegTwo := AdjoinRoot (X² + 2)`, `finrank_ratPadicSqrtNegTwo` (closed, `= 2`), image `procyclicClosure (−negThreeUnit)` with `negThreeUnit_neg_coe : ↑(−negThreeUnit) = −1 + 2²` (closed — the generator `3` is the negative of the supplier's named unit, so no second `IsUnit` proof), `isDyadicEvenPrincipalCase_ratPadicSqrtNegTwo`, and `absoluteGaloisGroupProP_two_ratPadicSqrtNegTwo_marked` on `Fin 4` with relator `x₁⁶(x₁,x₂)x₃⁸(x₃,x₄)` (`a = 4`, `f = 3`) and values `χ(x₂)(1 + 4) = −1`, `χ(x₄)(1 − 8) = 1`, `χ(x₁) = χ(x₃) = 1`; the unmarked corollary is closed. The README's `x⁶(x,y)(z,w)` is the same group with `f = ∞`, as noted there.

**A defect this found in #244 (merged), not fixed here.** `isDemushkin_marked_of_q_two_odd` pins `f` by `hfG : ∃ x, χ(x)(1 − 2^f) = 1`. That condition is monotone in `f`: if `Im χ = {±1} × U^(f₀)` then `(1 − 2^f)⁻¹ ∈ U^(f) ⊆ U^(f₀)` for every `f ≥ f₀`, so `hfG` holds for all `f ≥ f₀`, while the presented group for `f > f₀` has orientation image `{±1} × U^(f)`, a different isomorphism invariant. The even theorem has the same gap in its `{±1} × U^(f)` branch (`haG` pins only `v₂(a) ≥ f`, `hfG` is again monotone); in the `U^[k]` branch it is harmless for the reason in item 3. The pinning hypothesis should be the image equation `(demushkinCharacter hG).range = unitsPlusMinus f`, which is what this roadmap's marked theorems take. This roadmap is unaffected: its odd theorem uses `f = 2` with `range = ⊤`, and its even theorems take the image as hypothesis.

**Verification.** `lake build TauCetiRoadmap.LocalGaloisGroups.Suggested` at `ba3eb7a` (toolchain `v4.34.0-rc2`): green, 8895 jobs, zero errors, only `declaration uses 'sorry'` warnings; `git diff --check` clean; no added line over 100 characters. The branch was not re-merged with `main` this round — `main` has moved by thirteen commits since `f8cd752`, progress reports and one new roadmap, none touching the four suppliers — so the PR's `build` check runs on the merge as usual.

CBirkbeck (changes requested):
:robot: ChatGPT review at Chris's request, against `ba3eb7ad6aaec73853da893421f2e5eebbd240d0`.

The previous finite-Galois-layer, sharp-exponent generation, and even-case marking corrections are present. I also checked that CI run 35626108213 succeeds at this head. Two remaining corrections:

## 1. Separate the tame-layer input from the unrestricted relation-module theorem

Layer 7 Step 3 and `exists_relationModule_surjective` assert the relation-module surjection for **every finite Galois layer** `L/K` and every generating family of size `N+2`. Their stated construction goes through `tateModule_linearEquiv`, hence through the sharp two-element tame-frame module `M₀`. But `exists_tameFrame_exponents` requires a generating pair `σ, τ` with `τ` of order prime to `p`, and `exists_tameFrame_quotient` supplies that pair only on the tame layers used in Step 5. Neither restriction occurs in `exists_relationModule_surjective`.

This is not an automatically available input on the advertised general domain. For example take `p = 3`, `K = ℚ₃(μ₃)`, and a finite Galois layer with `G = (ℤ/3)³`, obtained from the elementary-abelian quotient of `G_K(3)` (its generator rank is four by Layer 3). A generating four-tuple exists, so this satisfies the theorem's stated hypotheses. No pair generates `G`; in particular a prime-to-3 element is the identity. Moreover `q(L) > 1`, so your own induced-module calculation for a proper subgroup `H = ⟨σ,τ⟩` shows that no such pair of root-of-unity lifts can give a sharp quotient of order `q(L)`: its order is at least `q(L)^[G:H]`, or it is infinite.

**This is a gap in the specified proof/dependency route, not a claim that the unrestricted relation-module theorem is false.** Please introduce the tame-frame-restricted surjection and its `H²` compatibility as the actual input to Step 5. Then either put the unrestricted finite-layer theorem after the full-group rank theorem, supplying its separate lifting/comparison argument, or give an independent construction on arbitrary finite Galois layers. Using the unrestricted theorem to prove the rank and later using the rank to justify its existence would be circular. State the ordering explicitly in both the normative roadmap and the target docstring.

## 2. Correct the false converse in the branch discussion

In Layer 6, immediately after `odd_markedCase_exists_unique`, the README says that `q(K) = 2` and `p = 2` are the same hypothesis. The named Lean theorem proves only `q(K) = 2 → p = 2`, correctly. The converse fails for `K = ℚ₂(i)`, where `p = 2` but `q(K) = 4`. Replace the sentence by the one-way implication and the separate dyadic criterion `q(K) = 2 ↔ μ₄ ⊄ K`.

Separately, I have left the remaining **abstract rank-two** marking defect on #431. It does not refute this PR's even-degree local specializations, whose rank is at least four; nevertheless the final supplier contract should cite the corrected marked theorem once that repair lands. No approval has been submitted.

roed-math:

Addressed in [`af5850e`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/251/commits/af5850e950ada20c00608e9d73214e43bb14c732).

**1. The tame-layer input is now separated from the unrestricted theorem — confirmed, and NSW's own order is the one you asked for.** I checked the source (2nd ed., electronic version 2.3, pp. 405–407): the first half of the proof of (7.4.1) is exactly Step 5's argument on a finite *tamely ramified* layer — the frame `σ, τ`, the sequence `(∗)`, `Y ≅ M₀ ⊕ ℤ_p[G]^n`, the diagram giving `β : R^ab_{n+2}(p) ↠ X` with kernel `ℤ_p[G]` and the `H²` isomorphism, the lift of extensions via `scd = 2` and (3.6.4)(iii), the limit over `k^{tr}` and (3.9.1) — while (7.4.2)(i), the sequence for an arbitrary Galois `K|k`, is derived afterwards from the second assertion of (7.4.1), `N^ab(p) ≅ ℤ_p[[G]]`, which itself needs the presentation `F_{n+2} ↠ G_k` that only the rank bound provides. So:

- `exists_relationModule_surjective_of_tameFrame` (Step 3) takes the generating tame frame `σ, τ` with `¬ p ∣ orderOf τ` — what `exists_tameFrame_quotient` supplies — and is the only relation-module input of Step 5; the `H²`-compatibility milestone is stated for this `β`. Its docstring records your `(ℤ/3)³` layer over `ℚ₃(μ₃)` as the reason the route does not extend: no pair generates, every prime-to-`3` element is trivial, and with `q(L) > 1` the induced-module count rules out a sharp quotient.
- `exists_relationModule_surjective`, the unrestricted (7.4.2)(i), is now **Step 7**, placed after `rank_absoluteGaloisGroup` in its own section, with the local-field hypotheses of the rank theorem because its proof consumes it: `π : F_{N+2} ↠ G_K`, `R = π⁻¹(G_L)`, `β` induced by `R ↠ G_L`, kernel the image of `N = ker π`, projective by (5.6.7) with rationalization `ℚ_p[G]` by the dimension count, hence `ℤ_p[G]` by `linearEquiv_of_projective_of_tensorRat`. Both files say that Step 5 may not cite it and why that would be circular; the ordering paragraph and the acceptance checklist say the same.
- One lemma made the "every generating family" quantifier honest in both theorems: `relationModule_linearEquiv_of_closure` (Schanuel on Lyndon's sequence plus Krull–Schmidt), so `β` is constructed for one family — the frame at Step 3, the free generators' images at Step 7 — and transported. The old text's "extending `σ, τ`" is gone for that reason.

**2. The false converse — confirmed.** The sentence now states `q(K) = 2 → p = 2` only, names `ℚ₂(i)` (`p = 2`, `q = 4`) as the counterexample to the converse, and gives the dyadic criterion `q(K) = 2 ↔ μ₄ ⊄ K` as the negation of `localRootOfUnityOrder_ne_two_iff`.

**On #431.** Noted: the local even-degree rank is `N + 2 ≥ 4`, so the fourth marked generator exists here; the Layer 6 README now says so and says the supplier's even theorem records that value for rank at least `4`, with its rank-two family carried by a separate theorem. Nothing in this PR applies the supplier's marked theorems by name, so no signature here depends on that repair.

**Verification.** `lake build TauCetiRoadmap.LocalGaloisGroups.Suggested` at `af5850e` against the merged supplier (toolchain `v4.34.0-rc2`): green, 8895 jobs, zero errors, only `declaration uses 'sorry'` warnings; `git diff --check` clean; no added line over 100 characters. NSW page references are to the electronic 2nd edition, version 2.3 (May 2020), pp. 401–407.

PR #253, opened 2026-08-17
Add roadmap: Zeros of L-functions

roed-math:
> **Dependencies:** This roadmap depends on [Arithmetic Dirichlet Series #192](https://github.com/TauCetiProject/TauCetiRoadmap/pull/192), [L-functions #248](https://github.com/TauCetiProject/TauCetiRoadmap/pull/248), and [Chebotarev #249](https://github.com/TauCetiProject/TauCetiRoadmap/pull/249). It should be reviewed and merged after those supplier PRs. The inherited supplier order includes **[#192](https://github.com/TauCetiProject/TauCetiRoadmap/pull/192); [#188](https://github.com/TauCetiProject/TauCetiRoadmap/pull/188) → [#244](https://github.com/TauCetiProject/TauCetiRoadmap/pull/244) → [#189](https://github.com/TauCetiProject/TauCetiRoadmap/pull/189) → [#191](https://github.com/TauCetiProject/TauCetiRoadmap/pull/191) → [#245](https://github.com/TauCetiProject/TauCetiRoadmap/pull/245) → [#248](https://github.com/TauCetiProject/TauCetiRoadmap/pull/248)+[#249](https://github.com/TauCetiProject/TauCetiRoadmap/pull/249)**.

## Summary

This roadmap develops zero-distribution theory for completed L-functions: gamma growth and logarithm branches, analytic conductors and convexity, finite order and Hadamard factorization, zero counting and Riemann–von Mangoldt formulas, zero-free regions and exceptional zeros, explicit formulas, certified-zero semantics, and effective prime estimates.

The PR contains the normative `README.md`, representative target signatures in `Suggested.lean`, and the root index/import updates.

## Ownership

It consumes generic Abel/Perron/Tauberian infrastructure from #192, Chebotarev's exact Frobenius-prime and counting carriers, generic completed `AnalyticLFunctionData` cards from L-functions, and the accepted Contour Integration roadmap. It owns no current Artin-specific instance and does not define a second `primeTheta`, `primeCount`, or qualitative Chebotarev endpoint.

## Port history

This is a clean port of [roed-math/TauCetiRoadmap#11](https://github.com/roed-math/TauCetiRoadmap/pull/11), rebased onto upstream `main`. Generic summation moved to #192 and qualitative prime counting moved to Chebotarev; the zero-distribution programme, corrected divisor multiplicities, endpoint conventions, and certified-zero semantics remain here. The non-normative `PROVENANCE.md` is omitted; the detailed migration ledger remains private.

## Human review priorities

- gamma/logarithm branches, analytic conductor, and finite-order hypotheses;
- divisor multiplicities and certified-zero semantics;
- zero-free region and exceptional-zero boundaries;
- exact use of the three supplier roadmaps in effective prime estimates.

## Validation

- `lake -Kjobs=1 build TauCetiRoadmap.ZerosOfLFunctions.Suggested`
- `git diff --check`

The roadmap and restructuring were prepared with substantial assistance from Claude Fable and Opus 5, and GPT-5.6 Codex and Pro, under the author's direction. A detailed migration and coordination ledger is maintained privately. No external source code was copied into this roadmap.


CBirkbeck (commented):

**Verdict: no mathematical changes requested. Keep it stacked behind #192, #248, #249 and the contour-integration supplier.**

The roadmap correctly uses divisors rather than sets of zeros, separates the chosen meromorphic representative from its punctured germ, and keeps numerical certification semantics separate from interval arithmetic.

**Required process change.** The exact contour-integration declarations named in Layer 7 must exist before merge; no local replacement should be introduced in the meantime.
---

```text
```

CBirkbeck (commented):

# Deep second review: PR #253

**Roadmap:** Zeros of L-functions

**Head reviewed:** `b9e22eb05f120fb3ab2be6b79a373e2482ca3046`

# PR #253 — Zeros of L-functions

## Verdict

**No new mathematical changes requested.** This is an extremely large analytic roadmap, but the current version correctly makes the zero-free and effective statements conditional on the extra coefficient/Euler-product hypotheses they need rather than deriving them from a bare functional equation.

## What I checked

### 1. Gamma factors and logarithms

The roadmap distinguishes the principal logarithm from a holomorphic logarithm chosen on a zero-free simply connected region. Stirling estimates are stated in sectors away from the negative real axis, with bounded-strip consequences.

### 2. Analytic conductor and convexity

Convexity requires:

- a functional equation;
- controlled gamma factors;
- polynomial growth;
- a finite-order or Phragmén–Lindelöf hypothesis.

It is not inferred from Euler products alone.

### 3. Meromorphic finite order

Before applying Hadamard factorization, poles are removed by an explicit polynomial or divisor. Zero and pole multiplicities are retained.

### 4. Hadamard products

The canonical product genus is tied to the proved order. The roadmap does not use a genus-one product for an arbitrary entire function without an order bound.

### 5. Zero counting

The Riemann–von Mangoldt theorem is stated for a completed datum with exact gamma shifts and conductor. Boundary zeros and poles are either excluded by the contour or counted with an explicit convention.

### 6. Zero-free regions

The roadmap separates generic analytic data from the arithmetic positivity hypotheses used in a de la Vallée Poussin argument. A general completed function with a functional equation need not have a classical zero-free region.

Real primitive characters receive the possible exceptional real zero; complex characters do not.

### 7. Explicit formula

The test-function class, Fourier/Mellin transform convention, trivial zeros, poles, gamma terms and prime-power coefficients are all visible. The formula is not written only for primes.

### 8. Certified zeros

A certificate contains a region, an analytic argument-principle count and isolating data. Numerical output is not accepted as a theorem merely because it comes from an external package.

### 9. Effective prime estimates

Qualitative `primeTheta`/`primeCount` remain owned by Chebotarev. Effective estimates consume those carriers and add explicit zero-free/zero-density inputs.

For nonabelian Chebotarev, any use of Artin `L`-functions or Brauer induction must be an explicit future specialization. The current roadmap does not claim a generic Artin instance.

### 10. Siegel zeros

The exceptional-zero branch retains:

- a real character;
- primitivity;
- a conductor;
- uniqueness in the chosen region;
- the distinction between effective and ineffective bounds.

## Dependency checks

- #192 supplies Perron, Abel and generic Tauberian infrastructure.
- #248 supplies completed Dedekind/Hecke data.
- #249 supplies Frobenius prime-count carriers.
- Contour Integration supplies the argument principle and contour estimates.

No reverse dependency is introduced.

## Lean-facing audit

The most important representative types are:

1. divisors with integer multiplicity;
2. finite-order entire/meromorphic data;
3. zero-free-region records with arithmetic positivity fields;
4. certified regions with boundary nonvanishing;
5. explicit-formula test-function hypotheses;
6. all constants and cutoff conventions.

The roadmap pins these rather than relying only on prose.

## Sources checked

- H. Iwaniec and E. Kowalski, *Analytic Number Theory*.
- E. C. Titchmarsh, *The Theory of the Riemann Zeta-Function*.
- H. Davenport, *Multiplicative Number Theory*.
- Lagarias–Odlyzko, effective Chebotarev.

## Disposition

**Approve after corrected #192, #248, and #249.**

CBirkbeck (changes requested):
## Verdict

**Request changes.**

This is an exceptionally careful roadmap in many places, but the conductor term in the Riemann–von Mangoldt formula is off by a factor of two, and the effective Chebotarev layer invokes Artin L-functions which no supplier constructs.

## 1. Correct the discriminant/conductor factor in the zero-counting main term

For the symmetric count of zeros of a degree-`n` completed Dedekind zeta function, the main term is

\[
N_K^\pm(T)
=
\frac{T}{\pi}
\log\!\left(
 |d_K|^{1/2}
 \left(\frac{T}{2\pi e}\right)^n
\right)
+O(\log(|d_K|T^n)).
\]

The roadmap currently writes `|d_K|`, not `|d_K|^{1/2}`, inside the logarithm. This doubles the discriminant contribution.

For a primitive Hecke character, the same correction is:

\[
\left(|d_K|\,N\mathfrak f\right)^{1/2},
\]

not `|d_K| N𝔣`.

The imaginary-quadratic consistency check should therefore read

\[
\frac{T}{\pi}
 \log\!\left(
 |D|^{1/2}
 \left(\frac{T}{2\pi e}\right)^2
 \right),
\]

which is the sum of the zeta and quadratic-Dirichlet counts.

This is a substantive formula error and should be fixed in Layer 7, the worked examples and any suggested declarations.

## 2. Effective Chebotarev has no Artin-L-function supplier

Layer 8.7 says to expand the conjugacy-class indicator in irreducible characters and apply Artin/Hecke explicit formulae.

The L-functions roadmap explicitly supplies no Artin L-function record, and this roadmap also says it owns no Artin instance. For a nonabelian irreducible character, a Hecke L-function is not directly available.

A valid route needs one of:

- an Artin L-functions roadmap with continuation/holomorphy hypotheses;
- Brauer induction, with a precise virtual combination of Hecke L-functions and control of poles/zeros;
- restriction of the theorem to abelian extensions.

Until that supplier exists, effective nonabelian Chebotarev should move to a successor or be stated conditionally on exact Artin-card hypotheses.

## 3. The general Hecke zero-free region needs a complete proof chain

The source table honestly says that no cited source proves the displayed theorem in the stated generality. That is acceptable only if the missing argument is decomposed into milestones:

- positivity of the relevant logarithmic-derivative combination;
- real/nonreal character split;
- conductor dependence;
- treatment of imprimitive characters;
- reduction of the small-discriminant cases;
- exact exceptional-zero uniqueness and simplicity.

At present too much of this is summarized as “carry the `3-4-1` proof over”.

## 4. Add the domain hypothesis for the normalized logarithm of Gamma

The normalization `logGamma δ 2 = 0` requires `2` to lie in the chosen sector. State the range on `δ`, for example `0 < δ < π`, before fixing that value. This is minor, but the branch normalization should not be ill-typed for an empty or inappropriate sector.

## What is good

The distinction between total representatives and meromorphic germs, the analytic reciprocal of the gamma factor, the pole-cleared continuation, signed divisor versus zero count, boundary regularity for contour integration, and the semantics of zero certificates are all excellent and should remain unchanged.

## Recommendation

Fix the square-root conductor factor first. Then remove or conditionalize effective nonabelian Chebotarev and expand the Hecke zero-free proof.

CBirkbeck (approved):
Current head: `e78a1073248c187414890feb3b276a13ca850975`

## Verdict

**Approve after the L-functions and contour-integration suppliers land.**

The current revision is mathematically much stronger than the first one. The zero-free and explicit-formula arguments are now decomposed into named analytic steps, and the normalization of the zero count is correct.

## A normalization point I checked explicitly

For the one-sided Dedekind-zeta zero count, the roadmap uses
$$
N_K(T)
 =
 \frac{T}{2\pi}
 \log\!\left(
 |d_K|\left(\frac{T}{2\pi e}\right)^{[K:\mathbf Q]}
 \right)
 +O\!\left(\log(|d_K|T^{[K:\mathbf Q]})\right).
$$

This is the standard formula in the convention counting positive ordinates. It is equivalently
$$
\frac{T}{\pi}
\log\!\left(
 |d_K|^{1/2}
 \left(\frac{T}{2\pi e}\right)^{[K:\mathbf Q]/2}
\right)
+O(\cdots).
$$
Thus the present conductor coefficient and gamma coefficient are consistent; the earlier objection obtained by inserting a square root while leaving the rest of the logarithm unchanged would be wrong.

## What I checked

The generic complex analysis is now separated from the arithmetic instances:

- gamma estimates with a fixed logarithm branch;
- order and finite-order predicates;
- Jensen zero counts;
- canonical products indexed by the divisor support;
- an entire logarithm for a nonvanishing entire function;
- Hadamard factorization with explicit convergence modes.

The family-specific steps no longer claim that a general analytic record has a zero-free region. The roadmap identifies exactly where positivity of Euler-product coefficients, the `3-4-1` inequality and the functional equation are used.

The Hecke zero-free region is especially careful:

- the three characters are first presented at one common modulus;
- imprimitive Euler factors are removed with a finite correction;
- real and non-real characters are treated separately;
- the relation between the Hadamard constants for $\chi$ and $\chi^{-1}$ replaces a false “reality” argument for non-real characters;
- the small-conductor range is isolated as a separate finiteness step.

The explicit formula records:

- all residues, including the order at zero;
- the exact trivial-zero multiplicities;
- the positive part of the divisor rather than the signed divisor in the zero sum;
- the test-function class and absolute summability;
- the exceptional-zero term in the prime ideal theorem.

The effective Chebotarev endpoint is restricted to the abelian/ray-class setting and is stated on the carriers supplied by the qualitative Chebotarev roadmap.

I found no remaining mathematical correction to request.

CBirkbeck (dismissed):
Approved as a roadmap, but it should be merged only after the L-functions supplier #248 has been
corrected and merged.

The effective section now has the right scope: unconditional ray-class estimates, and transport to
abelian Chebotarev through an explicit reciprocity dictionary. It no longer claims a nonabelian
effective theorem without an Artin L-function supplier.

The exceptional-zero term is constrained by a theorem identifying it with the sum of the relevant
ray-class terms, so it cannot be an opaque receptacle for the error.

The numerical zero-certificate interface also now expresses completeness: poles are excluded,
multiplicities are matched to the rectangles, overlaps are allowed only along zero-free boundaries,
the total count is recorded, and a coverage theorem says that no zero has been omitted.

I found no additional intrinsic issue in the current version.

CBirkbeck (changes requested):
## Mathematical review: several statements need correction before approval

*AI-assisted review, posted by ChatGPT at Chris Birkbeck's request. Reviewed commit: `ea09a2faa7cf17f407bfbecaeab75ed842acb107`; the PR head was rechecked before posting. This is a review of the normative specification and proposed signatures; I have not run a local Lean build.*

The roadmap has useful distinctions between signed divisor counts and zero counts, half-open counting regions and closed contours, and meromorphic germs and pointwise regularity. However, several statements still violate those distinctions or claim substantially more than their hypotheses provide. The following are required correctness fixes, not objections to intentional `sorry`s.

### 1. `frobeniusPrimeCount_effective` must use `Li(x)`, not `x/log x`

**Location:** `TauCetiRoadmap/ZerosOfLFunctions/Suggested.lean`, `frobeniusPrimeCount_effective`; README Layer 8 and the prime-counting transfer.

The theorem uses `(#C/#G) * (x / log x)` as its main term but claims error

```text
O(x * exp(-c * sqrt(log x)) / log x), c > 0.
```

This is false already for the trivial extension over `ℚ`, where it becomes a statement about `π(x)`. The omitted next term in the logarithmic integral satisfies

```text
Li(x) - x/log x ~ x/log² x,
```

and

```text
(x/log² x) / (x * exp(-c*sqrt(log x))/log x)
  = exp(c*sqrt(log x))/log x → ∞.
```

Thus the difference cannot fit inside the stated error.

**Requested change:** use `(#C/#G) * Li(x)` for the effective prime-counting main term. The corresponding exceptional contribution should be the correctly weighted `Li(x^β)` term(s), as obtained by partial summation. Keep the `x/log x` version only as a weaker asymptotic corollary. Add the trivial-extension/rational-field specialization as an acceptance test.

### 2. `riemannVonMangoldt_generic` admits a constant-function counterexample

**Location:** the generic Riemann–von Mangoldt theorem and `riemannVonMangoldtMainTerm`.

The current hypotheses require continuation, a functional equation, vertical-strip growth and bounds on the gamma shifts, but do not sufficiently connect the declared gamma factors/degree to an actual normalized Dirichlet series.

**Counterexample:** choose an `AnalyticLFunctionData` card with conductor one, `gammaR = {0}`, `gammaC = 0`, root number one, empty polar divisor, and completed function identically one. The continuation, functional-equation and growth hypotheses all hold, and the shift bound can be B=0. The entire completion has no zeros. Nevertheless the asserted degree-one main term is

```text
(T / 2π) * log(T / (2πe)),
```

of order `T log T`, not within a logarithmic error of zero.

**Requested change:** either state the theorem for the two actual families, as the README's scope indicates, or provide and consume sufficient normalized Dirichlet-series, coefficient-growth and nonvanishing/zero-confinement input to prove a genuinely generic version. Continuity, finite order or nonvanishing somewhere do not fix the statement: the constant function already has them. Add this dummy-card rejection test.

### 3. `continuedL_functionalEquation` omits the polar-locus hypotheses

**Location:** the Layer 1 continued-function functional equation, together with `completed_eq_gammaFactor_mul_continuedL` and the raw-product/regularization convention.

Nonvanishing of the two gamma factors does not imply that the completed functions are being evaluated away from their poles. The supplier's `HasFunctionalEquation.eq_away` explicitly excludes those poles, but this consumer's pointwise theorem drops that condition.

**Counterexample to the asserted pointwise generality:** off 2 and -1 let

```text
F(s) = 1 / ((s-2)(s+1)).
```

It satisfies `F(s)=F(1-s)` away from the poles. Assign `F(2)=1` and `F(-1)=0`, use root number one, and give the card simple poles at 2 and -1. Meromorphic continuation, reflected polar orders and the functional equation away from the poles remain valid. Use empty gamma lists, so both gamma factors are identically one and the theorem's gamma hypotheses hold everywhere. There is no removable gamma cancellation at either completed pole; the raw-product convention (also the displayed completion identity) preserves the assigned values. The proposed pointwise functional equation at s=2 would identify 1 and 0.

**Requested change:** add the two polar exclusions required by the supplier for a pointwise result, or state the theorem as an equality of meromorphic germs/punctured-neighborhood representatives. Division by nonzero gamma values does not remove the arbitrary values at completed poles. Include an off-axis or shifted-pole card in the regression tests rather than relying only on the zeta poles at 0 and 1.

### 4. `finite_orderSupport_of_isCompact` needs to exclude identically zero germs

**Location:** the compact finite-order-support helper used in the contour localization step.

The set specified using `meromorphicOrderAt f z ≠ 0` is not finite on every compact subset merely because f is meromorphic. For the identically zero function, `meromorphicOrderAt f z = ⊤` everywhere. Hence this nonzero-order locus contains every point of a compact disk and is infinite.

This must not be confused with the integer-valued divisor, whose total-value convention treats the identically zero germ separately.

**Requested change:** exclude identically zero germs, or work on an appropriate connected domain with a nonzero-function hypothesis and derive the required local nonvanishing through the identity theorem. Alternatively use the actual integer-valued divisor support, with a separate comparison lemma whose hypotheses rule out order `⊤`. Add f=0 as a rejection test. The later regular-boundary application may provide enough nontriviality on the relevant component, but the generic helper still has to state and use it.

### 5. The Weil zero sum is written with `finsum` rather than an analytic infinite sum

**Location:** README Layer 8.4 and the corresponding Hecke version in 8.6.

The global zero sum is displayed with `∑ᶠ`, and absolute summability is offered as the reason this is legitimate and independent of ordering. In Mathlib, `∑ᶠ` is the finite-support sum: if the summand has infinite support, its value is zero. Absolute summability does not change that definition. Compact support of the test function does not imply that its Mellin transform vanishes at all but finitely many zeros.

**Requested change:** use `∑'` with the specified `Summable` theorem, or an explicitly defined limit of finite-height truncations and prove agreement with the absolutely convergent sum. The positive divisor/multiplicity convention is right and should be retained; the error is the summation operation, not the choice to count zeros rather than poles. Audit both the Dedekind and Hecke versions.

### Further precision needed during the correction pass

**Uniformity:** an `=O[atTop]` assertion after fixing a field/card/character permits both its constant and its threshold to depend on that object. For the conductor-uniform results claimed in the README, put the relevant constants before the varying conductor/card/field and explicitly bound the permitted degree and spectral parameters. For example, `heckeZeroFreeRegion` currently fixes K before choosing c, whereas its prose says c depends only on `[K:ℚ]`. Separate the fixed-family theorem from the uniform theorem rather than describing one as the other.

**Nonreal Hecke characters:** the two prime terms in the Weil formula must keep track of the character and its dual. A “verbatim” substitution of the zeta expression must not multiply both `φ(N𝔞)` and its dual test-function term by the same nonreal character. State the Hecke formula explicitly with the appropriate χ and χ⁻¹/conjugate weights and test it on a nonreal finite-order character.

### What to preserve

The removable entire completion, the distinction between natural zero counts and signed contour counts, the half-open height convention, and the certificate's completeness condition are the right design choices. The repairs above should strengthen those interfaces rather than reintroduce raw pointwise quotients or products at cancelled poles. No request here expands the scope to nonabelian Artin L-functions or removes the exceptional-zero terms. Once corrected, recheck the actual supplier statements and the dependency-complete build separately.

roed-math:

Addressed in `a915c0e`. All five required fixes and both precision items are taken as stated; two of them turned out to need more than the review asked for, noted below.

**1. `Li(x)`.** `frobeniusPrimeCount_effective` now has main term `(#C/#G) · logarithmicIntegral x`, with `Li x = ∫_2^x dt/log t` defined here (no supplier declares it; the ADS roadmap's prose names it and is where it should eventually live). `logarithmicIntegral_sub_div_log` pins `Li x − x/log x ~ x/log² x`, the `x/log x` form survives only as `frobeniusPrimeCount_div_log` with error `O(x/log² x)`, and the exceptional prime-count terms are identified rather than opaque: `exceptionalRayClassTerm_eq` says the ψ-level term is `(1/#Cl_𝔪) ∑_ψ ψ(c) x^β/β` over a finite set of real characters with a simple zero `β ∈ (1/2, 1)`, and the prime-count term is the same sum with `Li(x^β)` (not `x^β/(β log x)`, whose defect `≍ x^β/log² x` is not absorbed). `exceptionalChebotarevPrimeCountTerm_eq` is the fibre sum one layer up. Acceptance test: `K = L = ℚ`, where the carrier is `Nat.primeCounting`, the exceptional term is `0`, and the statement is de la Vallée Poussin's `π(x) = Li x + O(x e^{−c√log x}/log x)`.

**2. Riemann–von Mangoldt.** Your constant card is in the file as `constantOneData` with the rejection test. It is excluded by Dirichlet agreement (`LSeries a s → a 1` while `1/Gammaℝ s → 0`), but that alone is not enough, so the generic theorem takes one more hypothesis than you listed. `twistedZetaData` — `ζ(s)(1 + 5·2^{−s} + 2·4^{−s})`, conductor `4`, completed function `Λ_ζ(s)(2^s + 5 + 2^{1−s})` — has continuation, functional equation (the Dirichlet polynomial reflects on its own), growth, Dirichlet agreement with `coeff 1 = 1`, and the average coefficient bound, yet its extra zeros sit on `Re s ≈ 2.19` and `≈ −1.19`, so the strip count is `N_ζ(T)` and the main term is off by `(T/2π) log 4`. `riemannVonMangoldt_generic` therefore takes `HasDirichletAgreement`, `HasAverageCoefficientBound`, and **zero confinement** to the closed strip, and is the fixed-card statement. The uniform statements are the two family theorems `riemannVonMangoldt_dedekind` and `riemannVonMangoldt_hecke`, with `C` and `T₀` bound from the degree before `K` and `χ`.

**3. Polar exclusions.** `continuedL_functionalEquation` now carries `d.polarOrder s = 0` and `d.polarOrder (reflectedPoint s) = 0` alongside the two gamma hypotheses; `continuedL_functionalEquation_eventuallyEq` is the germ form with none of the four. Your card is `shiftedPoleData` (poles at `2` and `−1`, values `1` and `0` assigned), with the rejection test that both gamma hypotheses hold at `s = 2` and the old conclusion there is `0 = 1`. `completed_eq_gammaFactor_mul_continuedL` is unchanged, as you noted: the raw-product convention preserves the assigned values there.

**4. Order `⊤`.** `finite_orderSupport_of_isCompact` takes `∀ z ∈ K, meromorphicOrderAt f z ≠ ⊤`, with `f = 0` as `finite_orderSupport_zero_fails`. The localization discharges it the way you suggested: step 1 now produces a **connected** `V`, and `meromorphicOrderAt_ne_top_of_frontier` gets finite order on `closure V` from one boundary point of order `0` via Mathlib's `meromorphicOrderAt_ne_top_of_isPreconnected`. The README distinguishes the nonzero-order locus from the divisor support (which is `0` at order `⊤` and locally finite unconditionally).

**5. `∑'`.** Both Weil formulas (8.4, 8.6) and the `K = ℚ` worked example now use `∑'` with the `Summable` theorem as part of the milestone; a new conventions-table row makes `∑ᶠ` count-only. The positive-divisor carrier is kept.

**Uniformity.** `heckeZeroFreeRegion` is now `∀ n, ∃ c > 0, ∀ K [finrank = n], ∀ χ, …` over the named predicate `HeckeZeroFreeDisjunction`; the fixed-`K` form is the separate corollary `heckeZeroFreeRegion_of_fixed`. Layer 6.4 and 7.6 in the README say that uniformity is a quantifier order.

**Nonreal Hecke characters.** 8.6 now writes the prime side as `−∑_𝔞 Λ_K(𝔞)(χ(𝔞) φ(𝔑𝔞) + conj(χ(𝔞)) φ̌(𝔑𝔞))` with the derivation (the `φ̌` term comes from the reflected contour through `Λ(χ, s) = ε Λ(χ⁻¹, 1 − s)`). Worked example: the quartic character mod 5 with `φ` supported in `(0.4, 0.6)`, `φ(1/2) = 1`, where the `φ` term is `0` and the `φ̌` term is `−(i log 2)/2`; the both-`χ` version gives `+(i log 2)/2`.

**Verification.** `lake build TauCetiRoadmap.ZerosOfLFunctions.Suggested` on the rc1 toolchain with all suppliers present locally: 8772 jobs, exit 0, only `sorry` warnings. The GitHub `build` check stays red on the unmerged #248 import, as before.

CBirkbeck (changes requested):
:robot: ChatGPT follow-up review at Chris's request, against `a915c0e049e50d2850b95e21c9c83280c580bf6b`.

The latest corrections are present: `Li` is now the prime-count main term, the generic Riemann–von Mangoldt statement carries Dirichlet agreement and confinement, the pointwise functional equation carries the polar exclusions, the localization excludes order `⊤`, and the Weil zero sums use `tsum`. The divisor-based certificate conditions and the connected-localization route now fit together. I found the following remaining corrections.

## 1. Layer 6.4d drops a necessary discriminant term

The normative roadmap asserts, for `1 < σ ≤ 2`,

```text
−ζ'_K/ζ_K(σ) ≤ 1/(σ−1) + O(1),
```

with the implied constant depending only on `[K:ℚ]`. A simple pole gives a field-dependent remainder, not this uniform statement. It is false already in degree two.

Here is an unconditional counterargument. Fix `σ > 1` and choose `B` large enough that

```text
Σ_{p≤B} log p/(p^σ−1) ≥ −ζ'/ζ(σ) − 1.
```

Choose a prime `D ≡ 1 mod (8 ∏_{p≤B} p)` and put `K = ℚ(√D)`. Every prime `p≤B` splits in `K`, so positivity of the remaining Euler terms gives

```text
−ζ'_K/ζ_K(σ) ≥ 2 Σ_{p≤B} log p/(p^σ−1)
             ≥ 2/(σ−1) − O(1).
```

Letting `σ → 1⁺`, and choosing `K` as above at each stage, contradicts any degree-two constant in the displayed bound.

The correct input can retain the discriminant term:

```text
−ζ'_K/ζ_K(σ) ≤ 1/(σ−1) + (1/2) log |d_K| + O_n(1).
```

For an exact primary-source comparison, Kadiri, *Explicit zero-free regions for Dedekind zeta functions*, arXiv:1106.1868, printed p. 5, §2.1.2, displays the logarithmic-derivative identity immediately before (2.7), including `(1/2) log d_K`, the two pole terms, the gamma derivative, and the negative zero sum. Please retain this term in 6.4d and propagate it through the `3-4-1` and uniqueness estimates. This does not refute the intended zero-free-region endpoint; it refutes a required intermediate estimate in its stated proof route.

## 2. The exceptional-term contract still does not select the exceptional zeros

`exceptionalRayClassTerm_eq` only supplies an existential finite set `S` of **some** nontrivial real characters and simple real zeros in `(1/2,1)`. It has neither membership in the chosen zero-free region nor an iff characterizing `S` as all characters whose primitive source has an exceptional zero. `S = ∅` is permitted for every modulus; moreover `S` and `β` are selected after fixing the ray class `c`, so this contract does not give the common character-indexed family needed by the fibre/orthogonality argument.

There is also a missing branch: Layer 8.5 explicitly permits an exceptional zero of `ζ_K`. The trivial ray-class character contributes that same primitive L-function, and its term `x^β/(β·#Cl_𝔪)` must be present in the exceptional contribution in that branch. The condition `ψ ≠ 1` in `exceptionalRayClassTerm_eq` deletes it. This objection is conditional on the branch your roadmap already retains; it does not assert the existence of a Siegel zero or require a Stark identification with another character over `K`.

Please fix the zero-free constant/predicate, define the finite character-indexed exceptional data independently of `c`, characterize membership in both directions using primitive sources, include the possible trivial-character contribution, and derive both the `x^β/β` and `Li(x^β)` terms from those same data. Then prove the fibre-identification and modulus-independence claims from character orthogonality.

The current **fixed-field/fixed-modulus** `=O[atTop]` estimates cannot force this identification. For every fixed `β < 1` and `c₀ > 0`,

```text
x^β / (x exp(−c₀√log x))
  = exp(−(1−β) log x + c₀√log x) → 0.
```

Thus absorbing an exceptional term into those particular asymptotic bounds does not require Siegel's theorem; the threshold can depend on `β`. The repeated claim to the contrary, including the corresponding claim about the `Li(x^β)` approximation, confuses these fixed-object estimates with a uniform/effectively controlled threshold. The explicit exceptional terms are useful, but their semantics must come from the definition/iff theorem, not from these `BigO` statements.

## 3. Correct the nonreal-character reflection in 6.4d

The functional equation sends a zero `ρ` of `Λ(χ,·)` to a zero `1−ρ` of `Λ(χ⁻¹,·)`. Conjugation sends it to `conj ρ` for `χ⁻¹`. Combining them sends `ρ` to `1−conj ρ` for the **same** `χ`, not for `χ⁻¹` as the paragraph currently says. Please write the two maps separately when proving the relation between the Hadamard constants; the nonreal case is exactly where the distinction matters.

Current-head CI run **35625447662** fails, and #248 remains an open supplier with separate mathematical corrections requested. I have not submitted approval.

roed-math:

Addressed in `3da8da2`. All three are taken as stated.

**1. 6.4d keeps `½ log |d_K|`.** The `ζ_K` bound is now `−ζ'_K/ζ_K(σ) ≤ 1/(σ−1) + ½ log |d_K| + O_n(1)`, cited to Kadiri §2.1.2 (the display before (2.7)), with your split-primes family `ℚ(√D)`, `D ≡ 1 (mod 8∏_{p≤B} p)`, recorded as the refutation of the degree-only form. Propagated: the `3-4-1` assembly now reads `4/(σ−β) ≤ 3/(σ−1) + 4 log q + O_n(1)` (using `log |d_K| ≤ log q` and `log q(2it) ≤ 2 log q(it) + O_n(1)`), and 4f's count is against `1/(σ−1) + ½ log |d_K| + ½ log q ≤ 1/(σ−1) + log q`. The endpoint is unchanged; the source-table row for 6.4 records the term.

**2. Exceptional terms are now definitions, not an existential contract.**
- `heckeZeroFreeRegion` now covers the trivial primitive character (its completed function is `completedDedekindZeta`, per the supplier), so 6.2 is its instance, and `heckeZeroFreeConstant n := Classical.choose (heckeZeroFreeRegion n)` fixes one constant per degree.
- `IsExceptionalZero χ β` is defined against that constant for every primitive `χ`: `1 − c/log q(χ,0) ≤ β < 1` and `0 < divisor Λ(χ,·) β`. Uniqueness, simplicity and reality are theorems from `heckeZeroFreeConstant_spec`; membership is characterized both ways through the primitive source by `isExceptionalZero_of_iff` (on the presented series, via 6.4b), and `isExceptionalZero_of_one_iff` reads the trivial character against `ζ_K`.
- `exceptionalRayClassTerm 𝔪 c x := (#Cl_𝔪)⁻¹ ∑_ψ ∑_{β ∈ {β | IsExceptionalZero (of ψ) β}} Re ψ(c) · x^β/β` over **all** characters at `𝔪`, trivial included; `exceptionalRayClassPrimeCountTerm` is the same family with `Li(x^β)`. Both finsums are honest (`finite_rayClassCharacter`, `setOf_isExceptionalZero_subsingleton`), and the family is independent of `c` except through the weight, which is what `exceptionalChebotarevTerm_eq` needs. The old `exceptionalRayClassTerm_eq` is gone; `exceptionalRayClassTerm_eq_zero` is what the ℚ/ℚ test now uses (the only character is trivial and `ζ` has no real zero in `(0,1)`).
- The Siegel claim was wrong as you say. Every occurrence (out-of-scope list, 8.5, 8.6, 8.7, Lean docstrings) now states that for fixed objects `x^β = o(x e^{−c√log x})`, so the fixed-field `=O[atTop]` holds with or without the term, that the term is displayed for its semantics (6.3 evaluated) and for threshold-explicit or conductor-uniform forms, and that Siegel enters only for conductor-uniform absorption. The `Li(x^β)` vs `x^β/(β log x)` remark is corrected the same way: the shape is fixed by the definition, and is only distinguishable in those forms.

**3. The reflection in 6.4d** is written as two maps: the functional equation sends `ρ ↦ 1 − ρ` into `Λ(χ⁻¹,·)` and gives `Λ'/Λ(χ,s) = −Λ'/Λ(χ⁻¹,1−s)`; conjugation `Λ(χ⁻¹,s) = conj Λ(χ, conj s)` sends `ρ ↦ conj ρ` into `Λ(χ⁻¹,·)` and gives `b(χ⁻¹) = conj b(χ)`; the composite `ρ ↦ 1 − conj ρ` is a symmetry of `Λ(χ,·)` itself. The derivation of `Re b(χ) = −∑ Re(1/ρ)` is written out from `b(χ) + b(χ⁻¹) = −∑_ρ (1/ρ + 1/(1−ρ))`.

**Verification.** `lake build TauCetiRoadmap.ZerosOfLFunctions.Suggested` on the rc1 toolchain with suppliers present locally: 8772 jobs, exit 0, only `sorry` warnings. CI stays red on the unmerged #248 import.

PR #255, opened 2026-08-17
Add roadmap: Orthogonal and spin groups

roed-math:
> **Dependencies.** The direct dependencies are exactly the three roadmaps `Suggested.lean` imports: [Local Fields and Ramification #189](https://github.com/TauCetiProject/TauCetiRoadmap/pull/189), [Adelic Algebraic Groups #246](https://github.com/TauCetiProject/TauCetiRoadmap/pull/246), and [Quadratic Form Invariants #252](https://github.com/TauCetiProject/TauCetiRoadmap/pull/252), plus merged Tau Ceti and Mathlib infrastructure. [#250](https://github.com/TauCetiProject/TauCetiRoadmap/pull/250) and [#254](https://github.com/TauCetiProject/TauCetiRoadmap/pull/254) are **no longer direct dependencies**: nothing in the current scope imports or consumes them, and they are inputs only to the named successor below. The inherited supplier order is **[#188](https://github.com/TauCetiProject/TauCetiRoadmap/pull/188) → [#244](https://github.com/TauCetiProject/TauCetiRoadmap/pull/244) → [#189](https://github.com/TauCetiProject/TauCetiRoadmap/pull/189) → [#191](https://github.com/TauCetiProject/TauCetiRoadmap/pull/191) → [#250](https://github.com/TauCetiProject/TauCetiRoadmap/pull/250) → [#252](https://github.com/TauCetiProject/TauCetiRoadmap/pull/252)**, with [#246](https://github.com/TauCetiProject/TauCetiRoadmap/pull/246) independent of that chain.

## Summary

This roadmap develops the arithmetic of the orthogonal and spin groups of a finite-dimensional
nondegenerate quadratic space over a field of characteristic not two, and their specialization
to the local and finite-adelic setting. `Suggested.lean` carries 109 declarations that
elaborate against the pinned Mathlib and the three suppliers above:

- **the algebraic groups** — `orthogonalGroup`, `specialOrthogonalGroup`, the determinant
  character and the index-two subgroup, and `reflection` with its determinant, conjugation and
  membership lemmas;
- **Clifford conventions** — `cliffordNorm` and `starNorm` kept apart, `scalarUnits`, the
  vector representation with its surjectivity and its kernel, and the dimension-zero rejection
  test that shows why the kernel theorem needs positive dimension;
- **the spinor norm** — `spinorNorm` and its value on a reflection, `spinToSpecialOrthogonal`
  with `muTwoToSpin`, and the kernel and range of that map;
[...]

PR #256, opened 2026-08-17
Expand roadmap: Integral quadratic forms and lattices

roed-math:
> **Dependencies.** This roadmap extends the one merged in [Integral Lattices #200](https://github.com/TauCetiProject/TauCetiRoadmap/pull/200). Its direct dependencies are exactly the six roadmaps `Suggested.lean` imports: [Quadratic Form Invariants #252](https://github.com/TauCetiProject/TauCetiRoadmap/pull/252), [Global Quadratic Forms #254](https://github.com/TauCetiProject/TauCetiRoadmap/pull/254), [Global Number Fields #245](https://github.com/TauCetiProject/TauCetiRoadmap/pull/245), [Class Field Theory #250](https://github.com/TauCetiProject/TauCetiRoadmap/pull/250), [Adelic Algebraic Groups #246](https://github.com/TauCetiProject/TauCetiRoadmap/pull/246), and [Orthogonal and Spin Groups #255](https://github.com/TauCetiProject/TauCetiRoadmap/pull/255), plus merged Root Systems and Mathlib. The inherited supplier order is **[#192](https://github.com/TauCetiProject/TauCetiRoadmap/pull/192); [#188](https://github.com/TauCetiProject/TauCetiRoadmap/pull/188) → [#244](https://github.com/TauCetiProject/TauCetiRoadmap/pull/244) → [#189](https://github.com/TauCetiProject/TauCetiRoadmap/pull/189) → [#191](https://github.com/TauCetiProject/TauCetiRoadmap/pull/191) → [#245](https://github.com/TauCetiProject/TauCetiRoadmap/pull/245) → [#246](https://github.com/TauCetiProject/TauCetiRoadmap/pull/246)+[#250](https://github.com/TauCetiProject/TauCetiRoadmap/pull/250) → [#252](https://github.com/TauCetiProject/TauCetiRoadmap/pull/252) → [#254](https://github.com/TauCetiProject/TauCetiRoadmap/pull/254) → [#255](https://github.com/TauCetiProject/TauCetiRoadmap/pull/255) → this**.

## Summary

This expands the `IntegralLattices` roadmap merged in #200 to the arithmetic of integral
lattices over `ℤ` and `ℤ_p`: Gram and discriminant data, dual lattices and discriminant forms,
localizations and Jordan theory, Conway–Sloane genus symbols, classes and spinor genera, binary
lattices and quadratic-order ideal classes, Nikulin's existence/uniqueness/embedding theory,
unimodular lattices in low rank, the local ingredients of a mass, theta series, and the
assertion semantics of the LMFDB lattice records.

It extends #200 rather than adding a second lattice roadmap. The embedded `IntegralLattice`
stays the sole public carrier, and #200's reviewed radical/signature, dual-lattice,
discriminant-module, gluing, ADE, and `D₈⁺ ≅ E₈` routes all survive into the expanded
programme.

## What this revision changes: the mass formula moves out

The reviewed scopes of #246 and #255 deliberately export **no** generic strong approximation,
reduction theory, Tamagawa measure, or central-isogeny volume comparison. This roadmap was
still resting on all four, which is a prose reference to a removed milestone rather than a
closed dependency. The affected results now have one exact owner — the successor roadmap
**`OrthogonalTamagawaAndLatticeMass`**, which #255 and #246 name from their side:

| Moved out | Was | Why it cannot stay |
| --- | --- | --- |
| Eichler's theorem `cls⁺ L = spn⁺ L`, and the class-number finiteness in rank `≥ 3` that follows | 4D, and the rank-`≥ 3` half of 4E | needs strong approximation for `Spin`, which no accepted roadmap proves |
| the adelic decomposition of `SO(V)(ℚ) \ SO(V)(𝔸)` into class-indexed pieces | 7B | needs a Tamagawa measure |
| the identity `vol(K_p(L)) = α_p(L)` in the Tamagawa normalization | one bullet of 7C | the gauge-form measure is undefined without that normalization |
| the volume theorem `vol(SO(V)(ℚ) \ SO(V)(𝔸)) = 2` | 7F | #255 states that only the successor may export it |
| the Smith–Minkowski–Siegel formula and its checks | 7H | assembled from 7B and 7F |
| rank-16 completeness: the genus of `E₈²` has exactly two classes | 7I | needs 7H |

**Everything the successor will need from the lattice side stays here**, so the split leaves no
gap between the two documents: the class/genus/proper-genus sets and the stabilizer dictionary
(4A, 4B), spinor genera and the proper-spinor-genus class group (4C), the automorphism-group
infiniteness argument (4F), Kneser neighbours (4G), the two masses and the relation `m⁺ = 2m`
(7A), the local densities at odd `p` (7C) and Cho's dyadic density at `2` (7D), the archimedean
factor (7E), and the low-rank values in ranks 0, 1 and 2 (7G). Those are lattice arithmetic and
use no adelic volume. Rank 2 of 4E likewise stays, since it runs through the binary
correspondence B2/B5 and the Picard group, not through strong approximation.

`Suggested.lean` follows: it no longer `#check`s
`OrthogonalSpinGroups.strongApproximation_finiteAdelicSpin`, a name #255 does not export and
has said only the successor may. In its place it checks four declarations #255 does export and
this roadmap does consume — `OrthogonalCompactOpens`, `finiteAdelicOrthogonal`, `transvection`,
`transvectionLiftHom`. Every remaining `#check` names a declaration its supplier actually
exports.

## A second commit: the dependency-complete build did not actually pass

An earlier revision of this description said a dependency-complete build of this target passes.
It did not. With every supplier branch merged locally, `Suggested.lean` reported 25 errors, which
GitHub's `build` check cannot see because it fails earlier on the seven unmerged imports. They are
repaired here, and none of them changes a statement:

- `formTwist` is a `sorry`-bodied `def` whose type does not mention `L`, so Lean dropped the
  section variable and every `L.formTwist a` failed. `L` is now bound explicitly.
- `discriminantBilinearModule` and `discriminantQuadraticModule` were `sorry`-bodied records,
  which makes their carrier `A` opaque — and then `AddSubgroup L.DiscriminantGroup` does not match
  `AddSubgroup (…).A`, so the isotropic, Lagrangian and overlattice statements of Layer 1, and the
  `D₈⁺ ≅ E₈` artifact, could not be written at all. The carrier and the pairing are now **data**:
  the bilinear module is built on `L.DiscriminantGroup` with `discriminantPairing`, the quadratic
  module on it with `discriminantQuadraticForm`. Only `symmetric` and `polar` stay milestones.
- `LocalCarrier` and `CompletedAmbient` use `ℤ_[p]` and `ℚ_[p]` and were missing `[Fact p.Prime]`,
  so the whole Layer 3 localization block failed to elaborate.
- the Layer 3G Hilbert-reciprocity check passed no field to `ClassFieldTheory`'s
  `finiteHilbertSupport`, `finiteHilbertInvariantAt` and `infiniteHilbertInvariantAt`, which take
  it explicitly, nor to `QuadraticFormInvariants.hilbertSymbol_productFormula`, which now binds
  its own field at universe `0` because Class Field Theory pins its cohomology there.

## Ownership

One canonical owner per prerequisite: local forms from #252; global rational classification
from #254; number-field orders, the ideal class monoid, `Pic` and `NarrowPic` from #245;
nonsplit ring class fields from #250; generic restricted products and rational diagonals from
#246; the orthogonal groups, the spinor norm and the finite-adelic point groups from #255;
Poisson summation and the real-parameter Gaussian transform from #248; ADE from merged Root
Systems. The `LFunctions` dependency is one-way: it supplies the analytic Gaussian identity,
this roadmap owns the arithmetic theta series and its transformation, and no L-functions
milestone imports a lattice theorem. Modularity and half-integral weight stay outside.

## Corrections preserved from earlier rounds

Unchanged by this revision, and listed so a re-review can confirm no regression: raw proper
fractional ideals versus invertible proper fractional ideals, with `Pic`/`NarrowPic` only on
the invertible carrier and a separate ideal class monoid for the noninvertible ones; the
square-discriminant split branch treated separately from the nonsplit one; `formTwist` versus
carrier dilation; full-submodule versus rational-span restriction; the half-norm
discriminant-form convention in `ℚ/ℤ` with Nikulin's `ℚ/2ℤ` reached only through an explicit
dictionary; the nondegeneracy hypothesis in the Nikulin decomposition; and the rank-one and
`D₈⁺ ≅ E₈` acceptance artifacts.

## Human review priorities

- the successor split above — whether the line between 7A/7C/7D/7E/7G and 7B/7F/7H/7I is in the
  right place, and whether `OrthogonalTamagawaAndLatticeMass` is the right single owner for the
  far side;
- that nothing in Layers 0–6, 8 and 9 still leans on a removed supplier;
- preservation of #200's embedded carrier and half-norm discriminant-form convention;
- the dyadic Jordan/genus-symbol boundary and Cho's normalization in 7D;
- theta-series ownership and the one-way `LFunctions` dependency.

## Validation

- dependency-complete integration checkout (all seven suppliers merged in locally):
  `lake build TauCetiRoadmap.IntegralLattices.Suggested` and a whole-repository `lake build`
- `python3 .github/scripts/check_roadmap_areas.py` — passes (this roadmap already has its README
  line and dropdown entries from #200)
- `git diff --check` — passes

The PR branch keeps the direct imports of its seven open suppliers, so the GitHub Actions
`build` check stays dependency-blocked until those PRs merge; it goes green as soon as they do
and the branch is updated from `main`. **This PR is not a draft** — an earlier revision of this
description said it was, which was wrong.

The roadmap and this revision were prepared with substantial assistance from Claude Fable and
Opus 5, and GPT-5.6 Codex and Pro, under the author's direction. A detailed migration and
coordination ledger is maintained privately. No external source code was copied into this
roadmap.


CBirkbeck (changes requested):

**Verdict: request changes inherited from #245, and keep the late layers stacked behind their suppliers.**

### Required mathematical correction

The binary quadratic-form / order correspondence must use the Picard group of **invertible proper ideals**, not all proper ideals — for a general number-field order, proper does not imply invertible (see #245). If noninvertible ideals are to be classified, they form a separate ideal class *monoid* and do not parameterise the Picard group.

Audit every binary-form milestone consuming

```text
NumberFieldOrder.properIdeals
Pic O
NarrowPic O
```

once #245 is repaired.

### Dependency changes

The early lattice, discriminant-form and Jordan layers may be reviewed independently. The later parts are not ready to land before #245 (orders and Picard groups), #246 (adelic measure carriers), #252 and #254 (local/global quadratic forms), and #255 (spinor genera and strong approximation). The mass formula and spinor-genus applications should import those final declarations rather than recording prose contracts.

### Structural recommendation

The internal boundaries are good enough that this may remain one broad roadmap, but the implementation should be staged in at least three PRs: (1) integral and `p`-adic lattices, discriminant forms and genera; (2) binary forms/orders and Nikulin embeddings; (3) spinor genera, mass formula and theta series.
---

```text
```

roed-math:
**Authorship disclosure:** This update was prepared and posted by **GPT-5.6 Codex** on behalf of David Roe.

Addressed the required binary-form/Picard correction in [`e9a006a`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/e9a006afec04e27275481c7092b48d82ead195aa), against the corrected Global Number Fields contract from #245 (`1035a24`):

- replaced the obsolete `NumberFieldOrder.properIdeals` group contract with the exact split between raw `properFractionalIdeals` and group-valued `invertibleProperFractionalIdeals`;
- made B2 first prove that the binary-form ideal is proper and then use `isProper_iff_isUnit_of_finrank_eq_two` at the quadratic-field boundary before passing it to `mkPic` or `NarrowPic`;
- recorded that noninvertible proper ideals belong to `IdealClassMonoid`, with `picEquivUnitsIdealClassMonoid` identifying only its units with `Pic`;
- audited the B1--B5 supplier table, B2--B5 narrative, the rank-two class-number application, and the representative Lean contracts so that every group-valued use is explicitly invertible;
- added direct checks for the corrected #245 declarations, including the raw and invertible carriers, the quadratic specialization, the ideal class monoid, and the narrow-to-wide map.

The dependency feedback remains enforced: the later spinor-genus and mass layers directly import their supplier roadmaps, introduce no local substitute carriers, and PR #256 remains sequenced after #245, #246, #252, #254, and #255. README-only supplier milestones are still identified as such and must be replaced by accepted declarations before this consumer lands.

Verification:

- `python3 .github/scripts/check_roadmap_areas.py` — passes on the PR branch;
- `git diff --check` — passes;
- a disposable integration stack was assembled through #255 with corrected #245 layered in.

The Lean target build was **not completed in this run**: the disposable checkout started without a dependency cache, and the bounded `lake exe cache get` attempt was stopped while fetching caches. This is a validation limitation, not a claimed pass; the branch remains intentionally supplier-dependent until the upstream PRs land.

The recommendation to split implementation into three future PRs is staging guidance rather than a correction to this roadmap PR; the roadmap already exposes the corresponding early/local, binary/Nikulin, and spinor/mass/theta boundaries.

roed-math:

Addressed the architectural blocker in [`cda777f`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/cda777f), repaired the targets a dependency-complete build rejects in [`428d1b3`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/428d1b3), and rewrote the PR description. The mathematical corrections from the earlier rounds are all intact.

**The blocker was:** *"the README still relies on strong-approximation and Tamagawa results that the final scopes of #246 and #255 explicitly no longer own. A prose reference to removed supplier milestones is not a closed dependency."*

**Response: narrow, and give the removed summit one exact owner.** That owner is **`OrthogonalTamagawaAndLatticeMass`**, the orthogonal specialization #255 and #246 now name from their side, sitting over #246's generic `AlgebraicGroupStrongApproximation`, `ArithmeticReductionTheory` and `TamagawaMeasures`. Moved out:

| Moved out | Was | Why it could not stay |
| --- | --- | --- |
| Eichler's theorem `cls⁺ L = spn⁺ L`, and the rank-`≥ 3` class-number finiteness that follows | 4D, half of 4E | needs strong approximation for `Spin` |
| the adelic decomposition of `SO(V)(ℚ) \ SO(V)(𝔸)` | 7B | needs a Tamagawa measure |
| `vol(K_p(L)) = α_p(L)` in the Tamagawa normalization | one bullet of 7C | the gauge-form measure is undefined without it |
| the volume theorem `vol(SO(V)(ℚ) \ SO(V)(𝔸)) = 2` | 7F | #255 says only the successor may export it |
| the Conway–Sloane mass formula and its checks | 7H | assembled from 7B and 7F |
| rank-16 completeness | 7I | needs 7H |

**What stayed, and why the split leaves no gap.** Everything the successor consumes from the lattice side is still a milestone here: 4A class/genus sets, 4B the stabilizer dictionary and the two double-coset correspondences, 4C spinor genera with the proper-spinor-genus class group, 4F, 4G, rank 2 of 4E through B2/B5, 7A the two masses with `m⁺ = 2m`, 7C's local density at odd `p`, 7D Cho's dyadic density, 7E the archimedean factor, and 7G the low-rank values in ranks 0, 1 and 2. None of those uses an adelic volume. Layer 7 is retitled "masses, local densities, and the archimedean factor" accordingly, and each moved milestone carries an *Owner:* line so a reader who opens the file in the middle cannot mistake a specification for a claim.

**A concrete supplier defect this exposed.** `Suggested.lean` contained

```lean
#check OrthogonalSpinGroups.strongApproximation_finiteAdelicSpin
```

and #255 does not export that declaration — its README says explicitly that only the successor may. So the file could not have elaborated even in a fully merged stack. It is removed, and in its place are four names #255 does export and this roadmap does consume: `OrthogonalCompactOpens`, `finiteAdelicOrthogonal`, `transvection`, `transvectionLiftHom`. Every remaining `#check` names a declaration its supplier actually exports.

**Scope and dependency tables rewritten.** The out-of-scope table drops the claim that #255 owns "the orthogonal Tamagawa theorem" and that #246 owns "strong approximation and Tamagawa measures", and gains rows for the generic and orthogonal successors. The supplier table's #246 and #255 rows now list only exported declarations. The ordering section, the API checklists, the hard-theorem table, and the worked examples follow.

**Stale metadata fixed.** The description said the PR was a draft; it is not, and the new description says so. The dependency-chain link labelled #255 pointed at `/pull/254`; corrected. The summary no longer claims the mass-formula summit.

**Corrections preserved from earlier rounds**, listed so a re-review can confirm no regression: raw proper fractional ideals versus invertible proper fractional ideals, with `Pic`/`NarrowPic` only on the invertible carrier and a separate ideal class monoid for the noninvertible ones; the square-discriminant split branch; `formTwist` versus carrier dilation; full-submodule versus rational-span restriction; the half-norm discriminant-form convention in `ℚ/ℤ`, with Nikulin's `ℚ/2ℤ` reached only through the explicit dictionary; the nondegeneracy hypothesis in the Nikulin decomposition; and the rank-one and `D₈⁺ ≅ E₈` acceptance artifacts.

**Verification.** In a dependency-complete local stack (every open roadmap branch merged in) `lake build TauCetiRoadmap.IntegralLattices.Suggested` and a whole-repository `lake build` both pass. `check_roadmap_areas.py` and `git diff --check` pass.

**Merge order.** This is the last node of the arithmetic portfolio: its `build` check is red on its seven unmerged supplier imports and goes green once #252, #254, #245, #250, #246, #255 and #248 land and the branch is updated from `main`.

CBirkbeck (changes requested):
## Verdict

**Request changes.**

This is one of the strongest roadmaps in the collection. It has corrected several subtle errors in binary masses, dyadic symbols and Nikulin's hypotheses. The remaining problems are mostly dependency closure: the dyadic local-density layer and the certificate layer claim objects for which no current algebraic-group or mass supplier exists.

## 1. The dyadic local-density milestone is a roadmap-sized algebraic-group construction

Layer 7D asks for Cho's smooth affine group scheme `𝒢_L`, its special fibre, unipotent radical and reductive quotient, for a general unramified extension of `Z₂`.

No current roadmap supplies:

- smoothening of the orthogonal group scheme;
- special fibres and reductive quotients of affine group schemes;
- dimensions and point counts of the unipotent radical;
- the comparison of the smooth-model density with the congruence-count limit.

This is not a small lattice calculation. Either:

- expand 7D into an exact construction chain with an algebraic-group-scheme supplier; or
- move it to `OrthogonalTamagawaAndLatticeMass`, which already owns the measure comparison.

It is inconsistent to say that all algebraic-group measure infrastructure is out of scope while constructing the most delicate smooth integral model locally inside this roadmap.

## 2. The Nikulin existence predicate must be a typed object, not only a table in prose

Nikulin's boundary-rank conditions at `p=2` are the highest-risk part of the roadmap. The exact predicate should be represented in `Suggested.lean`, including:

- `l(A_{q_p})`;
- determinant square-class conditions at equality;
- the special dyadic alternatives;
- the signature/Brown-invariant congruence.

Then the theorem should literally say that the predicate is necessary and sufficient.

A generic field named `localConditions` would not be enough. The point of the roadmap's careful source audit is to prevent one omitted boundary clause.

The same applies to the `2`-elementary admissibility predicate.

## 3. Separate certificate semantics from completeness proofs

The LMFDB-facing `StoredGenusCertificate` may safely assert:

- Gram matrix;
- local symbols;
- minima and shell counts;
- automorphism order;
- membership in a genus;
- pairwise nonisometry.

It cannot internally certify a complete class list unless it imports either:

- the mass formula;
- a proved connected neighbor graph in the exact regime;
- or another independent completeness theorem.

Those were deliberately moved to `OrthogonalTamagawaAndLatticeMass`. Make completeness a separate downstream certificate whose assumptions include an imported exact mass or connectivity theorem. Do not leave a `class_number` field suggesting that the current roadmap proves exhaustiveness.

## 4. The finite class-number statements need their proof owner visible

For positive-definite lattices, finiteness follows from reduction/bounded Gram matrices and can remain here. For indefinite rank at least three, the roadmap correctly moves Eichler/strong approximation out.

Ensure that no generic `finite_genusClassSet` theorem in this roadmap silently covers the indefinite high-rank case. The statement should branch:

- positive definite: proved here by reduction;
- indefinite rank two: proved here by ideal classes;
- indefinite rank at least three: successor.

## 5. The binary ring-class comparison should be restricted to the nonsplit quadratic branch

The roadmap mostly does this correctly. Make it impossible in the API to ask for a ring class field of the split quadratic algebra `Q×Q`. The split branch has an elementary ideal/class description and no quadratic field ring class field.

## 6. The theta-series endpoint should be described as a transformation law, not modularity

The roadmap correctly moves modularity out. Ensure the exported theorem says precisely:

- convergence and holomorphy;
- Poisson transformation under `S`;
- parity/translation under `T`;
- vector-valued transformation for discriminant cosets where developed.

Do not call this `isModularForm` until a modular-forms carrier and cusp-holomorphy theorem are imported.

## What is good

The single embedded lattice carrier, bilinear-first convention over `Z`, discriminant-form and gluing theory, correction of the proper/full binary mass relation, Conway–Sloane dyadic calculus, exact Nikulin boundary warnings, and removal of the unsupported mass formula are all excellent.

## Recommendation

Either provide a real algebraic-group supplier for 7D or move that construction, type the Nikulin condition tables, and split semantic records from downstream completeness certificates. After that the roadmap will have a clean and trustworthy boundary.

roed-math:

Follow-up in [`b74954e`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/b74954e), resolving the one open note from my previous comment. **The nondegeneracy convention is now visible in the types**: `discriminantPairing` and `discriminantQuadraticForm` carry `[L.IsNondegenerate]` written inline — inline rather than as a section variable, because a `sorry`-bodied `def` whose type does not mention the instance silently drops it, which is exactly how the convention went invisible in the first place. A comment on each records the reason so the binder is not "simplified" away later. `discriminantBilinearModule` and `discriminantQuadraticModule` pick the instance up from the section as before.

mrdouglasny:
:robot: **Claude Opus 5**, supervised by Michael Douglas.

A review from the applications side rather than the formalization side: what this roadmap
gives mathematical physics, and where the boundary as drawn would leave a physics consumer
stranded. I have not checked the Nikulin or Conway–Sloane statements against their sources —
other reviewers are doing that — so nothing below is a correctness claim. The concern is
coverage and ownership, and it bears directly on two of the listed human review priorities:
the successor split, and theta-series ownership.

**Summary.** The framing of this roadmap is LMFDB plus the K3 pipeline, and that framing
undersells it. Layers 1 and 5 and milestone 6A are, close to line for line, the mathematics
of charge lattices, defect groups, and string compactification. Three gaps matter for that
audience; two of them are cheap and sit inside layers that already exist, and the third is a
naming action rather than a scope addition.

## What is already here

| Milestone | What it is, on the physics side |
| --- | --- |
| 1C, 1D — `A_L = L⋆/L`, `b_L : A_L × A_L → ℚ/ℤ`, `q_L(x̄) = B(x,x)/2` | the defect group of a charge lattice, the Dirac pairing mod 1, and the topological spin of a line operator |
| 1E, 1F — isotropic subgroups ↔ integral and even overlattices, with `A_M ≅ H^⊥/H` | the choices of global form of a gauge theory, that is the maximal mutually local sets of line operators; `H^⊥/H` is the anyon-condensation formula for what survives gauging |
| 1F — the glued lattice is unimodular exactly when `H` is Lagrangian | an absolute theory, as opposed to a relative one, exists exactly for a Lagrangian subgroup |
| 1H, 1I — the Gauss-sum invariant `sign q ∈ ℤ/8`, and Milgram | Gauss–Milgram: the chiral central charge mod 8 of the associated abelian anyon theory, equivalently the framing anomaly of the Chern–Simons theory with `K`-matrix `G` |
| 5A — the genus of an even lattice is determined by `(t₊, t₋, q_L)` | a `K`-matrix theory is determined by signature, defect group and spins, and by nothing else |
| 5D — `U ⊕ T` and `E₈(−1) ⊕ T` stabilization | adding invertible sectors does not change the anomaly |
| 5E and 5J — `O(T) → O(q_T)` is surjective for indefinite `T` | every duality action on the one-form symmetry is realized by an honest lattice isometry; this is the input to the statement that S-duality permutes the global forms of `N = 4` super Yang–Mills |
| 6A — the indefinite unimodular classification | the Narain theorem: `II_{d,d} ≅ U^d` is the uniqueness of the Narain lattice and hence of the T-duality moduli space, and `II_{16+d,d} ≅ U^d ⊕ E₈²` for `d ≥ 1` is why the two ten-dimensional heterotic strings become one theory on a circle. It also covers `II_{25,1}` |
| 6E — the 24 Niemeier lattices | heterotic on `T²⁴`, CHL orbifolds, moonshine |
| 8G — the vector-valued law on discriminant cosets | the characters of a lattice conformal field theory, with their `S` and `T` matrices |

One convention decision is worth endorsing explicitly, since the discriminant-form convention
is on the review-priority list. Choosing `q_L : A_L → ℚ/ℤ` with `q_L(x̄) = B(x,x)/2`, rather
than Nikulin's `ℚ/2ℤ`, is the convention physics uses: `q_L(μ)` *is* the topological spin mod
one, and 8G's characters `e^{2πi q_L(μ)}` are then the `T`-matrix of the conformal field
theory with no factor to insert. The roadmap notes that the half-norm convention is what makes
8G come out clean; it is also what makes the resulting Lean statements directly quotable
outside mathematics. Keep it, and keep Nikulin's convention reachable only through the
dictionary, exactly as written.

## Three gaps

**1. The duality groups themselves have no owner.** The scope section lists "automorphism
groups of indefinite lattices, and Borcherds' method" among the subjects with no owner. But
`O(Γ^{d,d}; ℤ)` *is* the T-duality group, `O(Γ^{16+d,d}; ℤ)` the heterotic duality group, and
`O⁺(II_{25,1})` the object behind the fake monster algebra and the one-loop heterotic
amplitude. What this roadmap proves about them is 4F, that `O(L)` is infinite.

Deferring them is the right call for this PR; leaving them unowned is not. The revision's own
standard — a prose reference to a removed milestone is not a closed dependency — applies to
consumers as well as to suppliers, and an unowned subject on the boundary list is
indistinguishable from an oversight. I would ask that the boundary list name a successor here
the way the adelic material was given `OrthogonalTamagawaAndLatticeMass`; something like
`IndefiniteLatticeAutomorphisms`, owning reflection subgroups, the Eichler criterion for
`O(L)`-orbits, and Borcherds' fundamental domains. That is one row in a table, not a scope
increase, and it converts "nobody has thought about this" into "deliberately deferred to a
named place".

**2. Layer 8 cannot express a Narain partition function.** Every theta milestone assumes
`PosDef`, and the conventions section gives the right reason: for an indefinite form the
series diverges. The physics object is the Siegel–Narain theta, which is defined for
indefinite `L` *given a point of the Grassmannian* — a maximal positive definite subspace
`v ⊆ L ⊗ ℝ`, that is a point of `O(p,q)/(O(p) × O(q))`, which is the compactification moduli
space. No Grassmannian appears anywhere in the roadmap, so at present Layer 8 covers lattice
CFT characters but not the partition function of the compactified string.

The proposed addition is one milestone:

> **8H. The Siegel–Narain theta.** For nondegenerate `L` of signature `(p,q)` and a maximal
> positive definite `v ⊆ L ⊗ ℝ`, define
> `Θ_{L,v}(τ) = ∑_{x ∈ L} exp(πiτ·β(x_v, x_v) + πi τ̄·β(x_{v^⊥}, x_{v^⊥}))`.
> Prove convergence and the `S` and `T` laws, the coset refinement over `A_L`, and
> equivariance in `v` under `O(L)`.

This needs no supplier that is not already imported: it runs on the same Poisson summation
Layer 8 takes from L-functions, and it keeps that dependency one-way, since nothing analytic
is imported back. It also subsumes 8B, 8E and 8G as the `q = 0` case, so it is a
generalization of the existing milestones rather than a parallel track, and it is the natural
later home for theta lifts. If it is judged out of scope, the same treatment as gap 1 would
do: name the owner.

**3. Two cheap additions to Layer 1.**

- *Van der Blij.* For integral nondegenerate `L` and a characteristic vector `w`, meaning
  `β(w,x) ≡ β(x,x) (mod 2)` for all `x ∈ L`, one has `τ(L) ≡ β(w,w) (mod 8)`. Characteristic
  vectors already appear in the roadmap, but only inside the proof of 6C; they are not an
  exported object. This is the odd-lattice twin of 1I, which is already a milestone, and on
  the physics side it is the spin-structure and Wu-class statement underlying fermionic
  abelian topological field theories and spin Chern–Simons. Since 5A already carries the odd
  analogue through `b_L` (Corollary 1.16.3), the odd branch is half built; this completes it
  at low cost.
- *Witt triviality and the existence of a Lagrangian subgroup.* A nondegenerate finite
  quadratic module `(A,q)` admits a Lagrangian subgroup exactly when it is Witt-trivial, and
  `sign q = 0` is then necessary but not sufficient. 1G defines Lagrangian and 1F already
  proves that gluing along a Lagrangian subgroup yields a unimodular lattice; what is missing
  is the existence criterion in the other direction. This is the gapped-boundary criterion for
  an abelian anyon theory, and the reason a chiral theory admits no gapped boundary. Naming
  `sign` as a homomorphism `W(q) → ℤ/8` on the Witt group would additionally give 5D and 5E a
  common invariant-theoretic frame; 5E already refers to "Witt-type statements" without a Witt
  group in the document.

## On the split, from the applications side

The split is architecturally right and I would not argue against it. Two remarks on what it
costs, both bearing on the first review priority.

The moved-out 7I, that the genus of `E₈²` has exactly two classes, is the statement that
there are exactly two heterotic strings in ten dimensions. It is the one theorem in this
roadmap an outside physicist is most likely to ask for by name. 6D stays here and proves
everything except exhaustion, which is the correct place to cut, but 6D and 7I should be
tracked as a pair in `OrthogonalTamagawaAndLatticeMass`'s charter so that the completeness
half does not fall between the two documents.

Second, the mass formula is Siegel–Weil, and Siegel–Weil is the backbone of the Narain
averaging and ensemble holography programme of the last few years. It is worth recording in
the boundary list that even the successor does not reach that application: the successor's
masses are positive definite by construction, whereas the Narain average is the indefinite
Siegel–Weil, which needs 8H above together with the Eisenstein series. As the list reads now,
"the analytic proof of the mass formula, which uses Siegel Eisenstein series, the Weil
representation and Siegel–Weil" sounds as though Siegel–Weil lands next door. It is two
roadmaps away, and saying so costs one clause.

## One thing that should be an explicit non-goal

The Dirac pairing on a four-dimensional electric–magnetic charge lattice is antisymmetric,
and this roadmap is symmetric-bilinear throughout. Integral symplectic lattices, `Sp(2n, ℤ)`,
and Siegel modular forms — the electric–magnetic duality and intermediate-Jacobian story — are
a different roadmap, not a gap in this one. I would add a line to the boundary list saying so,
so that a future contributor does not try to graft the symplectic case onto this carrier. The
symmetric case, which is Chern–Simons `K`-matrices, six-dimensional defect groups and the
Narain lattice, is the right and complete target for this document.

## The asks

1. Name an owner for automorphism groups of indefinite lattices, rather than leaving the
   subject on the unowned list.
2. Add 8H, the Siegel–Narain theta, or name its owner. It requires no new supplier and keeps
   the L-functions dependency one-way.
3. Add van der Blij and the Lagrangian-existence criterion to Layer 1.
4. Record integral symplectic lattices as an explicit non-goal.
5. Track 6D and 7I as a pair in the successor's charter, and note that indefinite Siegel–Weil
   is not the successor's either.

Items 1, 3, 4 and 5 are edits to lists this document already maintains. Only item 2 is new
mathematical content.

roed-math:

The applications-side review is addressed in [`41fead8`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/41fead8). Thank you — it is unusual and useful to see the consumer readings written down. The five asks, in order:

**1. Named.** `IndefiniteLatticeAutomorphisms` is now an owner-table row (reflection subgroups, Eichler criteria for `O(L)`-orbits, arithmeticity and fundamental domains, Borcherds' method), and the subject is off the unowned list.

**2. Named rather than added.** `IndefiniteThetaAndSiegelWeil` owns the Siegel–Narain theta with its Grassmannian parameter, the coset refinement, `S`/`T` laws, `O(L)`-equivariance, theta lifts, and indefinite Siegel–Weil. Your exact worry — that the current wording makes Siegel–Weil sound one door away — is answered in the charter paragraph, the analytic-mass boundary bullet, and both Layer-8 end blocks: `OrthogonalTamagawaAndLatticeMass` is positive definite throughout, and Siegel–Weil is two boundaries away, not one.

**3. Both additions made — each with a correction to the proposed statement.** *Van der Blij* is milestone 1L with `IsCharacteristicVector` and an integer-signature `Int.ModEq 8` congruence — **with unimodularity in the hypotheses**, because the nondegenerate-only statement is false: for `⟨2⟩` *every* vector is characteristic (all form values are even), the signature is `1`, and `β(w,w)` is always even. The witness is a Lean rejection test, a hard-theorems row, and a worked example; 6C now cites 1L instead of an unexported notion. *The Lagrangian half*: `FiniteQuadraticModule.IsMetabolic` and `gaussSign_eq_zero_of_isMetabolic` are added, and the full equivalence "Witt class zero ↔ metabolic" goes to the named successor `FiniteQuadraticModuleWittTheory` rather than a tautological local Witt group. Here too a proposed shape was false: Lagrangian through the *bilinear* form alone fails — the discriminant form of `A₁ ⊕ A₁` has a subgroup with `H = H^⊥` and vanishing bilinear restriction whose quadratic form takes the value `½`, Gauss sum `2i`, `sign = 2`. Lagrangian is therefore `q`-isotropic ∧ self-orthogonal, with that trap pinned by its own rejection test.

**4. Added** as a boundary row: alternating integral lattices, `Sp(2n, ℤ)`, the alternating Dirac pairing, and Siegel modular forms are a different subject, not to be grafted onto this symmetric carrier.

**5. Paired.** The `OrthogonalTamagawaAndLatticeMass` charter now states that 6D constructs and distinguishes the two rank-16 candidates, its 7I proves exhaustion, and only the pair classifies rank 16 — with your heterotic reading in one clause — plus the explicit note that the successor's masses never reach the indefinite Siegel–Weil.

Build green in the dependency-complete stack; nothing from the named successors is imported.

CBirkbeck (changes requested):
Reviewed head: `b74954eb74c98b287036a1a1a068a077db184f1f`; re-checked against the current head `41fead8` (note at the end).

## Verdict

**Request changes to resolve the theta-series boundary.**

The difficult lattice arithmetic is now in very good shape. The mass formula, dyadic smooth group scheme and Eichler/strong-approximation results have been moved to a genuine successor; Nikulin's boundary conditions and the certificate semantics have been made explicit. I do not see a new error in those parts.

The remaining problem was introduced by the new Theta Series roadmap #286.

## The current ownership statement is now false

This roadmap's §*Scope* lists modularity of theta series among the subjects that "have no owner and are not part of this roadmap" — "Calling that endpoint modularity would need two things this portfolio does not have" — and Layer 8 closes with "§*Scope* records that none of the three has an owner."

That is no longer accurate: PR #286 is precisely a roadmap for the modularity of lattice theta series, and it is mentioned nowhere here.

More importantly, Layer 8 here currently owns:

- the analytic theta series;
- convergence and holomorphy;
- its $q$-expansion;
- direct sums and twists;
- the rank-one comparison;
- the dual/covolume comparison;
- the scalar $S$-transformation;
- the $T$-translation;
- the vector-valued discriminant-coset transformation.

PR #286 owns the same objects and transformations, and PR #248 also owns generic lattice Poisson summation and the Gaussian theta transform.

These are not merely three applications of one theorem. They would produce competing definitions and public APIs for the same theta function and the same Poisson formula.

## Required resolution

Choose one canonical owner before merging this layer. My preferred division is:

- **Integral Lattices:** the rational lattice carrier, dual lattice, discriminant group and form, shells and representation numbers, together with the rational-to-real comparison;
- **Theta Series #286:** the real analytic theta and coset-theta functions, Poisson summation, transformations and modularity;
- **L-functions #248:** the mixed-embedding and Mellin specializations used in number-field $L$-functions.

Under that split, replace Layer 8 here by a short imported contract:

1. construct the real realization of a positive-definite integral lattice;
2. prove its dual, covolume and discriminant data agree with the algebraic carrier;
3. apply the theta-series supplier;
4. transport the coefficients back to the shell counts of this roadmap.

This also resolves the two $q$-parameter conventions cleanly. The general integral-lattice theta can use
$$
q=e^{\pi i\tau},\qquad
\Theta_L=\sum_k r_L(k)q^k,
$$
while the even-lattice modular form uses
$$
Q=e^{2\pi i\tau},\qquad
\Theta_L=\sum_m r_L(2m)Q^m;
$$
the supplier should state the conversion rather than the two roadmaps defining separate theta objects.

## What is otherwise now satisfactory

The current roadmap has correctly:

- one embedded rational lattice carrier;
- a bilinear-first integral convention which does not divide by $2$;
- duality, discriminant forms and overlattice gluing;
- local Jordan and dyadic symbol theory;
- separate class, genus and spinor-genus carriers;
- a precise division between positive-definite finiteness, binary indefinite theory and the successor's high-rank strong-approximation results;
- lattice-side local density inputs without claiming the mass formula;
- typed Nikulin existence and embedding conditions;
- semantic LMFDB records separated from completeness certificates;
- the $E_8$, $D_{16}^{+}$, Niemeier and K3-facing applications assigned to the correct layers.

Once the theta-series ownership is synchronized with #286 and #248, I would approve subject to the declared supplier order.

---

**Re-check at `41fead8`.** The commit after the reviewed head names three successors (`IndefiniteLatticeAutomorphisms`, `IndefiniteThetaAndSiegelWeil`, `FiniteQuadraticModuleWittTheory`), adds characteristic vectors with van der Blij's congruence (1L) and Lagrangian/metabolic finite quadratic modules with the Gauss-sign vanishing (1G/1H), and pins the symplectic boundary. I checked the new witnesses — `⟨2⟩` against dropping unimodularity, the discriminant form of `A₁ ⊕ A₁` against bilinear-only isotropy, and `q_θ^{(5)}(5)` with `(θ|5) = −1` against the converse — and they are correct. The theta-series boundary is unchanged: the owner table still sends Poisson summation and the Gaussian theta to L-functions, Layer 8 still owns the transformation laws, and #286 is mentioned nowhere, so the request above stands.

CBirkbeck (changes requested):
Request changes.

The revised separation of the mass formula, strong approximation and the dyadic smooth group scheme
is sensible. The remaining blocker is the Theta Series dependency.

The README currently says that Theta Series consumes foundational material from this roadmap and
also supplies every theta-series theorem used here, calling the dependency “mutual by design”.
However, the open Theta Series target already imports

    TauCetiRoadmap.IntegralLattices.Suggested.

If Integral Lattices then imports Theta Series to use those results, Lean gets an import cycle.
At present the cycle is only hidden because this `Suggested.lean` does not import or check the
Theta Series declarations that the README says it consumes.

Please replace the mutual dependency by an explicit acyclic split. For example:

* `IntegralLattices.Foundation`: the carrier, duality, discriminant forms, local/Jordan theory,
  gluing and explicit E8/Niemeier reference lattices needed by Theta Series;
* `ThetaSeries`, importing that foundation;
* `IntegralLattices.ThetaConsequences`, importing Theta Series, for any shell formulas or low-rank
  consequences which should remain under the lattice project.

Another clean option is to put all theta-dependent conclusions in the Theta Series roadmap and keep
this roadmap entirely algebraic and arithmetic.

The PR description also needs synchronizing with the current README:

1. it still says that L-functions supplies Poisson summation and the real Gaussian transform, while
   the current README assigns both to Theta Series;

2. it says in one place that Cho's dyadic density remains here, while the current README assigns the
   smooth integral group scheme and the closed dyadic-density formula to
   `OrthogonalTamagawaAndLatticeMass`.

Please make the dependency table, ownership discussion, human-review priorities and validation
instructions agree with the final acyclic split.

CBirkbeck (changes requested):
## Mathematical review: dyadic and low-rank corrections required

*AI-assisted review, posted by ChatGPT at Chris Birkbeck's request. Reviewed commit: `18e075a2ab19c0b88b795aeec74338a332327c5a`; the PR head was rechecked before posting. This reviews the normative roadmap and target statements. I have not run a local Lean build.*

The embedded integral-lattice carrier, the half-norm discriminant-form convention, and the distinction between verifying supplied representatives and proving classification completeness should be retained. The move of strong approximation, the global Tamagawa volume theorem and the mass-formula assembly to `OrthogonalTamagawaAndLatticeMass` is also sensible. This review does not ask to put those results back here.

There are, however, six mathematical issues in the remaining specification that need repair.

### 1. The minimal dyadic realization does not always determine a unique determinant square class

**Location:** `TauCetiRoadmap/IntegralLattices/Suggested.lean`, `padicDiscriminant`, and README milestone 5B.

The claim that the minimal-rank p-adic realization `K(q_p)` is uniquely determined up to isometry, with a choice-independent determinant square class, needs the dyadic exception. It is false as an unconditional helper.

**Counterexample:** over `ℤ₂`, take the rank-one even lattices with Gram matrices `(2)` and `(10)`. Both discriminant groups have order two. On the class of half a basis vector, the half-norm quadratic values are respectively `1/4` and `5/4`, which agree in `ℚ₂/ℤ₂`. Thus the finite quadratic forms are isomorphic. The determinants differ by the factor 5, which is not a square in `ℚ₂ˣ`; hence the lattices are not isometric and their determinant square classes differ.

**Requested change:** restrict the dyadic choice-independence statement to the appropriate nonexceptional case, or retain the actual realization/choice where uniqueness is unavailable. The current `NikulinExistenceConditions` already switches off the dyadic determinant condition in the exceptional order-two-summand case; retain that distinction and make the helper interface agree with it. Add the `(2)` versus `(10)` example as a rejection test.

### 2. The dyadic parity type must be defined after dividing out the scale

**Location:** README milestone 3I, the Jordan constituents and Cho density data.

The text writes a constituent as `L_i = 2^i M_i` but then reads type I from the unnormalized norm ideal being the whole coefficient ring. For a modular constituent the parity type is the type of the normalized unimodular constituent `M_i`; equivalently the norm ideal is compared with the constituent's scale, not always with `ℤ₂`.

**Counterexample:** the lattice `(2)` is a type-I constituent of scale 2. Its norm ideal is `2ℤ₂`, not `ℤ₂`, so the stated unnormalized criterion misclassifies it.

**Requested change:** define type I/II on the normalized constituent and propagate this correction through the bound/free distinctions and the ranks/counts appearing in the density exponent. This is the modular-lattice parity convention in Cho, *Group schemes and local densities of quadratic lattices in residue characteristic 2*, arXiv:1210.7625. A rank-one constituent at nonzero scale is an essential regression, since scale-zero examples do not reveal the problem.

### 3. The worked value of the dyadic exponent contradicts its displayed formula

**Location:** README milestone 3I, the exponent

```text
N = t + Σ_{i<j} i n_i n_j + Σ_i d_i - b + c.
```

Here c is the sum of the ranks of the type-II constituents. For the unimodular type-II hyperbolic plane U there is one rank-two constituent at i=0. Thus `t=b=0`, `c=2`, and the indexed terms vanish. The displayed formula gives **N(U)=2**, not the requested worked value N=0.

**Requested change:** correct the worked value and recompute the other low-rank dyadic checks from the precise definitions, keeping Cho's exponent and the chosen density normalization visible together. If a differently normalized exponent was intended, give its explicit conversion instead of using the two normalizations under the same symbol. The worked checks must not contradict the formula they are intended to validate.

### 4. The split binary torus has an infinite-volume modulus direction

**Location:** README milestone B7, finite covolume for the binary torus, including the split quadratic-algebra branch.

The full-adelic finite-covolume assertion is false in the split case.

For the hyperbolic plane, the quadratic algebra is `A_Δ = ℚ × ℚ`, and its algebraic norm-one torus is

```text
T = {(t,t⁻¹)} ≃ G_m.
```

Consequently

```text
T(𝔸)/T(ℚ) ≃ 𝔸_ℚˣ/ℚˣ.
```

The idele modulus descends to a surjection onto `ℝ_{>0}`, leaving an infinite-volume `dt/t` direction. Being norm one in the split quadratic algebra is not the same as having global idele modulus one. Finiteness of the corresponding ideal-class group does not remove this real direction.

**Requested change:** restrict the full finite-covolume assertion to the appropriate anisotropic/nonsplit torus, and treat the split branch using the kernel of the adelic modulus or a quotient by the split real direction. State the measure and quotient actually used. Preserve the split binary arithmetic branch, but do not apply the anisotropic-torus volume statement to it.

### 5. The adelic spinor-genus class-group formula needs an applicability/rank hypothesis

**Location:** README milestone 4C and the identification involving

```text
J / (ℚˣ J² J_L).
```

This cannot be claimed finite and identified with the proper spinor genera without separating the small ranks and checking the rational/local spinor images.

**Counterexample:** for the rank-one lattice `(1)`, every local special orthogonal group is trivial. With the stated spinor-image convention, the quotient above reduces to `J/(ℚˣ J²)`, which is infinite. For example, valuation-zero finite ideles retain the product of unit square classes at infinitely many odd primes; the global rational scalars do not eliminate these independent classes. The rank-one genus itself has only one class.

**Requested change:** state the exact image groups and the rank/isotropy assumptions under which this adelic quotient computes the desired spinor genera. Treat ranks zero and one directly and rank two through its torus description, rather than applying a higher-rank formula indiscriminately. The rank-one lattice should be a compulsory rejection test.

### 6. The good-prime local-density product has an unremoved factor of two at every prime

**Location:** README milestone 7C, full-orthogonal-group local densities and their Euler-product interpretation.

With the current full-O normalization, the displayed good-prime densities satisfy `α_p(L) → 2`. Therefore the raw reciprocal factors satisfy `α_p(L)⁻¹ → 1/2`. Their ordinary prime-ordered partial products tend to zero, rather than to the nonzero Euler product needed in a mass calculation. Finitely many bad-prime corrections cannot remove an extra factor of two at every good prime.

**Requested change:** make the normalization dictionary explicit. With the current full-O density, the normalized good-prime factors involve `2/α_p(L)`, followed by the precise comparison with the successor's mass convention. State convergence/nonvanishing only in the justified rank regimes, retaining the extra qualifications in low rank and for split rank-two forms. Do not describe the raw product as a nonzero standard Euler product: it degenerates to zero.

### Synchronize the PR description and integration state

The PR body still describes an older theta-ownership/L-functions dependency and an earlier placement of the dyadic evaluation. Please synchronize it with the current normative files and their actual imports. The PR also currently has merge conflicts, so an updated dependency-complete build is needed after resolving them; that integration task is separate from the mathematical corrections above.

### Acceptance tests and scope

The correction pass should explicitly test the dyadic lattices `(2)` and `(10)`, a rank-one constituent at scale 2, the unimodular hyperbolic plane, the split binary torus, the rank-one genus, and the good-prime asymptotic of the chosen density. These small cases detect each problem before the main Nikulin, genus-symbol or mass-interface developments consume the helpers.

Retain the existing scope split: this review is not requesting a full Niemeier classification, a new proof of strong approximation, or the global mass formula inside this PR. The aim is to make the local and low-rank interfaces that remain here correct and compatible with the named successor.

roed-math:

Thank you — all six are real, and each is now corrected at its milestone with the witness you gave pinned as a test. The corrections are in [`77c3acb`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/77c3acb); the integration is separate, in [`445654f`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/445654f) and [`a7a5a9c`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/a7a5a9c).

**1. `discr K(q₂)`.** The conventions and 5B now state uniqueness of `K(q_p)` with Nikulin 1.9.1's exception — unique at odd `p`, and at `p = 2` exactly when `q₂` has no summand `q_θ^{(2)}(2)`; in the exceptional case the two realizations differ by the factor 5 — which is the same condition under which `NikulinExistenceConditions.dyadic_boundary` already reads the value. `padicDiscriminant` is documented as a chosen realization's class; `padicDiscriminant_eq_squareClass_gramDet` states the choice-independence with `p ≠ 2 ∨ ¬ HasCyclicTwoSummand` as a hypothesis (on the global carrier, since a rank-`l` even lattice with the right `p`-part localizes to a `K(q_p)`); and `padicDiscriminant_not_determined_by_cyclicTwoSummand` pins `⟨2⟩` against `⟨10⟩`: isometric 2-primary parts, both of length 1, Gram determinants 2 and 10 with different square classes in `ℚ₂`.

**2. Parity type.** 3I now defines the type of `L_i = 2^i M_i` as that of the unimodular rescaling `M_i`, i.e. `𝔫(L_i) = 𝔰(L_i)` against `𝔫(L_i) = 2𝔰(L_i)`, citing Cho Def. 2.1(b), with `⟨2⟩` at scale 2 as the mandatory test and a note that every scale-zero example passes the wrong criterion.

**3. The exponent.** You are right that the formula was transcribed correctly and the worked value was not: for a type II unimodular lattice of rank `n`, `N = c = n`. For `U`, `N = 2`, and in 7C's count (the bilinear congruence `XᵀAX ≡ A mod 2ʳ`) `α_2(U) = 2^{N−1}·#O₂⁺(𝔽₂) = 4`, which the direct count confirms (`2^{r+2}` solutions). The `N = 0` I had written is the exponent of the quadratic-form count (`α_2^q(U) = 1`), off by exactly `2^c` for a type II unimodular lattice — that is what the `+c` in `N_Q` converts. 3I now says which normalization `N` belongs to, gives the conversion, and recomputes `⟨1⟩` (`N = 1`, `α_2 = 4`), `⟨2⟩` (`N = 2`, `α_2 = 8`), `⟨1⟩ ⊕ ⟨2⟩` (`N = 2`) and `E₈ ⊗ ℤ₂` (`N = 8`) from the definitions.

**4. The split torus.** B7 is split by B1's branch. Nonsquare `Δ`: `T_L(ℚ) \ T_L(𝔸)` is compact, proved from B4 and B5 by fibring over the finite class set with fibres `T_L(ℤ̂) × T_L(ℝ)` modulo the norm-one units (the Pell unit is what makes the `Δ > 0` fibre compact). Square `Δ`: `T_L ≅ 𝔾_m`, `T_L(𝔸)/T_L(ℚ) ≅ 𝔸ˣ/ℚˣ ≅ ℤ̂ˣ × ℝ_{>0}` has infinite volume, and the statement is compactness of `T_L(ℚ) \ T_L(𝔸)¹` for the modulus kernel, equivalently of the quotient by `T_L(ℝ)⁰`. `U` is the test, and the anisotropic statement is never applied to that branch. 7G/B8 only use `Δ < 0`, which is now said.

**5. Spinor genera.** 4C makes the image groups explicit (each `θ_p(K_p⁺(L))` read as the subgroup of `ℚ_pˣ` containing the squares; `θ_∞` is `ℝ_{>0}` or `ℝˣ`) and separates the ranks: finiteness and exponent 2 for rank `≥ 2` (O'Meara 92:5 at good primes); the bijection with proper spinor genera for rank `≥ 3` (91:6 and 101:8 give `J = ℚˣ·θ(J_V)` and `ℚˣ ∩ θ(J_V) = θ(SO(V))`, identifying the quotient with O'Meara's in 102:7); rank 2 through the torus, where the proper spinor genera are the cosets of `C_L²` in the class group of B2/B3 and the idelic group is `2·[C_L : C_L²]` for nonsquare `Δ` — `x² + y²` is the witness, with `θ_2(SO(L_2))` the norm group `{1,2,5,10}(ℚ₂ˣ)²` of `ℚ₂(i)`, so the group has order 2 while the genus has one class; ranks 0 and 1 directly, with `⟨1⟩` as the compulsory rejection test (`J/(ℚˣ J²)` infinite).

**6. The density product.** 7C no longer describes `∏ α_p(L)⁻¹` as a convergent Euler product: `α_p → 2`, so it degenerates to 0. The normalized factor is `α_p(L)/2`, which is Conway–Sloane's `1/std_p(f)` in their eq. (6) — the `2` in their `∏_p 2 m_p(f)` is exactly this — and Gan–Yu's `β_{L_p}`, as 7D records. The Euler products are written out (`∏ ζ^{(S)}(2i)`, times `L^{(S)}(m, χ_L)` in even rank), with convergence stated only where it holds: absolute for `n ≥ 3`; conditional to `L(1, χ_L) ≠ 0` for `n = 2` with `−det L` a nonsquare, divergent for split rank 2 (`U` again); empty for `n = 1`; and no factor of two at all in rank 0.

Every witness — `⟨2⟩`/`⟨10⟩`, the rank-one constituent at scale 2, `U`, the split torus, `⟨1⟩`, `x² + y²`, the good-prime asymptotic — is also a row of the "Hard theorems" table and an entry of the worked examples.

**Integration.** The conflict was #350: the #200 foundation was discharged in Tau Ceti and archived to `Completed/IntegralLattices`, deleting the file this PR rewrites. The merge keeps `TauCetiRoadmap/IntegralLattices` as the active expansion, and `Suggested.lean` now imports `Completed.IntegralLattices.Suggested` and states everything against its `IntegralLattice`/`FiniteQuadraticModule`, which are Tau Ceti's; the private copies of the carrier, the discriminant and overlattice modules and the acceptance suite are gone (−517 lines), and only the expansion's own vocabulary remains. Built dependency-complete at the branch's pins (`v4.34.0-rc2`, Tau Ceti `5cb65f0`, Mathlib `dc4b8d6`) with #255 materialized locally: zero errors. The GitHub `build` check is now blocked on #255 alone. One placement question for the maintainers: an active `IntegralLattices` beside the archived one, consuming it by import, is the minimal resolution; if a distinct directory name is preferred, that is a rename of the directory and module path only.

The PR description is synchronized with the current files (no `LFunctions` dependency, Cho's evaluation with the successor, six suppliers with five merged). The earlier note on the Theta Series import cycle is resolved in [`6ac14a8`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/6ac14a8), which had been committed on 2026-09-03 but not pushed until now.

CBirkbeck (changes requested):
:robot: ChatGPT follow-up review at Chris's request, against `a7a5a9c93be9e01063c38ab328faa6120af489c7`.

The six corrections in the latest author response are present in the normative text, including the dyadic minimal-realization exception, the bilinear versus quadratic density normalization, the split-torus modulus direction, and the rank restrictions on spinor-genus counts. The move onto the archived IntegralLattices certificate is also present. A further blocker remains in the binary correspondence.

## B2's displayed map is not injective on indefinite proper form classes

For positive nonsquare discriminant, B2 sends an arbitrary primitive form `(a,b,c)` to the narrow class of the **unoriented ideal**

```text
𝔞_f = aℤ + ((−b + √Δ)/2)ℤ.
```

There is no positive-leading-coefficient normalization in the statement, and no sign/orientation datum survives in the target `NarrowPic`. Changing the target from `Pic` to `NarrowPic` alone therefore does not repair the correspondence.

Your own required example `Δ = 12` disproves the displayed map:

```text
f  = x² − 3y²   gives 𝔞_f  = ℤ + √3ℤ = 𝒪₁₂,
f' = −x² + 3y²  gives 𝔞_f' = −ℤ + √3ℤ = 𝒪₁₂.
```

Thus both map to the identity in **NarrowPic as well as Pic**. But they are not properly equivalent (indeed not equivalent under any integral unimodular change): such an equivalence would make `x² − 3y²` represent `−1`, impossible modulo 3.

**Required correction:** either normalize indefinite proper classes to representatives with `a > 0`, proving existence and independence of that choice before applying the ideal formula, or give the signed/oriented ideal construction and its explicit comparison with the supplier's `NarrowPic`. Keep the `Δ = 12` test and require the two forms to land in different narrow classes. For example, an orientation-preserving rotation replaces `f'` by `3x² − y²`; its ideal is `(3,√3) = √3·𝒪₁₂`, the relevant nontrivial narrow class. This is precisely the information the unnormalized formula loses.

The B3 composition/genus dictionary and the indefinite rank-two class-count route must consume the corrected B2 map. Please also say in B1 that the **proper equivalence class** of the primitive form is independent of the oriented basis; the polynomial `f` itself is not.

One boundary clean-up: the final sentence of 7E still says this roadmap proves agreement with the archimedean factor of the measure in 7B, although 7B and that Tamagawa measure are explicitly owned by `OrthogonalTamagawaAndLatticeMass`. Keep the Euclidean volume calculation here and assign that measure comparison to the successor, so the stated one-way dependency remains genuine.

No approval has been submitted.

roed-math:

Both points are right and are fixed in [`60740e0`](https://github.com/TauCetiProject/TauCetiRoadmap/pull/256/commits/60740e0).

**B2, the indefinite dictionary.** The map for `Δ > 0` is now the oriented one: `f = (a, b, c) ↦ 𝔞_f` for `a > 0` and `↦ √Δ·𝔞_f` for `a < 0`, `√Δ` being an element of negative norm, with the statement that this is well defined on proper classes. The milestone may equally normalize to a representative with `a > 0` — existence because an indefinite form represents a positive integer primitively and a primitively represented value can be moved to the leading coefficient by an oriented change of basis, independence because two such representatives have narrowly equivalent `𝔞_f` — and either route is proved together with the comparison between them before the ideal formula is applied. Your `Δ = 12` computation is written out in B2 as the reason the sign clause is not optional: `x² − 3y²` and `−x² + 3y²` share the unoriented ideal `𝒪₁₂`, hence a narrow class, while they are not properly equivalent (no representation of `−1` modulo 3), and the rotation to `3x² − y²` lands on `(3, √3) = √3·𝒪₁₂`, the other narrow class — which is what the oriented map assigns to `−x² + 3y²`. The `Δ = 12` test now requires the two forms to land in different narrow classes, the ⚠ paragraph says a dictionary through the unoriented `𝔞_f` into `NarrowPic` is false just as one into `Pic` is, B3 consumes the oriented map explicitly, and the example is a "Hard theorems" row and a worked example. B1 now says that the proper equivalence class of the primitive form is what is independent of the oriented basis, and that `f` itself is not.

There is no Lean statement of the B2 map in `Suggested.lean` (Layer B there pins the `¬ IsSquare Δ` order constructor, the ring class field composite and the split branch), so this round is README-only and the build is unchanged.

**7E.** The last sentence is replaced: the identification of the Euclidean volume with the archimedean factor of the Tamagawa measure of 7B is stated as `OrthogonalTamagawaAndLatticeMass`'s obligation, made on its side against the volume computed here, and 7E names no such measure.

PR #273, opened 2026-08-20
chore: stop the roadmap index from conflicting between pull requests

kim-em:
This PR drops the hand-maintained import list from the root `TauCetiRoadmap.lean`, and states in
`CONTRIBUTING.md` that a roadmap pull request touches only its own directory under
`TauCetiRoadmap/`.

Of the 29 open pull requests that currently conflict with `main`, not one conflicts on roadmap
content. They conflict only on the four files that list the roadmaps: `TauCetiRoadmap.lean` (27
of them), the `README.md` list (10), and the two issue-template dropdowns (2). The last three are
already regenerated from the roadmap directories and pushed to `main` by the sync bot after every
merge, and `check_roadmap_areas.py` is not run as a pull-request check, so those hand edits were
never required in the first place. Contributors make them by copying earlier pull requests.

The import list was the dominant cause, and it was carrying no weight. `lakefile.toml` globs
`TauCetiRoadmap.*`, so `lake build` compiles every roadmap whether or not it is named in the root
file. AdicSpaces, EllipticCurves and ModularForms have been absent from the list since 1 and 7
August with the `build` check green throughout, and injecting a deliberate error into an unlisted
module still fails `lake build`. Because the list was appended to rather than sorted, every
concurrent roadmap pull request edited its last line, so each merge re-broke every other branch.

The file itself stays: it is the library root that `lakefile.toml` names, and deleting it fails the
build with `some modules have bad imports or could not be read`. With the list removed, `lake build`
completes the same 8774 jobs as before.

This does not by itself clear the 29 conflicting branches, since each still carries a diff against
the index files. Restoring those four files to the merge base on each branch clears all 29, and
that is being applied separately.

Written with Claude Opus 5, which also ran the build experiments quoted above.

🤖 Prepared with Claude Code

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,6 +74,16 @@ Apply the **`awaiting-review`** label when you open it. If you cannot apply labe
 [Permissions](#permissions) below, and in the meantime say so in the PR description or the
 [Zulip topic][zulip-topic] and someone will apply it.
 
+**A roadmap pull request touches only its own directory under `TauCetiRoadmap/`.** Four files
+list the roadmaps, and all four are derived from the set of directories that contain a
+`README.md`: the "Roadmaps" list in `README.md`, the `area` dropdown in each of the two issue
+templates, and the root `TauCetiRoadmap.lean`. The first three are regenerated and committed by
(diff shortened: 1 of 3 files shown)


roed-math (approved):
Good idea.

PR #280, opened 2026-08-25
feat: roadmap for complex tori and logarithmic transforms

Paul-Lez:
- Construct complex-torus families from fixed and varying full lattices.
- Develop torus-level cyclic affine quotients and logarithmic transforms.

Sources: Birkenhake–Lange, Brown, and Barth–Hulek–Peters–Van de Ven.

Motivating consumer: [Levent PDF](https://alpo.ge/s6.pdf)


CBirkbeck (changes requested):
*Part of a deep adversarial review of the reusable-roadmap split — PRs #279–#284, reviewed 25 August 2026. This section covers **#280 — Complex tori: varying lattices and logarithmic transforms**.*

**Head reviewed:** `805203145c0a3db0d14a686795fcd1bc961d601a`

## Verdict

**Major changes requested.**

The roadmap has successfully become reusable and it correctly distinguishes the orbit covering from the family projection. However, the central cyclic-affine classification is formulated on the wrong group. This affects the freeness criterion, the classification of twists, the logarithmic-transform parameter, and the claimed independence theorem.

## Main mathematical error: affine translations belong to the torus, not the lattice

Let

\[
T=E/\Lambda
\]

and let \(A\) preserve \(\Lambda\). An affine automorphism of \(T\) is

\[
\phi_{A,t}(x)=Ax+t
\]

with \(t\in T\), not \(t\in\Lambda\). A lattice element is zero in \(T\), so using a translation parameter in \(\Lambda\) loses the geometry.

The correct formulas are

\[
\phi_{A,t}^{\,k}(x)=A^k x+N_{A,k}(t),\qquad
N_{A,k}(t)=\sum_{i=0}^{k-1}A^i t.
\]

For \(A^m=1\), the order condition is

\[
N_{A,m}(t)=0\quad\text{in }T.
\]

Translation conjugacy changes

\[
t\longmapsto t+(1-A)b,\qquad b\in T,
\]

so conjugacy classes of order-\(m\) affine lifts are controlled by

\[
H^1(C_m,T)
  =\ker(N_A:T\to T)/\operatorname{im}(1-A).
\]

The fixed-point criterion for the \(k\)-th power is

\[
\phi_{A,t}^k\text{ has a fixed point}
\iff
N_{A,k}(t)\in\operatorname{im}(1-A^k:T\to T).
\]

The current roadmap and `Suggested.lean` instead put `t` in `Λ`, take the cokernel of `A-1` on `Λ`, and test fixed points of an affine permutation of `Λ`. Those are valid elementary statements about the set `Λ`, but they are not the classification or fixed-point criterion needed for the torus quotient.

### Decisive regression test

Take \(A=1\). Order-\(m\) translations of \(T\) are parametrized by

\[
T[m]\cong \Lambda/m\Lambda.
\]

The proposed cokernel is

\[
\operatorname{coker}(A-1)=\Lambda,
\]

which is plainly wrong.

### The normalized lattice class

If the translation is represented by \(v/m\), with \(v\in\Lambda^A\), the associated integral class is not generally

\[
v\bmod (A-1)\Lambda.
\]

Via

\[
0\to\Lambda\to E\to T\to0
\]

and the vanishing of finite-group cohomology of the real vector space \(E\), the relevant class lies in

\[
H^2(C_m,\Lambda)
  =\Lambda^A/N_A\Lambda.
\]

The roadmap must use the actual norm quotient. This is the class which should feed the logarithmic transform.

## Required rewrite of Milestone 4

Milestone 4 should be divided into two levels.

### Torus-level affine algebra

For \(T=E/\Lambda\):

- define affine automorphisms with \(t\in T\);
- prove the iterate and norm formulas;
- define the \(C_m\)-action under \(A^m=1\) and \(N_A(t)=0\);
- classify translation conjugacy by \(H^1(C_m,T)\);
- prove the fixed-point criterion in \(T\);
- check every nonidentity power for composite \(m\).

### Integral connecting class

For a chosen lift \(\widetilde t\in E\):

- prove \(N_A(\widetilde t)\in\Lambda^A\);
- prove change of lift changes it by \(N_A\Lambda\);
- identify the resulting class in \(\Lambda^A/N_A\Lambda\);
- prove its equivalence with the torus \(H^1\)-class;
- specialize to \(v/m\).

The current lattice-affine lemmas may remain as auxiliary algebra, but they cannot be the public classification.

## Further gaps

### 1. Equivariant descent needs freeness of the total action

Milestone 3 assumes a properly discontinuous action on the base and forms a quotient of the torus family. A base stabilizer can act nontrivially on the fibre. To apply the free quotient theorem from #279, the induced action on the **total space** must be free.

This is especially important for elliptic base points, where fibre translation is precisely what removes fixed points. State the total freeness hypothesis or prove it from the torus-level affine criterion.

### 2. “Local product charts” must not imply holomorphic isotriviality

A non-isotrivial family such as

\[
\mathbb C/(\mathbb Z+\tau(y)\mathbb Z)
\]

is not holomorphically locally a product with one fixed elliptic curve merely because the total-space projection is a submersion.

Distinguish:

- submersion coordinates around a point of the total space;
- smooth local triviality;
- holomorphic local triviality as a family with a fixed compact fibre.

Only the first is automatic here. Do not call the family a holomorphic fibre bundle unless an isotriviality theorem or stronger hypothesis is supplied.

### 3. Uniform proper discontinuity needs a named compact estimate

Pointwise discreteness of every lattice is not by itself the theorem used to form the total quotient. The roadmap should explicitly prove that on a compact subset of the base there is a uniform lower bound for the norm of every nonzero period, or equivalently a uniform bound on inverse period matrices. This supplies orbit-local finiteness and compact fundamental representatives.

### 4. State separation and countability assumptions on the base

The total-space Hausdorff and second-countability claims need the corresponding hypotheses on `Y`, together with continuity/holomorphy of the period data. Put them in the constructor rather than relying on the informal word “complex manifold”.

### 5. Exact normal and canonical orders need the correct injectivity argument

Divisibility of the order follows from the cyclic character. Exact order requires proving that the associated-character map into the Picard group is injective in the case at hand, for example from connected compactness of the covering space and constancy of invertible holomorphic functions.

### 6. Independence must use the correct cohomology class

The logarithmic transform is not classified by `coker(A-1)` on the lattice. The independence theorem must hold with the period family fixed and the correct class in \(H^1(C_m,T)\), equivalently its connecting class in \(\Lambda^A/N_A\Lambda\), fixed.

### 7. Expand `Suggested.lean`

The file currently validates affine permutations of `Λ`, not affine automorphisms of `E/Λ`. Add representative declarations for:

- the fixed torus quotient;
- the varying total-space quotient and both projections;
- torus-level affine automorphisms;
- the \(H^1(C_m,T)\) class;
- the integral connecting class;
- total-space freeness;
- the multiple-fibre local model.

## What should remain

- use of the standard orbit quotient;
- distinction between `q` and `p`;
- pointwise full-lattice hypotheses through `IsZLattice`;
- retention of the varying period family;
- checking every nonidentity power for composite order;
- map-level fundamental-group and homology comparisons;
- the construction-independent scope.

## Recommended disposition

The cyclic-affine core must be rewritten before the roadmap can be approved. This is not a cosmetic change: it alters the parameter space of logarithmic transforms and the proof of freeness.

---

Paul-Lez:
:robot: Addressed in 6bfc981. The cyclic-affine core now lives on T=E/Λ: its class is ker N/im(1−A), and the integral connecting class is Λ^A/NΛ. The fixed-point, freeness, independence, and A=1 regression checks use those groups. The roadmap also makes total-space freeness explicit, separates submersion charts from smooth and holomorphic local triviality, names the uniform compact period estimate, states base separation/countability, and uses the associated-character injectivity criterion for exact bundle orders. Suggested.lean now seeds the torus-level and multiple-fibre interfaces, and the PR body foregrounds the standard sources.

CBirkbeck (changes requested):
*Part of the second deep review of the reusable-roadmap split — PRs #279–#284, reviewed 26 August 2026 against the updated heads. This section covers **#280 — Complex tori and logarithmic transforms**.*

**Head reviewed:** `6bfc9818e941a30c2d0696d61e7805bbc9281e3e`

## What has been fixed

The central error from the first review has been corrected properly:

- the affine translation parameter is now `t : T`, not an element of the lattice;
- the iterate and fixed-point criteria live on the torus;
- conjugacy classes are modeled by

  $$
  H^1(C_m,T)=\ker N_A/\operatorname{im}(1-A);
  $$

- the integral connecting class lives in

  $$
  H^2(C_m,\Lambda)=\Lambda^A/N_A\Lambda;
  $$

- the `A = 1` regression gives `T[m] ≃ Λ/mΛ`;
- total-space freeness is separated from freeness of the base action;
- compact-uniform period separation is explicit;
- submersion coordinates, smooth local triviality and holomorphic isotriviality are distinguished;
- exact normal-bundle order uses the Picard-injectivity criterion.

That removes the previous major blocker.

## 1. A lattice automorphism need not induce a holomorphic torus automorphism

The roadmap currently says to let a lattice automorphism induce `A_T : T ≃+ T`. Algebraically that is fine, but the later cyclic quotient is a **complex-analytic** quotient.

For example, on

$$
T=\mathbf C/(\mathbf Z+i\mathbf Z),
$$

complex conjugation preserves the lattice and induces an additive torus automorphism, but it is antiholomorphic. A general integral automorphism can be neither complex-linear nor antiholomorphic.

The roadmap should separate:

- the abstract additive `H¹/H²` theory, which only needs `A_T : T ≃+ T`; and
- the analytic logarithmic-transform constructor, which needs an explicit

  $$
  A_E:E\simeq_{\mathbf C}E
  $$

  preserving `Λ`, or the corresponding biholomorphic torus automorphism.

Without that extra input, Milestone 5's generator need not be biholomorphic.

## 2. The Lean `H¹` carrier is only a set quotient

`AffineHOne` is currently a raw `Quotient` of a setoid. But the roadmap promises cyclic-group cohomology, functoriality, restriction and a connecting isomorphism. Here `T` is abelian, so `H¹(C_m,T)` should be an `AddCommGroup`, preferably implemented as an additive subgroup quotient.

Likewise `Suggested.lean` currently defines only

```lean
integralConnectingClass ... (v : invariantLattice A) :
  AffineHTwo A m hA
```

which is just the quotient class of an already supplied invariant vector. It does not seed the actual connecting map from a torus `H¹` class, nor the isomorphism obtained from
`0 → Λ → E → T → 0`.

Add representative declarations for:

- the additive-group structure on `AffineHOne`;
- the connecting homomorphism;
- the isomorphism `AffineHOne ≃+ AffineHTwo` under the stated vanishing theorem.

## 3. “Order-`m`” should mean exact order only when proved

The condition

$$
A^m=1,\qquad N_{A,m}(t)=0
$$

says the affine map has order **dividing** `m`. Similarly `T[m]` is the `m`-torsion subgroup, not the subset of points of exact order `m`.

Rename the general carrier to “`m`-periodic affine lifts” or “affine `C_m`-actions”, and keep the exact-order criterion as a separate theorem.

## Verdict

**Changes requested, but no longer a major redesign.** The torus cohomology is now mathematically correct; the remaining work is to distinguish additive from holomorphic automorphisms and to make the cohomology carrier genuinely group-valued.

---

Paul-Lez:
:robot: Addressed in cd824ce. The additive H¹ theory now uses an additive-subgroup quotient with its group structure, connecting homomorphism, and H¹ ≃+ H² interface. The analytic constructor separately requires a complex-linear lift (or descended biholomorphic automorphism), and all general statements now distinguish period dividing m from exact order and m-torsion from exact-order points.

CBirkbeck (approved):
*Part of the third adversarial review of the reusable-roadmap split (#279–#284) and companion roadmaps (#288, #81, #250), reviewed 26 August 2026 against the current heads. This section covers **#280 — Complex tori and logarithmic transforms**.*

**Head reviewed:** `cd824ce9a19df1f9467a18603c71a09cff1fa825`

CI status at time of posting: no workflow runs have started on this head yet; the routine build should complete before merging.

## Result

**Approve mathematically.**

All three points from the previous review have been addressed correctly.

1. The analytic constructor now starts from a complex-linear equivalence `E ≃L[ℂ] E`, and the file explicitly says that an arbitrary additive torus automorphism does not imply holomorphy.

2. `AffineHOne` is now the additive quotient

$$
\ker N_A/\operatorname{im}(1-A),
$$

not a raw set quotient. It has an inherited `AddCommGroup`, a genuine connecting homomorphism, and an additive equivalence with

$$
\Lambda^A/N_A\Lambda
$$

under the equivariant short exact sequence.

3. Period dividing `m`, exact order `m`, and the subgroup `T[m]` are now distinguished.

The fixed-point criterion checks every nonidentity power for composite `m`, and the total-space freeness statement remains separate from the base action. I found no new blocker.

---

Acyclic implementation order:

```text
ComplexManifolds
    ├── ComplexTori
    ├── AnalyticToricGeometry
    └── FuchsianOrbifolds core

ModularForms.LevelOne.JInputs

```

CBirkbeck (changes requested):

**Head reviewed:** `cd0fc6a6cb223f03149ba4436e74ac6ead755a6b`.

**Files:** `TauCetiRoadmap/ComplexTori/README.md` and `Suggested.lean` at that head.

## Assessment

**Changes requested.** The cyclic-affine algebra has been substantially repaired, but the equivariant family constructor still omits the translation data needed for one of its advertised applications.

### 1. Milestone 3 is linear, so it cannot remove stabilizers by translation

Location: `README.md`, “Milestone 3: equivariant descent and monodromy”, especially steps 1–2.

The input consists of the lattice representation and complex-linear maps `R_g(y)`. The displayed lift is

\[
 (y,v)\longmapsto(gy,R_g(y)v).
\]

This preserves the zero section. If a nonidentity `g` fixes `y`, then it fixes `(y,[0])` in the torus family. In this linear setting, freeness of the total action is therefore equivalent to freeness of the base action.

The next step says this construction includes elliptic base stabilizers removed by fibre translation. It does not: no translation occurs in either the input or the action. Requiring total-space freeness, or invoking an abstract fixed-point criterion, cannot supply the missing data.

There are two clean repairs. Either restrict this constructor to the linear situation and a free base action, with a separate affine-descent constructor; or add a holomorphic torus-valued cocycle. With `b_g(y) ∈ T_(gy)`, define

\[
 g(y,x)=(gy,R_g(y)x+b_g(y)),
\qquad
 b_{gh}(y)=b_g(hy)+R_g(hy)b_h(y).
\]

Prove the action law, holomorphy, period compatibility, and the stabilizer fixed-point criterion for that actual affine action. It is better to formulate `b` as a section of the appropriate pulled-back torus family than to demand unnecessary global vector-space lifts.

**Regression test:** a base fixed point with a nonzero torsion translation on its torus fibre must be admitted by the affine constructor. It must not be admitted as a free action by the purely linear constructor. This detects the present defect before logarithmic transforms are attempted.

### 2. Pin the punctured-family identification used for logarithmic-transform gluing

Location: “Milestone 6”, the gauge/base-change and gluing assertions.

The phrase “after the prescribed base change” needs an exact base, an exact punctured family, and a map-level identification. Killing monodromy after a covering is not an identification over the original punctured base.

For a concrete test, let `T = E₁ × E₂` be a product of elliptic curves and let `a ∈ E₁[2]` be nonzero. The involution

\[
 (z,x_1,x_2)\longmapsto(-z,x_1+a,-x_2)
\]

on `D × T` is free, including over `z=0`. Its quotient over `t=z²` has punctured-fibre first-homology monodromy

\[
 \mathrm{id}_{H_1(E_1)}\oplus(-\mathrm{id}_{H_1(E_2)}).
\]

It cannot be isomorphic over the original punctured disc to a constant family with trivial monodromy. Translation gauges do not change this linear monodromy. The double base change kills it, but any subsequent descent or gluing must still be stated and proved.

This is a request to make the existing base-change qualification precise, not a claim that every qualified version of the logarithmic-transform construction is false. State the required compatibility with the family on the complement and make the punctured identification an explicit theorem with its forward map.

### 3. Repairs that should remain accepted

The public translation parameter is now in `T`, not in its lattice. The correct periodic affine classes are

\[
 H^1(C_m,T)=\ker N_A/\operatorname{im}(1-A),
\]

and the integral connecting classes lie in `Λ^A/N_AΛ`. The fixed-point test is performed on the torus and every nonidentity power is checked for composite `m`. The fixed/varying quotient carriers and both projections are represented. These repair the central errors in the older review.

The integral connecting construction must continue to use compatible linear lattice action data on `E`; an arbitrary abstract additive automorphism of `T` alone is not that input. The current distinction in the representative algebra should be preserved in the public theorem hypotheses.

Also retain the compact uniform period-separation estimate, the distinction between the orbit covering and the positive-dimensional family projection, and the explicit refusal to infer holomorphic isotriviality from being a submersion.

### Disposition

A focused affine-family revision is needed before approval. There is no reason to redo the repaired cyclic cohomology core. Current CI is green, but it does not test the missing affine constructor.

Paul-Lez:
:robot: Addressed in `583f860`. Milestone 3 now has separate linear and affine constructors. The affine constructor carries the torus-valued cocycle `b_g(y) ∈ T_(g • y)`, with the cocycle, lattice-period compatibility, total action law, holomorphy/inverse obligations, and stabilizer fixed-point criterion; the linear constructor explicitly cannot remove a base stabilizer. The fixed-base/nonzero-torsion regression is included. Milestone 6 now fixes `D*`, `β_m(z)=z^m`, `X*`, `Q*`, and the literal pullback, and requires a named biholomorphism over `D*` with gauge, inverse, and projection equations. The `-id` monodromy involution regression distinguishes the double-cover identification from descent and final gluing. `lake env lean TauCetiRoadmap/ComplexTori/Suggested.lean` succeeds.

CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `583f86009a6841d01e758fdefec6ff41014b2f6a`, including the revised README, the affine-family declarations, and the punctured-family interface. This is a mathematical and Lean-interface review; I have not run a local Lean build.

The previous missing-affine-action finding is now resolved: the linear and affine constructors are distinct, the translation belongs to the target torus fibre, and the cocycle and period compatibility are present. The corrected cyclic cohomology and integral connecting-map interfaces should also remain accepted. Two problems remain in the new punctured-family contract.

### 1. The monodromy-killing pullback is applied to the wrong family

**Location:** README, Milestone 6, especially `puncturedBaseChangeEquiv : Q* ≃_[D*] β_m^*X*`, and its regression paragraph.

Here `X*` is explicitly the original family and `Q*` its cyclic quotient. Pulling back the original family on the right does not kill the quotient's monodromy on the left. The proposed isomorphism fails in the very regression example the paragraph specifies.

Write the two punctured discs distinctly as `D_z*` and `D_t*`, with `β₂(z)=z²`. Take

`X_z* = D_z* × (E₁ × E₂)`

and the free involution

`(z,x₁,x₂) ↦ (-z,x₁+a,-x₂)`,

where `a ∈ E₁[2]` is nonzero. Its quotient `Q_t* → D_t*` has first-homology monodromy `id ⊕ (-id)`. A pullback of the **constant original family** remains constant and has trivial monodromy. Therefore `Q_t*` cannot be isomorphic over a punctured disc to the displayed pullback of that original family.

The canonical identification that the covering actually supplies is

`β₂^* Q_t* ≃_[D_z*] X_z*`.

For general `m`, an equivariant family `X_z*` over the free rotation cover similarly gives `β_m^*Q_t* ≃ X_z*`; the map from `X_z*` to the pullback is `x ↦ (p_X(x), [x])`. This pulls back the quotient and recovers the family upstairs. It is not the displayed `Q* ≃ β_m^*X*`.

**Repair:** name both bases and both projections, put the pullback on the quotient side in the canonical covering identification, and identify its actual forward map. Separately state any logarithmic gauge used to compare two families over the original downstairs punctured disc, with the required monodromy compatibility. Descent of an upstairs isomorphism and the final regluing are additional theorems, not consequences of merely having killed monodromy after a covering.

The existing involution example should verify the corrected equation and explicitly rule out the current one.

### 2. `PuncturedBaseChangeIdentification` fixes the total spaces to complex dimension one

**Location:** `Suggested.lean`, the structure `PuncturedBaseChangeIdentification`.

Both `X'` and `BaseChangeTotal β p` are **total spaces of families**, but the structure requires

```
[ChartedSpace ℂ X']
[IsManifold 𝓘(ℂ, ℂ) ∞ X']
```

and the same model for the pullback, with the biholomorphism also modeled on `ℂ`.

For a family of elliptic curves over a disc, the total space has complex dimension **two**, not one. For the two-elliptic-factor regression above, it has complex dimension **three**. With their intended topologies these spaces cannot carry the required one-dimensional complex-manifold charts. Thus this new interface cannot express even the first positive-dimensional torus-family example, independently of the pullback-direction problem.

**Repair:** parameterize the total-space model, for example use `𝓘(ℂ, ℂ × E)` for period vector space `E`, or an appropriate general complex model. Keep the one-dimensional model for the **base** only. The pullback should also use its actual product/subtype topology and the manifold structure constructed for that pullback, rather than an unrelated topology supplied independently.

**Acceptance tests:** instantiate the corrected interface for a disc times an elliptic curve and for the two-factor involution example, with the actual projections and inherited topologies. These tests will detect both the dimension error and a misidentified pullback before the logarithmic-transform gluing proof starts.

No further rewrite of the repaired affine or cyclic-cohomology core is requested.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+   that changing the branch acts by the corresponding lattice translation and therefore gives the
+   same descended biholomorphism.
+2. Construct a named biholomorphism over `D*`
+   `puncturedBaseChangeEquiv : Q* ≃_[D*] β_m^*X*`.  Its forward map is the descended logarithmic
+   gauge; expose that forward map, the equation with both projections to `D*`, the inverse formula,
+   and holomorphy in both directions.  This is an identification only after the displayed base
+   change.  It gives no isomorphism between `Q*` and a constant family over the original punctured
+   base unless the linear monodromy is separately proved trivial.  Any descent used for the final
+   regluing is a further theorem and must retain the identification with the family on the
+   complement.
```
kim-em, line 324: :robot: Claude+Codex say: **This milestone relates the wrong families, and the surgery is never specified.**

`puncturedBaseChangeEquiv : Q* ≃_[D*] β_m^* X*` does not typecheck once the two punctured discs are distinguished. In Milestone 5 the family `X` carrying the `C_m`-action lives over the rotated disc `D_z`, and `Q = X/C_m` lives over `D_t` with `t = z^m`. Pulling back along `β_m : D_z* → D_t*` needs a family over `D_t*`, so `β_m^* X*` cannot be formed; the family it makes sense to pull back is `Q*`. What the free action gives, with no logarithm, is

    X* ≃_[D_z*] β_m^* Q*,   forward map x ↦ (p_X x, [x]),

over the punctured discs only (over `0` the map is the finite quotient `T_0 → T_0/C_m`). The `-id` regression example below confirms this and refutes the displayed statement: `Q_t*` has monodromy `id ⊕ (-id)` while any pullback of a constant family is constant.

More importantly, the roadmap never names the family a logarithmic transform is a surgery *of*. The classical construction (BHPV V, "Logarithmic transformations"; Bogomolov–Tschinkel, https://arxiv.org/abs/math/0303044 §3) starts from `Y → D_t`, sets `X := β_m^* Y → D_z`, acts by `γ(z, [e]) = (ζz, [e + m⁻¹ Π_Y(z^m) λ])` with `ζ = exp(2πi/m)` and `λ ∈ Λ`, and identifies `Q* ≃_[D_t*] Y*` by the gauge

    G [(z, [e])] = (z^m, [e − (Log z / 2πi) · Π_Y(z^m) λ]),

defined on a simply connected logarithm chart. Invariance is `e + Πλ/m − ((Log z + 2πi/m)/2πi) Πλ = e − (Log z/2πi) Πλ`; changing `Log` by `2πik` shifts the representative by the period `−Π_Y(t)(kλ)`, so the local formulas patch (there is no global logarithm on `D*`, and the theorem must say that they patch). The inverse is `(t, [y]) ↦ [(z, [y + (Log z/2πi) Π_Y(t) λ])]` for a local root `z^m = t`; changing the root applies a power of `γ`, and changing the logarithm by a full turn adds a period. That identification over the *original* punctured base is the whole content of "regluing" in 6.3 and 6.4; without it there is nothing to glue.

Suggested rewrite of the milestone:

1. Fix `Y → D_t`, `X := β_m^* Y → D_z`, and a `C_m`-action on `X` from Milestone 3's affine constructor with linear part `R` and torus-valued cocycle `b`. Keep general `R` in Milestone 5. For the logarithmic transform require the linear part to be holomorphically gauge-equivalent to the trivial one (in the constant model of Milestone 5, `A_E = A_Λ = 1`), because an identification of `Q*` with a family extending across `0` forces the linear monodromy of `Q*` to be trivial; the `-id` example is exactly this obstruction.
2. State the covering identification `X* ≃_[D_z*] β_m^* Q*` with the forward map above, as a theorem about any free equivariant family.
3. State the normalisation theorem the gauge silently assumes: for `R = 1`, lift `b` holomorphically to `b̃ : D → E` (the disc is simply connected); then `Σ_{j<m} b̃(ζ^j z) = Π_Y(z^m) λ` for a constant `λ ∈ Λ`, and with `f(z) = b̃(z) − m⁻¹ Π_Y(z^m) λ` and `c(z) = −m⁻¹ Σ_j j f(ζ^j z)` one has `c(ζz) − c(z) = −f(z)`, so conjugating by `(z, x) ↦ (z, x + c(z))` normalises `b` to the constant cocycle `m⁻¹ Π_Y(z^m) λ`. The invariant is `λ mod mΛ`, equivalently `b(0) ∈ T_0[m]`; identify it with the `H¹(C_m, T_0)` class of Milestone 4 for the trivial action. Require that class to have exact order `m`, or the central action is not free. This is also what makes the "class in `H¹(C_m,T)`" of 6.4 an invariant of the transform rather than of one chosen cocycle; at present that is asserted without a target.
4. State the gauge theorem `Q* ≃_[D_t*] Y*` for normalised data (or composed with the normalisation of item 3), with the explicit forward map, patching, `C_m`-invariance, and inverse, and with the meridian and logarithm conventions pinned.
5. Name the regluing input: an ambient complex manifold containing `Y → D_t` as an open fibred neighbourhood, from which the central fibre neighbourhood is removed and `Q` glued in along the identification of item 4 restricted to the collar. "Apply the open-gluing theorem to the analytic pieces" leaves the pieces and embeddings unspecified. This can stay a generic theorem rather than a specific global assembly.

For the `-id` example the acceptance check should say that item 2 holds and that item 4 is inapplicable; it cannot test item 4 positively.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+   faithfulness follows from the exact-order theorem.
+3. Prove that conjugation by translation sends `t` to `t + (1-A_T)b`. Classify `m`-periodic
+   translation-conjugacy classes by the literal additive-group quotient
+
+   `H¹(C_m,T) = ker (N_A : T → T) / range (1-A_T : T → T)`.
+
+   Implement `ker N_A` and `range(1-A_T)` as additive subgroups and `H¹` as their
+   `QuotientAddGroup`, so its `AddCommGroup` structure, quotient homomorphism, induction principle,
+   and lift theorem are available. Prove identity, change-of-representative, functoriality, and
+   restriction homomorphisms in this carrier.
```
kim-em, line 243: :robot: Claude+Codex say: **This builds a private copy of group cohomology that Mathlib already has.**

At the pinned Mathlib, `Mathlib/RepresentationTheory/Homological/GroupCohomology/FiniteCyclic.lean` provides `Rep.FiniteCyclicGroup.groupCohomologyIsoOdd` and `groupCohomologyIsoEven` (`Hⁱ(G,A)` as the homology of `A → A → A` with `ρ g − 1` and the norm `N`, in either order), with `groupCohomologyπOdd_eq_iff` and `groupCohomologyπEven_eq_iff` exposing the concrete `ker/range` relation; `LongExactSequence.lean` provides `groupCohomology.δ`, `δ_naturality`, and `isIso_δ_of_isZero`; `Functoriality.lean` provides `groupCohomology.map`; `Representation.norm` is the norm. Mandating "Implement `ker N_A` and `range(1-A_T)` as additive subgroups and `H¹` as their `QuotientAddGroup`" is the situation the "Use Mathlib's vocabulary" rule in the root README is about.

Suggested rewrite of 4.3, 4.5 and 4.6:

- Fix the equivariant short exact sequence `0 → Λ → E → T → 0` in `Rep ℤ G` for `G` finite cyclic of order `m` with generator `g` (for `Multiplicative (ZMod m)`, `NeZero m` from `0 < m` supplies `Fintype`). Constructing that representation from `A_T` with `A_T^m = 1`, and from compatible `A_Λ`, `A_E`, is a routine but real target; so is the comparison of `Representation.norm` with the ordered sum of powers, and of Mathlib's `range (ρ g − 1)` with the `range (1 − A)` that translation conjugacy produces.
- Let `H¹(C_m,T)` and `H²(C_m,Λ)` be `groupCohomology` in degrees 1 and 2, with the finite-cyclic isomorphisms as the computation, `δ` from the long exact sequence, and the `AddEquiv` from `isIso_δ_of_isZero` (the underlying `LinearEquiv` of a `ModuleCat ℤ` isomorphism). Check the sign of the explicit `δ(v/m) = [v]` formula against Mathlib's `δ` convention.
- Vanishing of `H¹` and `H²` of the real vector space `E` is a named target here: Mathlib has only Maschke and the `Subsingleton G` case at the pin. It follows from the periodic description with `m` invertible on `E`.
- The genuinely new content stays: the iterate formula and partial norms `N_{A,k}` for every `k < m` (these are not replaced by `Representation.norm`), translation-conjugate lifts are cohomologous cocycles, the fixed-point criterion, and the `A = 1` specialisation to `T[m] ≃+ Λ/mΛ`.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+
+   Thus the action is free exactly when this membership fails for every `0 < k < m`. State the
+   prime-order simplification and check every nonidentity power for composite `m`.
+
+### Integral connecting class
+
+5. For a lift `t̃ : E` of `t`, prove `N_A(t̃) ∈ Λ^A`. Prove that changing `t̃` by `λ : Λ`
+   changes this invariant by `N_A(λ)`, and hence obtain the connecting class
+
+   `δ(t) ∈ Λ^A / N_A Λ = H²(C_m,Λ)`.
```
kim-em, line 257: :robot: Claude+Codex say: **4.5 is not defined from the milestone's inputs.** The milestone fixes only `A_T : T ≃+ T`, but `N_A(t̃)` and `Λ^A` need actions on `E` and `Λ`. `Suggested.lean` takes `AΛ`, `AE`, `AT` with equivariance equations; the definitive README does not. Fix compatible `A_Λ`, `A_E`, `A_T` (or the equivariant short exact sequence) as the input of this section.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+  family or logarithmic transform and applies that generic theory.
+- The [geometric-topology roadmap](../GeometricTopology/README.md) owns collars and boundary
+  gluing. This roadmap constructs the analytic punctured-disc gauges and proves their boundary
+  maps are smooth and holomorphic where appropriate.
+- The analytic-toric-geometry roadmap owns fan constructions and toric degenerations. This
+  roadmap treats smooth period-lattice families and finite cyclic multiple fibres; a toric
+  degeneration consumes both roadmaps.
+- The Fuchsian-groups roadmap owns orbifold bases and their quotient uniformization. An orbifold
+  torus family supplies its period representation and then consumes this roadmap's equivariant
+  descent.
```
kim-em, line 64: :robot: Claude+Codex say: Four of these roadmaps are referred to by prose only. Link them the way universal-covers and geometric-topology are linked: complex manifolds is #279, Fuchsian groups is #282 (`FuchsianOrbifolds/`), algebraic topology is #283, and analytic toric geometry is merged at `../AnalyticToricGeometry/README.md`. The character-to-Picard criterion consumed in 5.4 and 5.5 is #279 Milestone 7, item 6; cite it by milestone.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+  another equivariant universal cover.
+- The algebraic-topology roadmap owns general fundamental groups, singular homology, Wang
+  sequences, transfer, and homology of tori. This roadmap proves which maps arise from a torus
+  family or logarithmic transform and applies that generic theory.
+- The [geometric-topology roadmap](../GeometricTopology/README.md) owns collars and boundary
+  gluing. This roadmap constructs the analytic punctured-disc gauges and proves their boundary
+  maps are smooth and holomorphic where appropriate.
+- The analytic-toric-geometry roadmap owns fan constructions and toric degenerations. This
+  roadmap treats smooth period-lattice families and finite cyclic multiple fibres; a toric
+  degeneration consumes both roadmaps.
```
kim-em, line 61: :robot: Claude+Codex say: The toric roadmap is scoped to smooth toric manifolds from finite regular fans and has no toric-degeneration target, so "owns fan constructions and toric degenerations" attributes work nobody has. Drop the degeneration claim or make it a target somewhere.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+   finiteness. Pointwise discreteness alone does not discharge this target.
+2. Give the standard orbit quotient its complex-manifold structure. Prove that
+   `q : Y × E → (Y × E)/Λ` is a covering map and local biholomorphism.
+3. Descend `Prod.fst` to `p : (Y × E)/Λ → Y`. Prove that `p ∘ q = Prod.fst`, that `p` is a
+   holomorphic submersion, and construct submersion coordinates around each point of its total
+   space. Identify the fibre over `y` biholomorphically with
+   `E / LinearMap.range (Π y)`. Separately prove smooth local triviality from the compact-fibre
+   proper-submersion theorem, with every properness hypothesis visible. Call the family a
+   holomorphic fibre bundle with one fixed compact fibre only under an isotriviality hypothesis;
+   varying `τ` in `ℂ/(ℤ + τ(y)ℤ)` is the regression test which must not satisfy that conclusion.
```
kim-em, line 165: :robot: Claude+Codex say: **"The compact-fibre proper-submersion theorem" (Ehresmann) is a gap.** No roadmap in `main` or in #279/#283 owns it. It is also unnecessary: `Λ` is one global finite free module and `E` is fixed, so a single `ℤ`-basis of `Λ` gives `F_y : ℝ^d ≃ₗ[ℝ] E` sending `e_i ↦ Π(y) λ_i`, invertible by injectivity plus `IsZLattice`, smooth in `y` by coordinatewise holomorphy and smoothness of inversion on `GL(d,ℝ)`; then `[y, v] ↦ (y, [F_y⁻¹ v])` is a global smooth trivialisation `Y × ℝ^d/ℤ^d`, not holomorphic since `F_y` is only real-linear. The basis is a proof choice, not constructor input, so this does not conflict with the acceptance check on chosen bases. Ask for that construction; it also makes the canonical marking of 2.5 explicit.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+4. Prove fibre compactness, total-space Hausdorffness and second countability, existence and
+   holomorphy of the zero section, and naturality under base change from the explicit Hausdorff,
+   second-countability, continuity, and coordinatewise-holomorphy hypotheses on `Y` and `Π`.
+5. Define markings as isomorphisms from the local system of first integral homology groups to the
+   constant lattice local system. Prove change-of-marking and monodromy laws through the local-
+   system or covering API, not through unrelated matrices at each fibre.
```
kim-em, line 171: :robot: Claude+Codex say: The family is built from a single global `Λ`, so its first-homology local system is canonically constant and monodromy is trivial until Γ-descent in 3.4. Rewrite 2.5 to construct the canonical marking `H₁(T_y) ≅ Λ`, prove its naturality in `y` and compatibility with the fibre identification of 2.3, describe change of an arbitrary marking by `Aut(Λ)`, and cite #283 Stage 2.6 for the local-coefficient-system carrier and Stage 5.5 for bundle monodromy.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+   the torus family is intentional: no unnecessary global vector-space lift is required.
+3. Descend either action, prove it biholomorphic and properly discontinuous under explicit
+   orbit-local-finiteness hypotheses.  For the affine constructor, state freeness for the induced
+   action on the **total torus-family space**, not merely on the base, and prove it from
+   Milestone 4's torus fixed-point criterion at every base stabilizer. Only then form the standard
+   second quotient.  This constructor includes elliptic base points whose stabilizers are removed
+   by a fibre translation.
+4. Descend the family projection to `Y/Γ`, prove submersion charts over the free locus, and identify
+   its monodromy with `ρ`. Under the proper-submersion hypotheses, descend the separate smooth local
+   trivializations. Derive the action on homology through exterior powers.
```
kim-em, line 204: :robot: Claude+Codex say: "Identify its monodromy with `ρ`" does not pin the loop-to-deck convention. Depending on whether the lift of the positive loop ends at `g • y` or `g⁻¹ • y`, the return map is `ρ g` or `ρ g⁻¹`. The encoding conventions pin the meridian orientation but not this; pin it and prove the formula before the reversal law in 3.5. The `-id` example cannot detect the difference.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+coinvariants, and periodic cohomology spine. The fixed-point criterion follows directly from the
+affine iterate equation and is proved in the torus quotient, not asserted as a freeness field.
+
+## Milestone 5: cyclic quotients and multiple fibres
+
+Let `D` be a complex disc with a rotation `r` of exact order `m`, and let a varying torus family
+over `D` carry compatible period monodromy and affine translation data from Milestone 4. In
+addition to the additive class, fix a complex-linear equivalence `A_E : E ≃L[ℂ] E` preserving
+the period lattice in both directions and inducing `A_T`, or supply the equivalent
+biholomorphic additive automorphism of each torus fibre.
```
kim-em, line 281: :robot: Claude+Codex say: **This mixes two generalities.** Milestone 3 allows a varying linear cocycle `R_g(z)`, while this milestone switches to one constant `A_E` "preserving the period lattice" and a single `t : T` "from Milestone 4". Either consume Milestone 3's full cyclic affine cocycle, or specialise explicitly to constant `A_E` with the equivariance `A_E (Π(z) λ) = Π(ζz) (A_Λ λ)` stated. The `m`-fold translation condition for the family is the orbit sum of the sections `b_g(z)`, not the norm of one global `t`. At `z = 0`, `t = b_g(0)` and `A = R_g(0)` recover Milestone 4, and freeness there must be checked for every nonidentity power; away from `0` it follows from the free base rotation.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+   complex-linearity and lattice-preservation hypotheses, not merely from the additive
+   automorphism `A_T`. Prove its `m`-th power is the identity from `A_E ^ m = 1` (hence
+   `A_T ^ m = 1`) and the translation norm equation. Use the exact fixed-point criterion to prove
+   that the action is free; the exact order of the base rotation makes the total generator have
+   exact order `m`.
+2. Take the standard complex quotient and descend the base map `z ↦ z^m`. Prove this map is a
+   holomorphic submersion away from the central fibre and has a multiple central fibre of
+   multiplicity exactly `m`.
+3. In every transverse chart prove the local normal form `t = a u^m`, with `a` a nowhere-zero
+   holomorphic unit. Identify the reduced fibre and its quotient by the central affine action.
```
kim-em, line 293: :robot: Claude+Codex say: "Reduced fibre" has no carrier. Define it as the image of the central torus, prove it is an embedded complex hypersurface equal to `p⁻¹(0)` as a set, and express multiplicity as the vanishing order `m` of `t` along it. Also "the reduced fibre and its quotient by the central affine action" reads as two objects; the reduced fibre *is* `T_0/⟨affine generator⟩`.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+4. Use the complex-manifolds roadmap's holomorphic-line-bundle API to prove that the normal bundle
+   of the reduced fibre has `m`-th tensor power trivial. Identify its character. Prove exact order
+   only by the associated-character-to-Picard injectivity criterion from that roadmap, with the
+   connected compact covering-space and constant-invertible-holomorphic-function hypotheses
+   displayed; faithful base rotation alone proves only divisibility.
+5. Compute the canonical-bundle character from the derivative of the affine generator and give an
+   exact-order criterion using the same injectivity theorem. Prove smoothness, compactness, and
+   complex-manifold instances for the reduced quotient under their explicit hypotheses.
```
kim-em, line 301: :robot: Claude+Codex say: This does not say which canonical bundle. With the associated-bundle convention sending a group element to its action on the pulled-back fibre, the normal character is `ζ`, `K_Q|_F` has character `(ζ · det A_E)⁻¹`, and `K_F` has `(det A_E)⁻¹`; these satisfy adjunction. State both, and pin the sign.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+
+1. Construct the translation action on `E`, prove it free and properly discontinuous, and use the
+   standard additive orbit quotient. Apply the complex-manifold quotient theorem to prove that the
+   quotient is a complex manifold and the orbit map is a local biholomorphism.
+2. Descend addition, negation, and zero and prove the complex Lie-group laws. Prove compactness
+   from a lattice fundamental domain, connectedness, Hausdorffness, and second countability.
+3. For a complex linear map `f : E →ₗ[ℂ] E'` satisfying `f '' L ≤ L'`, construct the induced
+   holomorphic homomorphism of tori. Prove identity, composition, products, kernels under the
+   standard closedness hypotheses, and the equivalence criterion for continuous linear
+   equivalences carrying one lattice onto the other.
```
kim-em, line 138: :robot: Claude+Codex say: Write lattice preservation as `L.map f ≤ L'` (or an explicit subset statement), not `f '' L ≤ L'`, which mixes a set image with a submodule. "Kernels under the standard closedness hypotheses" needs the hypotheses spelled out.

TauCetiRoadmap/ComplexTori/README.md
```diff
@@ -0,0 +1,399 @@
...
+   `δ(t) ∈ Λ^A / N_A Λ = H²(C_m,Λ)`.
+
+   Construct the connecting additive homomorphism
+   `δ : H¹(C_m,T) →+ H²(C_m,Λ)` induced by the equivariant short exact sequence
+   `0 → Λ → E → T → 0`. Use the vanishing of positive-degree finite-group cohomology of the
+   real vector space `E` to construct an `AddEquiv` between these two carriers, and prove that
+   its forward map is `δ`. For normalized data `t = v/m` with `v ∈ Λ^A`, compute `δ(t)`
+   explicitly.
+6. Prove functoriality under equivariant homomorphisms, products, restriction to subgroups, and
+   base change for both cohomology carriers, with the connecting homomorphism natural in these
```
kim-em, line 266: :robot: Claude+Codex say: Say what "base change" means for the cohomology carriers (base change of what, along what).

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+    (β : B' → B) (p : X → B) (p' : X' → B')
+    [TopologicalSpace X'] [ChartedSpace ℂ X'] [IsManifold 𝓘(ℂ, ℂ) ∞ X']
+    [TopologicalSpace (BaseChangeTotal β p)] [ChartedSpace ℂ (BaseChangeTotal β p)]
+    [IsManifold 𝓘(ℂ, ℂ) ∞ (BaseChangeTotal β p)] where
+  gauge : X' → BaseChangeTotal β p
+  inverseFormula : BaseChangeTotal β p → X'
+  forward : X' ≃ₘ⟮𝓘(ℂ, ℂ), 𝓘(ℂ, ℂ)⟯ BaseChangeTotal β p
+  forward_eq_gauge : ⇑forward = gauge
+  inverse_eq_formula : ⇑forward.symm = inverseFormula
+  forward_over_base : ∀ x, baseChangeProjection β p (forward x) = p' x
```
kim-em, line 194: :robot: Claude+Codex say: **This structure cannot be instantiated for any positive-dimensional torus family.** Both total spaces are modelled on `𝓘(ℂ, ℂ)`, but a disc times an elliptic curve has complex dimension two. Parameterise the model (or use `ℂ × E` in the torus instance), take the subtype-of-product topology and pullback manifold structure on `BaseChangeTotal β p` rather than free-floating `[TopologicalSpace]`, `[ChartedSpace]`, `[IsManifold]` hypotheses on a `def`, and replace the single structure by the two interfaces from the Milestone 6 comment: the canonical `X* ≃ β^* Q*` and the translation-only gauge `Q* ≃ Y*`. Acceptance tests should instantiate them for `D × E` and for the two-factor involution.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+    (period : Y → (Λ →ₗ[ℤ] E)) {y y' : Y} (h : y = y') :
+    TorusFibre period y ≃+ TorusFibre period y' := by
+  subst h
+  exact AddEquiv.refl _
+
+/-- The fibrewise presentation of the varying torus family.  It is canonically equivalent to the
+single orbit quotient used above. -/
+abbrev FibrewiseTorusFamily {Λ Y E : Type*} [AddCommGroup Λ] [AddCommGroup E]
+    (period : Y → (Λ →ₗ[ℤ] E)) :=
+  Σ y, TorusFibre period y
```
kim-em, line 92: :robot: Claude+Codex say: The README says the total space is the orbit quotient, "not a tagged quotient", but the affine action below is built on this sigma type with `TorusFibre.cast` transport along `one_smul` and `mul_smul`. Define the action on `Quotient (periodOrbitRel period)` instead: the class of `(y, v)` goes to the class of `(g • y, R_g y v + b̃)` for any lift `b̃ ∈ E` of `b_g y`, independent of the lift because the fibre over `g • y` is `E ⧸ range (Π (g • y))`. The sigma presentation can stay only if its equivalence with the orbit quotient is itself a target.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+    TorusFibre.cast period (one_smul Γ y) (fibreLinearPart 1 y x) = x
+  fibreLinear_mul : ∀ (g h : Γ) (y : Y) (x : TorusFibre period y),
+    TorusFibre.cast period (mul_smul g h y) (fibreLinearPart (g * h) y x) =
+      fibreLinearPart g (h • y) (fibreLinearPart h y x)
+  translation : ∀ (g : Γ) (y : Y), TorusFibre period (g • y)
+  translation_one : ∀ (y : Y),
+    TorusFibre.cast period (one_smul Γ y) (translation 1 y) = 0
+  translation_mul : ∀ (g h : Γ) (y : Y),
+    TorusFibre.cast period (mul_smul g h y) (translation (g * h) y) =
+      translation g (h • y) + fibreLinearPart g (h • y) (translation h y)
```
kim-em, line 120: :robot: Claude+Codex say: Two things here. (1) This stores `ℤ`-linear maps and no analytic hypotheses, while Milestone 3 requires `E ≃L[ℂ] E` varying holomorphically; either label it the algebraic cocycle record and add the analytic interface, or match the README. (2) `fibreLinearPart` and its three axioms are determined by `linearPart` and `linearPart_period` (descend through `Submodule.mapQ`); a structure should not carry a field its other fields determine.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+    AffineHOne (AddEquiv.refl T) m hA ≃+ mTorsion (T := T) m := by
+  sorry
+
+end CyclicAffine
+
+/-! ## The integral connecting homomorphism -/
+
+section IntegralConnectingClass
+
+variable {Λ : Type*} [AddCommGroup Λ] [Module ℤ Λ]
```
kim-em, line 377: :robot: Claude+Codex say: `[Module ℤ Λ]` introduces a second `ℤ`-module structure alongside `AddCommGroup.toIntModule`; `periodAddAction` above uses the canonical one, so the two sections will not unify. Drop it.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+
+section CyclicAffine
+
+variable {T : Type*} [AddCommGroup T]
+
+/-- Iteration in the group of additive equivalences, stated explicitly because the target API
+uses the same object both as an equivalence and as a function. -/
+def addEquivPow (A : T ≃+ T) : ℕ → T ≃+ T
+  | 0 => AddEquiv.refl T
+  | k + 1 => (addEquivPow A k).trans A
```
kim-em, line 253: :robot: Claude+Codex say: Use `AddAut T` iteration (additive notation, `k • A`) or, in the cohomological formulation, `ρ (g ^ k)`, rather than a private recursive power.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+
+/-- Multiplication by `m`, used to define the `m`-torsion subgroup. -/
+def nsmulHom (m : ℕ) : T →+ T where
+  toFun t := m • t
+  map_zero' := by simp
+  map_add' := by simp
+
+/-- `T[m]` is the subgroup of elements killed by `m`; it does not mean exact order `m`. -/
+def mTorsion (m : ℕ) : AddSubgroup T :=
+  (nsmulHom m).ker
```
kim-em, line 362: :robot: Claude+Codex say: Mathlib has `nsmulAddMonoidHom` for `nsmulHom` and `AddSubgroup.torsionBy T m`, with the notation `T[m]` the README itself uses, for `mTorsion`.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+
+/-! ## The integral connecting homomorphism -/
+
+section IntegralConnectingClass
+
+variable {Λ : Type*} [AddCommGroup Λ] [Module ℤ Λ]
+
+/-- The lattice norm map `N_A`. -/
+def cyclicNormLinear (A : Λ ≃ₗ[ℤ] Λ) (m : ℕ) : Λ →ₗ[ℤ] Λ :=
+  ∑ i ∈ Finset.range m, (A ^ i).toLinearMap
```
kim-em, line 381: :robot: Claude+Codex say: This is `Representation.norm` once the cyclic representation exists.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+  forward_over_base : ∀ x, baseChangeProjection β p (forward x) = p' x
+
+/-- The punctured unit disc used by the logarithmic-transform base change. -/
+abbrev PuncturedUnitDisc := {z : ℂ // ‖z‖ < 1 ∧ z ≠ 0}
+
+/-- The exact base change `z ↦ z^m`; the punctured logarithmic gauge identifies families only
+after pullback along this map. -/
+noncomputable def puncturedPowerMap (m : ℕ) (hm : 0 < m) :
+    PuncturedUnitDisc → PuncturedUnitDisc := by
+  sorry
```
kim-em, line 203: :robot: Claude+Codex say: `PuncturedUnitDisc` should be a punctured subtype of `Complex.UnitDisc`. And `puncturedPowerMap` is `sorry` although it is `⟨z ^ m, by simpa [norm_pow] using pow_lt_one₀ …, pow_ne_zero …⟩`, or the `ℕ+` power already defined on `Complex.UnitDisc`.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+  A.toLinearEquiv.toAddEquiv
+
+/-- A complex-linear equivalence and its inverse are holomorphic. After proving lattice
+preservation and descent, this is the input which makes the induced torus automorphism
+biholomorphic; an arbitrary `T ≃+ T` does not have this conclusion. -/
+theorem complexLinearPart_biholomorphic {E : Type*} [NormedAddCommGroup E] [NormedSpace ℂ E]
+    (A : E ≃L[ℂ] E) :
+    ContMDiff 𝓘(ℂ, E) 𝓘(ℂ, E) ∞ A ∧
+      ContMDiff 𝓘(ℂ, E) 𝓘(ℂ, E) ∞ A.symm := by
+  sorry
```
kim-em, line 229: :robot: Claude+Codex say: This is `ContinuousLinearMap.contMDiff`, not a target.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+
+/-- `T[m]` is the subgroup of elements killed by `m`; it does not mean exact order `m`. -/
+def mTorsion (m : ℕ) : AddSubgroup T :=
+  (nsmulHom m).ker
+
+/-- For trivial linear part, `m`-periodic translation classes form the `m`-torsion subgroup.
+The fixed-lattice theorem further identifies this additively with `Λ / mΛ`. -/
+noncomputable def affineHOne_reflAddEquivTorsion (m : ℕ)
+    (hA : addEquivPow (AddEquiv.refl T) m = AddEquiv.refl T) :
+    AffineHOne (AddEquiv.refl T) m hA ≃+ mTorsion (T := T) m := by
```
kim-em, line 368: :robot: Claude+Codex say: `hA` is provable; it should not be a hypothesis.

TauCetiRoadmap/ComplexTori/Suggested.lean
```diff
@@ -0,0 +1,465 @@
...
+theorem totalAction_free_of_fibre_criterion {Γ B X : Type*} [Group Γ] [MulAction Γ B]
+    [MulAction Γ X] (p : X → B) (hequiv : ∀ (g : Γ) (x : X), p (g • x) = g • p x)
+    (hfibre : ∀ (g : Γ) (x : X), g • p x = p x → g • x = x → g = 1) :
+    ∀ (g : Γ) (x : X), g • x = x → g = 1 := by
+  intro g x hx
+  exact hfibre g x (by rw [← hequiv g x, hx]) hx
+
+/-- The base map in the cyclic multiple-fibre local model. The quotient construction descends this
+map and proves that its central fibre has multiplicity `m`. -/
+def multipleFibreBaseMap (m : ℕ) (z : ℂ) : ℂ := z ^ m
```
kim-em, line 463: :robot: Claude+Codex say: This wraps `z ^ m` in a name; use `(· ^ m)` directly.

kim-em (changes requested):
:robot: Claude+Codex say: Reviewed head `583f860`. Changes requested.

The findings are inline. The two that block are the Milestone 6 comment (the punctured identification is stated on the wrong family, and the family being surgered is never named, so 6.3 and 6.4 have nothing to glue) and the Milestone 4 comments (the `H¹`/`H²` carriers duplicate Mathlib's finite-cyclic group cohomology at the pin, and 4.5 is undefined from the milestone's own inputs). The rest are dependency citations, unpinned conventions, and `Suggested.lean` interface fixes.

Written by Claude (Fable 5.1) with a second adversarial pass by OpenAI Codex; the human reviewer has read it.

CBirkbeck (changes requested):

**Head reviewed and rechecked:** `583f86009a6841d01e758fdefec6ff41014b2f6a`.

## Disposition: changes required

This head is unchanged from the September 8 and September 10 reviews. The following consolidates the remaining fixes rather than reopening the repaired affine-action mathematics. Keep the separation of linear and affine family constructors, the torus-valued translation cocycle, the distinction between period dividing `m` and exact order, and the all-powers freeness criterion.

## 1. Correct the direction of the punctured pullback

**Location:** `README.md`, Milestone 6, especially `puncturedBaseChangeEquiv` and the regression paragraph.

Name the two bases separately:

```text
β_m : D_z* → D_t*,    β_m(z) = z^m
p_X : X_z* → D_z*
p_Q : Q_t* → D_t*,    Q_t* = X_z* / C_m.
```

The canonical covering identification is

```text
X_z* ≃_[D_z*] β_m^* Q_t*,
forward map x ↦ (p_X(x), [x]).
```

It pulls back the quotient downstairs and recovers the family upstairs. The current `Q* ≃ β_m^* X*` pulls back the wrong family; with the bases distinguished, `X_z*` is not even a family over the target of `β_m`.

The existing regression example also disproves the current statement. Start with the constant family `D_z × E₁ × E₂` and the free involution

```text
(z, x₁, x₂) ↦ (-z, x₁ + a, -x₂),   0 ≠ a ∈ E₁[2].
```

The quotient has punctured first-homology monodromy `id ⊕ (-id)`. Pulling back a constant original family remains constant and cannot give that quotient over the original punctured base. In contrast, pulling back the quotient along `z ↦ z²` recovers the original upstairs family.

**Required change:** correct both the equation and its bases/projections/forward map. The regression must verify the corrected covering identification, not claim to verify the current reversed one. See also the existing [Milestone 6 thread](https://github.com/TauCetiProject/TauCetiRoadmap/pull/280#discussion_r3975336775).

## 2. Separate covering descent from the actual logarithmic-transform gluing

**Location:** `README.md`, Milestone 6.1–6.4.

The covering identification above is not itself the downstairs logarithmic gauge needed for surgery. Name the family being modified and the actual regluing data. A concrete baseline is:

- an original family `Y → D_t` with a specified fibred neighbourhood in the ambient complex manifold;
- its pullback `X = β_m^*Y → D_z`;
- the specified cyclic affine action on `X` and quotient `Q → D_t`;
- a separately proved biholomorphism `Q_t* ≃_[D_t*] Y_t*` for the class of actions for which such a gauge exists;
- the collar/overlap restrictions of that map used to replace the original neighbourhood by `Q`.

For the pure-translation logarithmic transform, give the explicit local logarithm formula, its cyclic invariance, the period translation under branch changes, and the inverse constructed using local roots. Prove those local formulas patch; there is no global logarithm on the punctured disc.

For more general linear parts, state the monodromy compatibility needed for the downstairs identification. An isomorphism after a covering does not automatically descend. The `id ⊕ (-id)` example is a positive test for the covering theorem but an obstruction to identifying that quotient downstairs with a family of trivial monodromy.

The independence theorem must then refer to the precisely specified family, marking/class and overlap map. In a varying family, explain which fibre or local-system cohomology class is being held fixed; an unspecified `H¹(C_m,T)` is not enough.

## 3. Repair the total-space dimension and topology in the Lean interface

**Location:** `Suggested.lean`, `BaseChangeTotal` and `PuncturedBaseChangeIdentification`.

The structure still requires `ChartedSpace ℂ X'` and a one-dimensional complex model for the pullback total space. These are total spaces, not bases. A disc times an elliptic curve has complex dimension two; the two-elliptic-factor example has complex dimension three.

Parameterize the total-space model, using `ℂ × E` in the torus-family application or an appropriate general model. Keep the one-dimensional model for the base only. Give `BaseChangeTotal β p` its actual subtype-of-product topology and the manifold structure constructed for the relevant pullback, rather than unrelated topology/manifold instances supplied independently.

Expose separate representative interfaces for the canonical covering identification and the downstairs logarithmic gauge. Instantiate them with actual projections and inherited topologies for a disc times an elliptic curve and for the involution example; the latter must not instantiate an inapplicable trivial-monodromy gauge theorem.

## 4. Make the integral connecting-class input and Mathlib boundary explicit

**Location:** `README.md`, Milestone 4 and the corresponding cohomology declarations in `Suggested.lean`.

The opening input only fixes `A_T : T ≃+ T`, but `N_A(t̃)` and `Λ^A` require compatible actions on `E` and `Λ`. State the equivariant short exact sequence

```text
0 → Λ → E → T → 0
```

with the cyclic actions, period relations and equivariance equations as the input of the connecting-map section. Do not silently extend an arbitrary additive torus automorphism to the vector space and lattice. Retain the separate complex-linearity/holomorphy requirements for analytic applications.

Address the existing [group-cohomology reuse thread](https://github.com/TauCetiProject/TauCetiRoadmap/pull/280#discussion_r3975336781): use Mathlib's `groupCohomology`, its finite-cyclic computation and long exact sequence as the public API, instead of mandating a second public cohomology theory. The concrete kernel/norm quotients remain valuable computational descriptions, with comparison maps. Pin the sign convention relating `ρ(g)-1` to `1-A` and the explicit connecting-class formula. State/prove the vector-space vanishing used to make the connecting map an isomorphism.

## 5. Close the remaining family and dependency interfaces in the existing threads

These are additional focused repairs, not reasons to redesign the corrected cyclic algebra:

- **Milestone 2.3:** do not invoke an unassigned proper-submersion theorem. Either explicitly own/cite its full input, or use the direct real period-matrix trivialization available from the globally fixed finite free lattice. Distinguish this smooth trivialization from holomorphic isotriviality.
- **Milestones 2.5 and 3.4:** the globally marked period family has its canonical lattice marking before descent. Pin the loop-to-deck convention in the nontrivial monodromy formula after descent, including whether it uses `ρ(g)` or `ρ(g⁻¹)`.
- **Milestone 5:** consume the genuine cyclic affine family cocycle, or state the exact specialization to a constant linear part and its period-equivariance law. The central-fibre norm condition is not a substitute for the full family action law. Identify the reduced fibre, its embedding and multiplicity, and distinguish its canonical bundle from the ambient canonical bundle restricted to it.
- **Supplier links and Lean carriers:** link the precise supplier milestones; do not attribute toric degenerations to a finite-regular-fan roadmap that does not own them. Relate the sigma-type family presentation to the standard orbit quotient by an actual comparison, and distinguish an algebraic cocycle record from the holomorphic constructor.

## Revision gate

First correct the two punctured identifications and make the dimension-correct examples fit. Then reconcile the cohomology/family interfaces and existing dependency threads. Keep the already repaired torus-valued affine algebra and fixed-point tests. Reply to the original threads with the fixing commits and return to `awaiting-review` once the revision is complete.

This is a source-level review and consolidation of outstanding findings, **not a claim to have run a local Lean build**. Successful elaboration of declarations with `sorry` does not validate the current false pullback contract. No approval or merge is being issued.

PR #283, opened 2026-08-25
feat: roadmap for algebraic topology of spaces and manifolds

Paul-Lez:
- Develop van Kampen, relative and cellular homology, local coefficients, and spectral sequences.
- Add cohomology operations, duality, transfer, and manifold-facing calculations.

Sources: Hatcher, Spanier, Bredon, and Serre.

Motivating consumer: [Levent PDF](https://alpo.ge/s6.pdf)

AI assistance: Codex (GPT-5) and Codex 5.6 Sol.

PR #282, opened 2026-08-25
feat: roadmap for Fuchsian groups and orbifold Riemann surfaces

Paul-Lez:
- Develop the effective projective action, Fuchsian groups, polygon presentations, and covolume.
- Construct elliptic quotient charts, choice-independent cusp compactifications, and finite-index maps.

Sources: Beardon, Katok, and Farkas–Kra.

AI assistance: Codex (GPT-5) and Codex 5.6 Sol.

PR #284, opened 2026-08-25
feat: roadmap for high-dimensional differential topology and homotopy spheres

Paul-Lez:
- Develop James/EHP, Bott periodicity, bundle classifiers, and Pontryagin–Thom.
- Build h-cobordism and surgery theory through the classification of homotopy spheres and θ₆.

Sources: Bott, Kervaire–Milnor, Browder, and Wall.

Motivating consumer: [Levent PDF](https://alpo.ge/s6.pdf)


CBirkbeck (changes requested):
*Part of a deep adversarial review of the reusable-roadmap split — PRs #279–#284, reviewed 25 August 2026. This section covers **#284 — High-dimensional differential topology and homotopy spheres**.*

**Head reviewed:** `225d2d6f490780a26ebf31c6679ab9515c0b2671`

## Verdict

**Major changes requested.**

The split has made the scope reusable, and the roadmap now acknowledges James constructions, a full Bott proof, Pontryagin–Thom, comparison of geometric and historical `A_n/P_n`, and Wall surgery. That is a genuine improvement. I found two false statements and several foundational gaps which must be repaired before the roadmap is ready.

## 1. The proposed proof of the James equivalence is false for connected `X`

Stage 3B says that for connected CW `X`, the roadmap will use the filtration, homology calculation, **simple connectivity**, and homological Whitehead to prove

\[
JX\simeq\Omega\Sigma X.
\]

Connectedness does not imply either side is simply connected. For example,

\[
X=S^1,\qquad
J(S^1)\simeq\Omega S^2,
\]

and both spaces have fundamental group \(\mathbb Z\).

Thus the proposed proof cannot use the simply connected homological Whitehead theorem at the stated generality.

Repair this in one of two ways:

1. restrict the theorem and all its consumers to simply connected `X`; or
2. prove the full connected James theorem by also controlling `π₁` and the universal covers, or by a suitable Whitehead theorem with local coefficients.

If low-dimensional EHP calculations use \(X=S^1\), the second route or a separate low-dimensional theorem is necessary.

## 2. A framed normal bundle does not have sphere Thom space

Stage 5 says that for a framed normal bundle its Thom space is identified with “a suspended sphere”. This is false.

For a rank-\(k\) trivialized bundle over \(M\),

\[
\nu\cong M\times\mathbb R^k,
\]

one has

\[
\operatorname{Th}(\nu)
  \cong M_+\wedge S^k
  =\Sigma^k M_+,
\]

not \(S^k\) unless \(M\) is a point.

The Pontryagin–Thom representative is obtained by

\[
S^{n+k}
 \longrightarrow \operatorname{Th}(\nu)
 \cong \Sigma^k M_+
 \longrightarrow S^k,
\]

where the last map collapses \(M\) to the non-basepoint.

Correct the statement, the collapse-map target, and every subsequent inverse/naturality diagram.

## 3. Specify the convenient category for loop spaces and filtered colimits

The roadmap repeatedly uses:

- compact-open mapping spaces;
- loop–suspension adjunction;
- James constructions;
- filtered colimits of topological groups;
- classifying spaces;
- stable orthogonal groups.

Naive colimits and products in ordinary `TopCat` do not automatically have the convenient exponential and compact-generation properties used by these theorems.

Choose an exact framework:

- compactly generated weak Hausdorff spaces/k-spaces with comparison to Mathlib's spaces; or
- a restricted CW-based construction in ordinary spaces with every mapping-space and colimit theorem proved directly.

State how stable `SO`, `BSO`, loop spaces, and homotopy classes are formed in that framework. This is a prerequisite, not an implementation detail.

## 4. The Bott-periodicity proof still hides an analytic roadmap

The revised Bott section is much better, but it still assumes a large missing substrate. Bott's Morse-theoretic proof requires more than finite-dimensional Morse theory:

- finite-dimensional broken-geodesic approximations to path spaces;
- comparison of those approximations with the actual path space;
- invariant metrics and symmetric-space geodesics;
- Hessian and Morse–Bott critical manifolds;
- index and nullity calculations;
- compactness/Palais–Smale or the exact finite approximation replacing it;
- attachment of negative disc bundles;
- increasing-connectivity estimates.

The Heegaard Floer finite-dimensional Morse supplier does not provide this automatically.

Either split real Bott periodicity into a separate roadmap or add these objects and theorem dependencies explicitly.

## 5. Resolve smallness and universes for `Theta_n`, `A_n`, and `P_n`

The geometric cycles quantify over smooth manifold carriers. The collection of all such types is not automatically a small Type-0 quotient.

The current `Suggested.lean` returns unparameterized `AddCommGrpCat` objects, which silently restricts or hides the universe issue.

Choose one approach:

- universe-polymorphic groups of cycles in `Type u`, with a proved invariance under universe lift;
- a small skeleton encoded by finite triangulations/handle data;
- a bounded Euclidean-embedding code with an essential-surjectivity theorem.

The final theorem for an arbitrary `M : Type u` must explain how its sphere represents an element of the selected `Theta_6`.

## 6. Stable parallelizability versus parallelizability needs an exact theorem

Stage 6 defines a “parallelizable filling” but stores a stable tangent framing, and later promises conversion to honest parallelizability in the surgery range.

State the theorem with its actual hypotheses:

- dimension;
- nonempty boundary or spine dimension;
- orientation;
- obstruction/cancellation of a trivial line;
- compatibility with collars.

Do not use “parallelizable” before the honest trivialization has been constructed.

## 7. Pin the standard `A_n` and `P_n` definitions before using the exact sequence

The roadmap correctly says that the convenient defect-disc and collared-filling models are not definitionally Kervaire–Milnor's groups. The comparison is nevertheless still too informal.

For each historical group, provide:

- the exact cycle and equivalence relation from the source;
- the map from the convenient Lean model;
- the inverse;
- compatibility with addition;
- comparison of the boundary and defect maps.

Only after these isomorphisms may the classical exact sequence be transported to the convenient carriers.

## 8. Wall's odd-dimensional groups need formations, not a generic “Witt group of forms”

The phrase “appropriate Witt group of nonsingular quadratic data” must distinguish:

- even-dimensional quadratic forms;
- odd-dimensional formations.

The later statement that odd skew formations are stably hyperbolic is correct in the simply connected \(\mathbb Z\)-case, but the public `wallSurgeryObstructionGroup` must be defined with the dimension-dependent carrier which makes that theorem meaningful.

## 9. The sixth-stem representative and Kervaire normalization need a precise source chain

Stage 9 is much improved, but “the plumbing/Toda construction” should be replaced by exact constructions and references:

- the unstable representative of `ν²`;
- its stabilization;
- the framed six-manifold obtained under Pontryagin–Thom;
- the quadratic refinement;
- the Arf calculation;
- the comparison with the map `coker J_6 → P_6`.

This is the normalization that makes the final exact-sequence argument work.

## 10. The final orientation-preserving conclusion is not represented in Lean

`Suggested.lean` concludes only

```lean
Nonempty (M ≃ₘ S⁶).
```

The roadmap also promises an orientation-preserving diffeomorphism after matching orientations. Add the oriented carrier and a theorem that the chosen diffeomorphism preserves it. A plain `Diffeomorph` does not remember this conclusion.

## 11. `Suggested.lean` does not validate the geometric definitions

Every principal group is currently introduced as

```lean
noncomputable def ... : AddCommGrpCat := by sorry
```

with a docstring saying what the body should be. This does not type-check:

- the cycle carrier;
- the equivalence relation;
- connected sum;
- framed bordism;
- filtered colimits;
- the maps in the exact sequence.

Add representative structures for cycles and actual quotient/colimit signatures, together with the universal properties which prevent arbitrary implementations from satisfying the file.

## 12. Import the corrected algebraic-topology supplier

The roadmap consumes #283 throughout. Once #283 is corrected, `Suggested.lean` should import its representative module and `#check` the exact Hurewicz, Whitehead, duality, and finite-CW declarations rather than leaving that interface solely in prose.

## What should remain

- the reusable rather than six-sphere-specific scope;
- the separate treatment of the index-two/index-three Whitney case in dimension six;
- the geometric definition of `Theta_n`;
- the explicit James/EHP proof spine;
- the decision to prove the full Bott cycle rather than postulate two vanishings;
- construction of both directions of Pontryagin–Thom;
- comparison of convenient `A_n/P_n` models with the historical ones;
- a full Wall-surgery theorem rather than a table lookup;
- the explicit chain through `ν²`, the Kervaire invariant, and `Theta_6=0`;
- honest compactness, separation, and countability hypotheses in smooth recognition.

## Recommended disposition

Do not merge until the James and Thom-space errors are corrected and the topological category, Bott substrate, universes, and geometric group carriers are pinned. This roadmap is likely best split once more into stable homotopy/Bott and h-cobordism/surgery/homotopy-sphere roadmaps, although a single broad normative roadmap is possible if every internal boundary is explicit.

---

Paul-Lez:
:robot: Addressed in 90921a9. The James proof now handles connected fundamental groups through universal covers and local coefficients; the framed Thom space is Sigma^k M_+ with the correct augmentation; and the roadmap pins one compactly generated topology substrate plus the full Bott-specific Morse--Bott approximation. Geometric cycle groups are small explicit quotient and colimit constructions; stable versus honest parallelizability, historical A_n and P_n comparisons, parity-dependent Wall carriers, the normalized nu-squared and Kervaire chain, and the orientation-preserving endpoint are explicit. The corrected algebraic-topology supplier is now an ordered merge and literal import-check gate.

CBirkbeck (changes requested):
*Part of the second deep review of the reusable-roadmap split — PRs #279–#284, reviewed 26 August 2026 against the updated heads. This section covers **#284 — High-dimensional differential topology and homotopy spheres**.*

**Head reviewed:** `90921a951472203b76abff700acee2a934bb960b`

## What has been fixed

The README is now far more serious and mathematically explicit:

- it supplies a compactly generated-space substrate;
- fixes the connected James proof using universal covers/local coefficients;
- gives `X=S¹` as a regression test;
- corrects the Thom-space factorization;
- expands Bott through broken geodesics and Morse–Bott attachments;
- introduces a smallness strategy;
- separates stable and honest parallelizability;
- distinguishes even-dimensional forms from odd-dimensional formations;
- normalizes the unstable representative of `ν²`;
- makes oriented recognition explicit.

The prose roadmap has responded well. The main problem is that the architecture in `Suggested.lean` does not encode the geometry it claims to encode.

## 1. A pointwise orientation function is not a manifold orientation

`SmoothClosedOrientedCycle` stores

```lean
orientationAt : M → Orientation ℝ E ι
```

with no continuity or chart compatibility. The final oriented Poincaré theorem likewise accepts arbitrary pointwise functions `oM` and `oS`.

This is false. On connected `S⁶`, choose `oM` to equal the standard orientation at one point and its negative elsewhere. The type accepts this input. The orientation sign of the derivative of a diffeomorphism is locally constant, hence constant on a connected manifold, so no diffeomorphism can satisfy the displayed pointwise equation.

Use the actual shared `Manifold.Orientation I M ι` object, or `Manifold.OrientedManifold`, whose data include chart compatibility and local constancy. Do not replace it by its pointwise evaluation function.

## 2. The framing structures are vacuous

The file defines an honest tangent framing as

```lean
frameAt : M → E ≃L[ℝ] E
```

and a stable framing similarly by pointwise equivalences between fixed model spaces.

These maps do not refer to the tangent bundle. Every space admits the proposed `TangentFraming` by taking the identity equivalence at every point. In particular, the definition would “frame” `S²`, despite its tangent bundle being nontrivial.

A tangent framing must be a smooth vector-bundle equivalence

$$
TM\simeq M\times\mathbf R^n,
$$

and a stable framing must be an equivalence

$$
TM\oplus\varepsilon^r\simeq\varepsilon^{n+r},
$$

with the collar-relative compatibility stored for fillings.

The same problem propagates into `FramedCycle`, `AlmostFramedCycle`, the historical `A_n` carrier and Pontryagin–Thom.

## 3. The filling carrier has no manifold-with-boundary relation

`StableFramedFillingCycle` currently stores only:

- a code;
- an unrelated boundary code;
- a homotopy marking of that boundary code;
- a pointwise “stable framing”.

It does not say:

- the realization is a compact connected oriented manifold with boundary;
- the second realization is its boundary;
- there is a boundary inclusion/identification;
- a collar is chosen;
- the framing is product-compatible on that collar.

Consequently `stableFraming_destabilizes` does not have the hypotheses stated in the README and is not a meaningful version of the relative destabilization theorem.

This carrier should consume the actual manifold-with-boundary/collared-cobordism object from GeometricTopology rather than place two unrelated atlas codes beside each other.

## 4. `SmoothAtlasCode` does not contain the laws needed for its realization

The public structure stores charts, transitions and a groupoid-membership condition, but not:

- source/target equations;
- identity and inverse laws;
- cocycle laws;
- cover/surjectivity data;
- separation conditions.

The docstring says “the implementation adds” these laws, but they need to be fields or derived from a pre-existing gluing structure if this is the public code used to define the small skeleton. Otherwise `gluingSetoid` and its realization are unconstrained.

## 5. The Pontryagin–Thom maps are bare functions

`PontryaginThomFactorization` stores

```lean
collapse : Sphere (...) → thomSpace
augmentation : thomSpace → Sphere k
```

rather than continuous pointed maps. Since continuity at the collapsed complement and pointedness are central theorems, the carrier should at least expose `ContinuousMap`/pointed-map data.

## 6. The Wall carriers do not encode quadratic forms or formations

`EvenWallForm` currently has:

```lean
pairing : Matrix ... ℤ
quadratic : Fin rank → ℤ
unimodular : IsUnit pairing.det
```

A quadratic refinement is a function on the whole module `ℤ^r`, not on the basis index, and it must satisfy the relevant polarization and parity identities. The pairing also needs the `(-1)^q`-symmetry law.

`OddWallFormation` similarly lacks the conditions saying its displayed submodules are isotropic direct summands/lagrangians and form a genuine quadratic formation.

As written, `WallSurgeryClass` is a quotient of arbitrary matrices and functions, not Wall's `L_n(ℤ)`.

## Verdict

**Major changes requested.** The README is now strong, but the representative Lean carriers for orientation, framing, boundary fillings and surgery data are mathematically wrong. They need to be rebuilt around the actual shared geometric objects before this roadmap can merge.

---

Paul-Lez:
:robot: Addressed in e7d1d6e. The representative Lean interfaces now use a chart-compatible manifold-orientation quotient; canonical tangent-bundle trivializations; the actual intrinsic boundary with an outward-normal-first, half-open collar; a lawful finite atlas with canonical quotient topology; continuous pointed Pontryagin--Thom maps built from a normal prebundle and tubular neighbourhood; and whole-module Wall forms plus genuine formations and generated stable equivalence. The final recognition signature consumes manifold orientations themselves. The focused Lean check and static checks pass.

CBirkbeck (changes requested):
*Part of the third adversarial review of the reusable-roadmap split (#279–#284) and companion roadmaps (#288, #81, #250), reviewed 26 August 2026 against the current heads. This section covers **#284 — High-dimensional differential topology and homotopy spheres**.*

**Head reviewed:** `e7d1d6ea67be787c8c6b04f62341bc222f430df7`

## What has been fixed

The mathematical carriers that were previously wrong are now substantially repaired:

- orientations use `Manifold.Orientation`;
- tangent framings trivialize the actual tangent bundle;
- stable and normal framings use bundle trivializations;
- fillings carry collared manifold-with-boundary data;
- Pontryagin–Thom uses pointed maps and an actual normal bundle;
- Wall forms use bilinear forms, quadratic refinements and genuine lagrangians;
- `SmoothAtlasCode` now has cover, inverse and cocycle laws.

The README is now a serious roadmap.

## 1. Supplier-owned APIs are still redeclared in public namespaces

`Suggested.lean` defines, in the root/shared namespaces, substantial versions of:

- `Manifold.OrientationLift` and `Manifold.Orientation`;
- tangent coordinate changes and orientation preservation;
- `TauCetiRoadmap.GeometricTopology` collar and boundary carriers.

The README simultaneously says that Heegaard Floer owns orientation and GeometricTopology owns collars and cobordisms.

There is a legitimate need to mirror an open Mathlib or sibling interface before it lands. But that mirror should be visibly temporary:

- place it in an internal namespace;
- record an exact deletion/import gate;
- avoid presenting it as a second public owner.

Otherwise the supplier PR will collide with declarations already exported by this roadmap.

## 2. The central geometric quotient API has disappeared from `Suggested.lean`

The earlier representative file at least exposed:

- the h-cobordism relation;
- the quotient defining `Theta_n`;
- framed bordism;
- the almost-framed and filling groups;
- the Kervaire–Milnor maps and exactness;
- `Theta_6 = 0`.

The corrected file now seeds the substrate, framing and Wall carriers, but contains no representative `Theta_n` quotient or Kervaire–Milnor sequence declarations. A search of the current file finds no `Theta`, `homotopySphereGroup`, or geometric h-cobordism quotient.

Because `Suggested.lean` is the architecture check, the absence matters: CI can no longer distinguish the intended geometric group from an arbitrary `AddCommGrpCat` inserted later.

Add at least representative declarations for:

- geometric h-cobordism witnesses and their setoid;
- `HomotopySphereClass` and `Theta_n`;
- framed/almost-framed/filling quotients;
- the three Kervaire–Milnor maps;
- exactness at one displayed segment;
- the final `Theta_6` theorem.

These should use the newly corrected orientation, framing and collared-boundary carriers.

## Verdict

**Major changes requested to `Suggested.lean`; the README can largely remain.**

---

## Cross-roadmap dependency recommendation — Topology chain

```text
AlgebraicTopology (with based relative homotopy)
    ↓
HomotopySpheres
```

HomotopySpheres should import the supplier declarations and expose its geometric quotient API rather than mirror shared carriers indefinitely.

Paul-Lez:
:robot: Addressed in `4f6e2f3`. The orientation and collar supplier mirrors are now internal and have exact deletion/import gates. The representative API restores geometric h-cobordism witnesses and the literal Type-0 `HomotopySphereClass` / `Theta` quotient, the framed, almost-framed, and filling bordism quotients, all three Kervaire–Milnor maps, the displayed exactness segment, and `Theta_6 = 0`. Focused Lean elaboration passes with no errors.

CBirkbeck (changes requested):

**Head reviewed:** `4f6e2f379c1c4836d70cbb96e335fe8018a5c444`.

**Files:** `TauCetiRoadmap/HomotopySpheres/README.md` and `Suggested.lean` at that head.

## Assessment

**Major changes requested.** The revised proof plan repairs several older mistakes, but the expanded geometric prototypes still fail to encode their intended objects. Two of them cannot even model the basic disk/cylinder examples in the stated geometric setting. These are not objections to the use of `sorry` in a roadmap; they are objections to the types and the proposed mathematical claims.

### 1. The exterior collar has an impossible ambient openness condition

Location: `Suggested.lean`, `Internal.GeometricTopologySupplierMirror.SmoothEmbeddedClosedDisk`, especially `exteriorCollar`, its source/target conditions, and `exteriorNeighbourhood`.

The exterior collar has source `[0,1) × boundary`, contains the disk boundary at parameter zero, and sends positive parameters outside the closed disk. Its target is required to be open in the ambient manifold `M`.

For a genuine codimension-zero disk in a closed manifold, this is impossible. Every ambient open neighbourhood of a boundary point meets the disk interior: interior points of the disk approach that boundary point, and the embedding is continuous. The image of the proposed one-sided exterior collar contains no disk-interior points. It therefore cannot be such an ambient open neighbourhood.

An exterior collar is a collar in the **closed complement of the disk interior**, a manifold with boundary. Alternatively, construct a two-sided collar in `M` and restrict it to the exterior half, without requiring that half to be open in `M`.

**Required test:** construct the standard embedded disk and its exterior collar in a standard sphere using the actual proposed interface, before defining almost-framed cycles on top of it.

### 2. The filling bordism witness omits its vertical boundary

Location: `Suggested.lean`, `StableFillingBordismWitness`.

The README correctly describes a cobordism with corners whose horizontal faces are the two fillings and whose vertical face is an h-cobordism of their boundaries. The Lean record instead requires the incoming and outgoing filling images to be disjoint and to cover the entire closed boundary carrier of the trace. There is no vertical face and no corner compatibility.

For the reflexive bordism `W × I`, the boundary has three pieces:

\[
 W\times\{0\},\quad W\times\{1\},\quad \partial W\times I.
\]

The third cannot be omitted when `∂W` is nonempty. More intrinsically, the two compact disjoint images in a Hausdorff boundary carrier would both be open and closed if they covered it. That would make a filling with nonempty boundary into a boundaryless open component. Taking `W=D^n` already contradicts the intended geometry.

Replace this record by genuine collared corner/face data. The horizontal and vertical faces have prescribed intersections, and the vertical face must carry the stated oriented h-cobordism data. Smoothing corners must retain the identifications and framing restrictions.

**Required test:** construct the cylinder witness and prove reflexivity from that witness. Do not just assert abstract reflexivity of the relation later.

### 3. The h-cobordism relation forgets the endpoint orientations

Location: `Suggested.lean`, `BoundaryIdentification`, `GeometricHCobordismWitness`, and the relation used to define the homotopy-sphere quotient.

The input cycles carry orientations, and the trace carries an orientation, but the boundary-identification record only specifies embeddings, disjointness, and coverage. It does not compare the orientations of the endpoints with the induced boundary orientation of the trace.

Consequently the witness type is insensitive to reversing an endpoint's orientation while retaining the same underlying manifold data. This is not the oriented h-cobordism relation required for the Kervaire–Milnor group, and the intended oriented connected-sum operation cannot simply be transported to this quotient.

Require the incoming identification to have the conventional reversed boundary orientation and the outgoing identification to preserve the specified orientation, using the shared orientation API and actual derivatives/induced boundary orientation. Pin the sign convention once and use it throughout connected sum and the boundary maps.

**Required test:** the cylinder should realize `∂(M × I) = (-M) ⨿ M` with the displayed identifications. Endpoint orientation compatibility must be an actual field or proposition, not a comment.

### 4. The almost-framed and filling relations forget endpoint framings

Location: `Suggested.lean`, `AlmostFramedBordismWitness`, `StableFillingBordismWitness`, and `StableFramedFillingCycle.StabilizesRelBoundary`.

The almost-framed witness uses the underlying oriented endpoint cycles but does not require its complement framing to restrict to the displayed endpoint complement framings. Its defect object is also not tied to the endpoint defect disks in the way promised by the README's properly embedded defect track.

The stable-filling witness similarly lacks the restriction equations comparing the trace framing with the two given filling framings. A geometric relation which ignores those data is not the framed relation required by the exact sequence.

The relative-stabilization record also does not retain the global homotopy of trivializations relative to the collar required by README 6B. Equality only on the collar is not a substitute for that global relative homotopy.

Add the endpoint pullback identifications, a common stabilization convention, normal-line/order conventions, the framing equations, and the actual relative homotopy. A defect track must have endpoint/collar identifications with the selected defect disks; an unrelated embedded ball in the trace is not that object.

These conditions must be fixed before forming quotient groups or claiming comparison with historical `A_n` and `P_n`. A later theorem about an already misdefined quotient cannot reconstruct information discarded by its equivalence relation.

### 5. The relative destabilization proof has an impossible handle step

Location: `README.md`, Stage 6B, step 3.

The proof asks for a handle decomposition relative to the full boundary collar with no top-dimensional handles. For a compact connected oriented `n`-manifold with boundary,

\[
 H_n(W,\partial W;\mathbb Z)\cong\mathbb Z.
\]

A relative handle decomposition with no `n`-handles would have zero top relative cellular group and hence zero top relative homology. The disk `(D^n,S^(n-1))` already disproves the proposed step.

An **absolute** spine of dimension at most `n-1` is a different statement. It can be useful for proving existence of an honest parallelization, but it does not automatically prove extension or a prescribed homotopy **relative to the whole boundary collar**. The roadmap must distinguish these assertions and state the actual relative obstruction/cancellation theorem and its hypotheses.

This review establishes that the stated proof step fails. It does not replace it by an unverified general theorem saying that every prescribed relative framing destabilizes. That stronger compatibility claim needs its own careful check against the chosen normal-line and stabilization conventions.

### 6. Correct the compact-generation adjunction and reuse the existing inclusion

Location: `README.md`, Stage 3A; compare Mathlib at the pinned commit, `Mathlib/Topology/Category/CompactlyGenerated.lean`.

The existing category and its fully faithful inclusion should be reused. In particular the file already has `compactlyGeneratedToTop`, `fullyFaithfulCompactlyGeneratedToTop`, and the Full/Faithful instances.

The usual compact-generation functor is a **coreflector**, not a reflector. If `i : CG → Top` is the inclusion and `kX` is compact generation, then

\[
 \operatorname{Hom}_{Top}(iY,X)
 \cong\operatorname{Hom}_{CG}(Y,kX),
 \qquad i\dashv k.
\]

The continuous comparison is `kX → X`. This follows directly by testing continuity out of `Y` on compact Hausdorff probes. Do not rely on the pinned Mathlib file's unproved TODO calling the subcategory reflective; that comment does not establish the needed adjunction. State the direction and the resulting limit/colimit constructions explicitly.

### 7. Integration tests are still necessary

`Suggested.lean` imports Mathlib but does not yet exercise the promised concrete AlgebraicTopology supplier declarations. Add imports or deliberate temporary supplier checks for the corrected Hurewicz, Whitehead, finite-CW, and duality interfaces. Name exactly which geometric-topology and orientation fields each bordism construction consumes.

The essential early acceptance cases are standard disks, punctured spheres, genuine exterior collars, product bordisms with vertical faces, and correctly oriented/framed endpoint restrictions. These are more informative tests of the current architecture than another final `Theta₆ = 0` signature.

### Repairs that should remain accepted

The connected James theorem is no longer justified by wrongly assuming simple connectivity. The Thom-space target has been corrected to `Σ^k M_+` followed by the augmentation to `S^k`. The Bott section spells out the path-space/broken-geodesic and Morse–Bott infrastructure. The small-carrier encoding, real geometric cycles, even/odd surgery data, and orientation-preserving final conclusion are significant improvements.

The requested changes concern specific defects that survive those repairs, not the size or ambition of the overall project.

### Disposition

Repair the geometric records, relations, and relative-handle argument before merge, and validate them on the disk/cylinder examples. Current CI is green but cannot turn these ill-specified or impossible geometric contracts into the intended objects.

CBirkbeck (commented):
:robot: Two wording items for the same round, on the Mathlib references — kim-em's "dated prose" point (#142, #185): the roadmap keeps the timeless conditional, PR status goes in the PR description. The first suggestion also folds in the coreflector correction from item 6 of the review above, so the inventory sentence states the adjunction in the right direction.

— Claude, on Chris's behalf

TauCetiRoadmap/HomotopySpheres/README.md
```diff
@@ -0,0 +1,817 @@
...
+  mapping cones, homotopy lifting, and fundamental groupoids.
+- Singular homology, `TopPair`, Mathlib's CW-complex structures, metric spheres, and the
+  algebraic-topology roadmap's relative Hurewicz, Whitehead, duality, and finite-CW-type results.
+- `ModelWithCorners`, `ChartedSpace`, `IsManifold`, tangent bundles, smooth maps,
+  `Diffeomorph`, and `exists_embedding_euclidean_of_compact`.
+- `Matrix.orthogonalGroup`, `Matrix.specialOrthogonalGroup`, Lie groups, topological groups,
+  vector bundles, pullbacks, direct sums, quotient spaces, `CompactlyGenerated.{u,w}`,
+  categorical filtered colimits, and `AddCommGrpCat`.  Mathlib's compactly generated category
+  does not yet supply the reflector or Cartesian closed API needed here; those are targets, not
+  inferred instances.
```
CBirkbeck: "Does not yet supply" is a statement about today's Mathlib, and "reflector" is the wrong direction: compact generation `k` is right adjoint to the inclusion, `i ⊣ k`, with the comparison `kX → X` (item 6 of the review).

```suggestion
  categorical filtered colimits, and `AddCommGrpCat`.  Mathlib's compactly generated category
  supplies the carrier and its fully faithful inclusion; the compact-generation coreflector (the
  right adjoint of that inclusion), its adjunction, and the Cartesian closed API are targets here,
  not inferred instances.
```

TauCetiRoadmap/HomotopySpheres/README.md
```diff
@@ -0,0 +1,817 @@
...
+- Geometric topology's boundary, collar, gluing, tubular-neighbourhood, connected-sum, handle,
+  surgery, cobordism, and triangulation outputs, plus Heegaard Floer's orientation, degree,
+  Morse, Sard, and transversality outputs.
+
+These open Mathlib pull requests determine target shapes.  Consume declarations already present,
+build the missing interface in Tau Ceti now in the same shape, and replace it by imports when it
+lands.
```
CBirkbeck: Drop "open" and "when it lands": the rule is conditional on the pin, not on a future merge.

```suggestion
These Mathlib pull requests determine target shapes.  Consume every declaration present at the
dependency pin, build the missing interface in Tau Ceti now in the same shape, and replace it by
imports whenever a later pin supplies it.
```

Paul-Lez: :robot: Addressed in `bd74132`. The Mathlib-PR paragraph no longer says “open” or “when it lands”; it now conditions imports and local replacements entirely on the dependency pin.

TauCetiRoadmap/HomotopySpheres/README.md
Paul-Lez: :robot: Addressed in `bd74132`, with the concrete API correction in `a13b35e`/`e0400e5`. The inventory reuses Mathlib’s fully faithful inclusion, names compact generation as its right adjoint/coreflector in `i ⊣ k`, and makes the counit `i(kX) ⟶ X` explicit. Stage 3A also distinguishes colimits preserved by the inclusion from limits obtained by applying `k`.

Paul-Lez:
:robot: Addressed through head `3416730`. The exterior collar now lives in the closed complement of the disk interior, with a standard disk-in-sphere target. Boundary identifications carry incoming-reversed/outgoing-preserved orientation equations from induced boundary orientations. Almost-framed and filling witnesses now track endpoint defects/collars and compare endpoint framings after a common stabilization. `StabilizesRelBoundary` contains a genuine globally continuous homotopy fixed on the collar. Filling bordisms have three named faces, exact corner intersections, a vertical oriented h-cobordism and induced product framing, plus an explicit cylinder reflexivity witness. Stage 6B now separates the absolute spine argument from the prescribed-collar obstruction and no longer claims a relative handle decomposition without top handles. Compact generation is the right adjoint/coreflector in `i ⊣ k`, reusing Mathlib’s existing fully faithful inclusion and stating the limit/colimit directions. The corrected #283 supplier is stacked here, imported literally, and its four named declarations are `#check`ed. `lake build TauCetiRoadmap.HomotopySpheres.Suggested` succeeds (8703 jobs).

CBirkbeck (changes requested):
## Review by ChatGPT Pro

Reviewed `34167306e5a1726e2574b039dc1a27f913985f0f`, the complete HomotopySpheres README and representative declarations, the corrected AlgebraicTopology supplier, and relevant Mathlib definitions at the actual pin `05ae0103f49b1ad1248f6039bbbad43d8aeb52a9`. This is a mathematical and source-level Lean-interface review, not a claim to have run a local build.

Several previous findings are genuinely resolved: `i ⊣ k` is now correct; closed h-cobordism endpoints have orientation equations; filling bordisms now have a vertical face and corner intersections; endpoint framing comparisons and a global relative homotopy are present; the impossible relative handle-decomposition claim has been removed. The connected James proof and corrected Thom-space factorization should also remain accepted.

However, the revised contracts still have foundational problems, including a new misuse of Mathlib's designated trivialization atlas.

### 1. `MemTrivializationAtlas` is not the condition that a global framing is smooth

**Locations:** `TangentFraming.memAtlas`, `StableTangentFraming.memAtlas`, and their consumers.

At this Mathlib pin, the canonical tangent bundle is constructed from `tangentBundleCore I M`, indexed by the manifold atlas. `FiberBundleCore.fiberBundle.trivializationAtlas'` is literally `Set.range Z.localTriv`, and `tangentBundleCore_localTriv_baseSet` identifies each such base set with the source of its manifold chart. `MemTrivializationAtlas` means membership in this **designated** atlas, not compatibility with it or membership in a maximal smooth atlas.

Consequently, combining

```
global : trivialization.baseSet = Set.univ
memAtlas : MemTrivializationAtlas trivialization
```

forces a single manifold chart to have source all of `M`. For a nonempty compact boundaryless positive-dimensional manifold, that is impossible: it would identify `M` with a nonempty compact open subset of Euclidean space. In particular this excludes honest tangent framings of `S¹` and `S³`, which the programme must allow.

Stabilization does not repair it. The canonical fibrewise-product atlas consists of products of designated trivializations; its base sets are intersections. The trivial summand's chart is global, so a global chart in that product atlas still forces a global tangent chart. Thus the current stable-framing carrier excludes the closed framed manifolds needed for Pontryagin–Thom.

**Repair:** retain the canonical bundle/topology but require a global linear trivialization with actual smooth compatibility (or a smooth vector-bundle equivalence to the trivial bundle), rather than membership in the original designated atlas. Do not solve this by replacing the canonical bundle instances with arbitrary ones. Test the Lie-group framing of `S³`, the stable framing of a sphere, and stabilization of those actual framings.

Pinned sources: [tangent bundle construction](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Geometry/Manifold/VectorBundle/Tangent.lean), [core atlas](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/FiberBundle/Basic.lean), and [product atlas](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/FiberBundle/Constructions.lean).

### 2. The revised disk exterior still has the wrong manifold model; a defect tube needs a different collar contract

**Locations:** `SmoothEmbeddedClosedDisk.closedExteriorCharted`, `closedExteriorManifold`, `exteriorCollar`, `standardClosedDiskInSphere`, and `AlmostFramedBordismWitness.defectTrack`.

Moving the exterior collar's target into the closed exterior correctly removes the old ambient-openness error. But that closed exterior is still required to use the **ambient** model `I` and chart model `H`.

For the standard disk in a boundaryless sphere, `I = SmoothModel n`. Its closed exterior is a closed disk with its inherited topology, so it has boundary; it cannot be a boundaryless Euclidean-model manifold. Local top homology at a boundary point is zero, whereas an interior Euclidean-manifold point has local top homology `ℤ`. Therefore the new `standardClosedDiskInSphere` target still cannot be constructed with these fields.

Give the closed exterior its own manifold-with-boundary model and its actual boundary identification. Its **open interior** can use the ambient boundaryless model; the closed exterior cannot.

There is also a distinct endpoint problem with using this same disk structure as `defectTrack`. In the reflexive defect bordism, the tube is `Dⁿ × I ⊂ M × I`. The interiors of its endcaps lie in the ambient trace boundary and have neighbourhoods in the trace entirely contained in the tube. No positive exterior-collar points in the trace approach such an endcap point. The required exterior collar along the **whole intrinsic boundary of the tube** is therefore impossible, even after correcting its manifold model.

A proper defect-arc tubular neighbourhood needs named endcaps and a collar along its **lateral boundary**, together with the endpoint disk/collar identifications. It is not an embedded interior disk with an exterior collar along every boundary point. Test the actual product defect tube before asserting `almostFramedBordant_equivalence`.

### 3. `restrictStableFramingAlongCodimOne` is false at its displayed generality

The function takes an arbitrary smooth embedding into a stably framed manifold and returns a framing of `TM ⊕ ε^(r+1)`. It has no dimension equation, codimension-one hypothesis, trivialization of the normal line, coorientation, or boundary/collar datum.

Even adding codimension one alone would not suffice. The interior of an embedded Möbius band is a smooth nonorientable hypersurface of framed `ℝ³`. A trivialization of its tangent bundle plus one trivial line would orient it, which is impossible. The actual normal line is nontrivial.

**Repair:** consume the chosen normal-line trivialization and the bundle decomposition `f*TN ≃ TM ⊕ ε`, or specialize the construction to a genuine boundary/face inclusion with its collar and outward normal. For general codimension, use the actual normal bundle and its chosen framing. Pin the incoming normal reversal explicitly rather than placing it only in the docstring. The current function underlies both defect and filling framing equations, so those equations cannot validate the geometry until this constructor is corrected.

### 4. The general bordism carrier must allow empty/disconnected objects, and the sphere statements need their dimension range

**Locations:** `CollaredOrientedManifold.connected`, `boundary_nonempty`, `FramedBordismWitness`, `framedBordant_equivalence`, and the unrestricted h-cobordism cylinder/group declarations.

The shared trace is required to be connected and have nonempty boundary, although framed bordism is supposed to use disjoint union and include the empty framed manifold. Once the intended empty cycle is admitted by the corrected framing API, `BoundaryIdentification empty empty W` makes both ranges empty; `rangesCover` forces `W.B` empty, contradicting `boundary_nonempty`. The empty-to-empty reflexivity witness is therefore excluded by the trace interface itself.

Separately, `geometricHCobordismCylinderWitness` and `geometricHCobordant_equivalence` quantify over **all** `n`. At `n = 0`, the cycle is `S⁰`; a connected trace cannot be homotopy equivalent to this disconnected space. The README's `n ≥ 5` scope has also disappeared from `homotopySphereClassAddCommGroup`.

**Repair:** use a general collared bordism carrier permitting empty and disconnected traces/boundaries. Put connectedness and nonempty-boundary assumptions on the filling or high-dimensional h-cobordism objects where appropriate, rather than on every bordism. State the geometric sphere/group theorems in their intended range. Include empty and disjoint-union cylinders as acceptance cases.

### 5. The new filling faces still do not compare their orientations with the trace

**Locations:** `StableFillingBoundaryFaces` and `StableFillingBordismWitness`.

The vertical h-cobordism now compares its own endpoint orientations, which is good. But the horizontal and vertical face embeddings into `trace.B` have no equations comparing their induced orientations with `trace.boundaryOrientation`. The trace framing likewise is not required to induce `trace.orientation`.

One can reverse the trace orientation and its boundary orientation together, preserving `ambient_induces_boundary`, without changing the face maps or framing equations. Thus the promised incoming-minus/outgoing-plus comparison with the trace is not encoded and cannot follow from these fields.

Add the actual derivative/normal-line pullback equations for all three faces and agreement of the trace framing with its orientation. This is the remaining filling-face issue, not a reopening of the now-fixed closed `BoundaryIdentification` orientation fields.

### 6. The odd Wall relation does not kill the elementary formations required by the standard definition

**Locations:** `OddWallFormation.ElementaryModification`, `hyperbolicWallFormation`, and the odd constructors of `WallSurgeryDatum.StablyEquivalent`.

The displayed elementary modification is a transvection along a vector in the first Lagrangian. It fixes that Lagrangian pointwise and is bijective, hence preserves the rank of the intersection of the two Lagrangians. Formation isometries preserve it as well. Stabilization by the standard hyperbolic formation with complementary Lagrangians also preserves this intersection rank.

Take the hyperbolic plane `H(ℤ)` with Lagrangian `L = ℤe`, but use **the same** `L` for both displayed Lagrangians. It has intersection rank one. It therefore cannot become the zero formation under the displayed moves and complementary-hyperbolic stabilization. Yet it is an elementary zero formation in the standard odd `L`-group, because the two Lagrangians share the common complement `ℤf`.

Changing the single fixed stabilizer to the diagonal formation is not a solution: that instead preserves `rank(L₁) − rank(L₁ ∩ L₂)` and fails to kill the complementary formation. The definition needs the actual stable/cobordism relations, not just a single stabilizer plus these intersection-preserving transvections.

**Repair:** allow the standard elementary formations with complementary Lagrangians **and** those with a common Lagrangian complement, or implement an equivalent presentation and prove the comparison. Require both `(H(ℤ); L, L*)` and `(H(ℤ); L, L)` to represent zero as explicit small tests before claiming the odd group vanishes.

Source: [Ranicki, *Foundations of algebraic surgery*, §2, printed pp. 6–7](https://arxiv.org/pdf/math/0111315), which specifies both kinds of stabilization.

### 7. The James tensor-algebra homology calculation needs coefficient hypotheses

**Location:** README Stage 3B.3, the unqualified tensor-algebra isomorphism used before the integral universal-cover argument.

The formula `H_*(JX; R) ≅ T(Ĥ_*(X; R))` is not an unrestricted integral homology statement. The usual Bott–Samelson formulation assumes the needed freeness of `H_*(X; R)` (or works over a field). Connectedness does not imply this.

For the Moore space `X = M(ℤ/2, 2)`, the tensor algebra on reduced integral homology has no degree-five part. However, the smash square has `Ĥ₅(X ∧ X; ℤ) ≅ Tor₁(ℤ/2, ℤ/2) ≅ ℤ/2`; the James stable splitting gives a corresponding nonzero degree-five summand in `H_*(JX; ℤ)`.

**Repair:** qualify the tensor-algebra theorem by its coefficient/freeness hypotheses. For the general integral James proof, retain the chain-level or smash-power calculation and its Tor terms instead. The full connected James equivalence remains the desired theorem; this only corrects a false intermediate homology formula. Add the Moore-space test alongside `X = S¹`.

A precise formulation is [Cutler, *The Bott–Samelson Theorem*, Theorem 6.2](https://www.math.uni-bielefeld.de/~tcutler/pdf/Week%2011%20-%20The%20Bott-Samelson%20Theorem.pdf).

The requested changes concern actual counterexamples and the exact meanings of the chosen Mathlib objects. They should be settled before the quotient groups and historical-comparison maps are built on these carriers.

CBirkbeck (changes requested):
## Review by ChatGPT Pro

**Head reviewed:** `34167306e5a1726e2574b039dc1a27f913985f0f`.

**Changes are still required.** This head is unchanged from the [previous detailed review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/284#pullrequestreview-5139196960). I reread the README, the representative geometric and algebraic interfaces, and the author's responses. The earlier seven findings remain applicable; approving the AlgebraicTopology supplier does not resolve them. I also found the following additional concrete obstruction in the atlas construction.

## New finding: an allowed empty-chart code makes `realizedChart` impossible

**Locations:** `SmoothAtlasCode`, `SmoothAtlasCode.Realization`, and `SmoothAtlasCode.realizedChart` in `Suggested.lean`.

`SmoothAtlasCode` permits empty chart domains. Take a code with exactly one chart, its domain empty, all overlaps empty, and the transition equal to the identity partial homeomorphism restricted to the empty set. These transitions satisfy the smooth-groupoid condition. All source/target, identity, inverse and cocycle conditions hold; the identification relation on the empty disjoint union satisfies the equivalence and closedness conditions vacuously.

For this perfectly admissible code `C`, both

```
C.PrePoint = Σ i, C.chartDomain i
C.Realization = Quotient C.gluingSetoid
```

are empty. Nevertheless, the file asks for

```
C.realizedChart 0 : OpenPartialHomeomorph C.Realization (ModelSpace n).
```

At the actual Mathlib pin, an `OpenPartialHomeomorph X Y` contains total forward and inverse functions; only their inverse laws and continuity are restricted to the source and target. Its inverse here would therefore be a total function

```
ModelSpace n → C.Realization.
```

The model space contains zero. Evaluating that inverse at zero would give an element of the empty realization. Thus the current `realizedChart` constructor cannot exist for every allowed code. This is not a request to fill in a difficult proof: the input contract admits a counterexample.

**Repair:** require each *listed* chart domain to be nonempty, while still allowing `chartCount = 0` to represent the empty manifold. Alternatively, prune empty charts and index `realizedChart` only by the nonempty ones, with the comparison between the realizations made explicit. Do not repair this by globally requiring every realized manifold to be nonempty: the empty framed cycle is needed for the bordism group.

**Acceptance tests:** the zero-chart code has empty realization and a valid empty-manifold charted structure; a one-chart code on all of `ℝ^n` realizes `ℝ^n`; and a one-empty-chart input is either excluded or removed by the pruning construction. These small tests should precede the small-skeleton and bordism quotient declarations.

Pinned source for the total-function convention: [Mathlib/Topology/OpenPartialHomeomorph/Defs.lean](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/OpenPartialHomeomorph/Defs.lean), which explains the underlying `PartialEquiv` functions explicitly.

## Outstanding repairs, in implementation order

I am not copying the earlier counterexamples in full; the linked review contains them. The unchanged declarations still require these corrections:

1. **Framings:** replace membership in the designated tangent trivialization atlas by actual smooth compatibility of a global linear trivialization. The combination of `global` and `memAtlas` still excludes the intended closed framed examples. Preserve the canonical bundle and topology rather than changing them to make the condition true.
2. **Embedded disks and framing restriction:** give the closed exterior its own boundary model; distinguish a proper defect tube's lateral collar from its endcaps; and supply the chosen normal-line framing/coorientation to `restrictStableFramingAlongCodimOne`. The present arbitrary-embedding signature does not provide it.
3. **Bordism witnesses:** permit empty/disconnected objects in the general framed-bordism carrier, retain the appropriate dimensional restrictions for sphere-group results, and compare all filling-face orientations and the trace framing with the actual trace orientation. A connected filling carrier with nonempty boundary cannot serve as every bordism trace.
4. **Algebraic targets:** repair the odd Wall equivalence so that both standard types of elementary zero formation are killed, and qualify the James tensor-algebra homology formula by its coefficient/freeness hypotheses. The previous formation and Moore-space tests still apply.

Coordinate the boundary/orientation changes with the named suppliers, rather than making another public copy of their interfaces. The internal-mirror deletion gates remain useful, but a future import cannot repair a false theorem about the present mirror types.

The most effective next revision is to make the standard disk/exterior, a genuine closed stable framing, the empty/product bordisms, and the defect product tube fit the corrected definitions first. Only then should the equivalence relations, group structures and Kervaire–Milnor maps be rebuilt on those definitions. The already repaired connected James scope, Thom-space factorization, coreflector direction and supplier ownership should remain accepted.

### Validation

The head-associated CI run `34027448769` is successful; these declarations still have roadmap `sorry` proofs, so that establishes elaboration, not the truth of the constructor above. I did not run a local Lean build. The head was rechecked before submission and remains unchanged. No approval, merge, draft-state change or branch change is being issued.

CBirkbeck (changes requested):
:robot: **AI-assisted mathematical and Lean-interface review, posted at Chris's request.**

**Head reviewed and rechecked before submission:** `34167306e5a1726e2574b039dc1a27f913985f0f`.

**Verdict: changes required.** This is still the head covered by the [September 8 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/284#pullrequestreview-5139196960) and [September 10 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/284#pullrequestreview-5171898005). I reviewed the complete HomotopySpheres README and `Suggested.lean`, the supplier import, the relevant pinned Mathlib definitions, and the earlier discussion. Item 1 below is an additional independent counterexample. Items 2–9 summarize the previously reported issues that remain in the unchanged declarations, with the required repairs.

The objections are to specific mathematical contracts, not to the ambition of the roadmap or to the legitimate use of `sorry` in target signatures.

## 1. New: the atlas-realization laws fail even with one nonempty chart

**Location:** `Suggested.lean`, `SmoothAtlasCode.realization_atlas` and `SmoothAtlasCode.realized_transition`.

The file promises literal equality

```lean
(C.realizedChart i).symm ≫ₕ C.realizedChart j = C.transition i j
```

between partial homeomorphisms. This is stronger than agreement on the overlap: Mathlib's partial homeomorphisms contain total forward and inverse functions, whose values outside the source and target also participate in literal equality.

Here is a counterexample independent of the empty-chart problem.

Work in dimension one, identifying the model with `ℝ`. Take exactly one chart, with domain `U = (-1,1)`, and its self-transition equal to `OpenPartialHomeomorph.ofSet U hU`. That transition's total forward function is the identity on **all** of `ℝ`, although its source and target are `U`. The stated code conditions hold: the overlap is `U`, the identification relation is equality on the chart domain, and the identity, inverse, cocycle, smooth-groupoid and separation conditions are satisfied.

Write `e = C.realizedChart 0`. The two promised realization theorems now contradict each other:

1. `realization_atlas` makes the atlas the singleton `{e}`. Since the atlas covers the realization, `e.source = Set.univ`.
2. The source formula for `e.symm ≫ₕ e`, together with `realized_transition 0 0`, then gives `e.target = U`.
3. Equality of the total forward functions in `realized_transition 0 0` gives `e (e.symm t) = t` for **every** `t : ℝ`.
4. At `t = 2`, `e.symm 2` is an element of the realization, hence belongs to `e.source`. Its image must belong to `e.target = (-1,1)`, but step 3 says that image is `2`.

These constructors are asserted for every `SmoothAtlasCode`; an informal restriction of intended applications to compact manifolds does not exclude this input.

**Repair:** replace literal transition equality by the appropriate `OpenPartialHomeomorph.EqOnSource` relation, together with explicit overlap/source identifications:

```lean
(C.realizedChart i).symm ≫ₕ C.realizedChart j ≈ C.transition i j
```

The total extensions outside the overlaps are not part of the geometric gluing data. Mathlib itself uses `≈`, not `=`, in `self_trans_symm` and `symm_trans_self`.

**Pinned sources:** [partial-homeomorphism definitions](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/OpenPartialHomeomorph/Defs.lean) and [composition, including `coe_trans`, `trans_source`, and `symm_trans_self`](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/OpenPartialHomeomorph/Composition.lean).

**Acceptance test:** realize a single proper nonempty open chart such as `(-1,1)`, not only a global Euclidean chart. Requiring chart domains to be nonempty will not fix this separate error.

## 2. The earlier empty-chart obstruction remains

**Location:** `SmoothAtlasCode`, `SmoothAtlasCode.Realization`, and `SmoothAtlasCode.realizedChart`.

One listed chart with empty domain is still permitted. The realization is then empty, but `realizedChart` must return a partial homeomorphism from it to the inhabited Euclidean model. Its total inverse would map zero into the empty realization.

**Repair:** require every *listed* chart domain to be nonempty while permitting zero charts, or prune empty charts before constructing realized charts. Do not globally exclude empty manifolds: the framed-bordism development needs them.

**Acceptance tests:** zero charts, a rejected or pruned empty chart, one global Euclidean chart, and the proper nonempty chart from item 1. The [September 10 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/284#pullrequestreview-5171898005) contains the full empty-chart counterexample.

## 3. The framing definitions still exclude the intended closed framed manifolds

**Location:** `TangentFraming.memAtlas` and `StableTangentFraming.memAtlas`.

The structures combine a global trivialization with membership in the bundle's **designated** trivialization atlas. At the actual Mathlib pin this is not the condition that an arbitrary global trivialization is smoothly compatible with the bundle structure.

The canonical tangent atlas consists of trivializations induced by manifold charts; their base sets are those chart sources. A global member therefore forces one manifold chart to cover the entire manifold. For a nonempty compact boundaryless positive-dimensional manifold, that would identify it with a nonempty compact open subset of Euclidean space, which is impossible. The product-bundle atlas has base sets given by intersections of designated base sets, so stabilization does not remove the problem.

**Repair:** use an actual smooth global linear bundle trivialization, or a smooth vector-bundle equivalence with the trivial bundle, compatible with the **canonical** tangent-bundle structure. Do not change the canonical bundle instances to force atlas membership.

**Acceptance tests:** a Lie-group framing of `S³`, its stabilization, and a stable framing of a standard sphere.

**Pinned sources:** [tangent-bundle construction and base sets](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Geometry/Manifold/VectorBundle/Tangent.lean), [`MemTrivializationAtlas` and the core atlas `Set.range Z.localTriv`](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/FiberBundle/Basic.lean), and [product-bundle construction](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/FiberBundle/Constructions.lean).

## 4. The closed disk exterior has the wrong manifold model; a proper defect tube needs a different collar

**Location:** `SmoothEmbeddedClosedDisk.closedExteriorCharted`, `closedExteriorManifold`, `exteriorCollar`, `standardClosedDiskInSphere`, and `AlmostFramedBordismWitness.defectTrack`.

Moving the exterior collar into the **closed complement of the disk interior** correctly removed the old ambient-openness condition. However, that closed exterior still uses the ambient manifold's chart model and `ModelWithCorners`.

For the standard disk in a boundaryless sphere, the closed exterior is a disk **with boundary**, not a boundaryless Euclidean-model manifold. Thus `standardClosedDiskInSphere` still cannot instantiate the intended contract.

There is a separate defect-track issue. The reflexive tube is `Dⁿ × I ⊂ M × I`. An interior point of an endcap has a neighbourhood in the ambient trace contained in the tube. Exterior points therefore cannot approach it through the trace as required by an exterior collar along the tube's **whole intrinsic boundary**.

**Repair:** give the closed exterior its own boundary model and actual boundary identification. Separately model a proper defect tube with named endcaps and a collar along its **lateral boundary**, together with endpoint disk/collar identifications. An interior embedded disk with an exterior collar along every boundary point is not the appropriate substitute.

Construct the genuine product defect tube before asserting almost-framed bordism reflexivity.

## 5. Framing restriction lacks the necessary normal-bundle data

**Location:** `restrictStableFramingAlongCodimOne`.

Despite its name, the function accepts an arbitrary smooth embedding into a stably framed manifold and returns a framing of the source after adding one trivial line. It has no dimension equation, codimension-one hypothesis, normal-line trivialization, coorientation, or collar datum.

The bundle decomposition supplied by an embedding is `f*TN ≃ TM ⊕ ν_f`, not automatically `TM ⊕ ε¹`. Even codimension one alone is insufficient: the interior of an embedded Möbius band in framed `ℝ³` is nonorientable, whereas a framing of its tangent bundle plus one trivial line would orient it.

**Repair:** consume the chosen normal-bundle trivialization and actual bundle decomposition, or specialize to a genuine boundary/face inclusion with its collar and outward normal. Encode incoming normal reversal rather than leaving it in the docstring.

This constructor underlies both almost-framed and filling endpoint equations. Those equations cannot validate the intended geometry while the restriction constructor has this generality.

## 6. A connected filling is still used as the general bordism trace; sphere results need their dimension range

**Location:** `CollaredOrientedManifold.connected`, `boundary_nonempty`, `FramedBordismWitness`, and the unrestricted h-cobordism cylinder/group declarations.

The common trace requires connectedness and nonempty boundary. These may be appropriate hypotheses for particular fillings, but not for general framed bordism.

Even after the framing carrier is corrected to admit the intended empty cycle, an empty-to-empty bordism cannot fit: the boundary coverage equation forces the boundary carrier empty, contradicting `boundary_nonempty`. Disconnected product bordisms are excluded by the connectedness field.

There is also a range error in `geometricHCobordismCylinderWitness` and `geometricHCobordant_equivalence`, which quantify over all `n`. At `n = 0`, an endpoint of homotopy type `S⁰` cannot be homotopy equivalent to the required connected trace. The representative group declaration also omits the README's `n ≥ 5` scope.

**Repair:** separate the general collared-bordism carrier from connected fillings and high-dimensional h-cobordisms. Put connectedness, nonempty-boundary and dimensional hypotheses on the particular objects and results that need them.

**Acceptance tests:** empty and disconnected cylinders, as well as the connected high-dimensional sphere cylinder.

## 7. Filling-face orientations are still not tied to the trace orientation

**Location:** `StableFillingBoundaryFaces` and `StableFillingBordismWitness`.

The horizontal and vertical faces and their intersections are now present, and the vertical h-cobordism has oriented endpoint identifications. Those repairs should remain accepted.

But the face embeddings have no equations comparing their induced orientations with `trace.boundaryOrientation`. The trace framing is also not required to induce `trace.orientation`. Reversing the trace orientation and its induced boundary orientation together leaves the face maps, corner equations and displayed framing comparisons unchanged.

**Repair:** add the normal/derivative pullback orientation comparisons for all three faces and agreement of the trace framing with its orientation. This is distinct from the already repaired closed-endpoint `BoundaryIdentification` orientation equations.

## 8. The odd Wall relation still misses elementary-zero relations

**Location:** `OddWallFormation.ElementaryModification`, `hyperbolicWallFormation`, and `WallSurgeryDatum.StablyEquivalent`.

The displayed elementary modification is a transvection along a vector in the first Lagrangian. It fixes that Lagrangian pointwise and is bijective, so it preserves the rank of the intersection of the two Lagrangians. Formation isometries preserve this rank as well. With the standard complementary hyperbolic stabilizer, stabilization preserves it too.

Consequently `(H(ℤ); L, L)` cannot become zero under those moves: its intersection rank is one. Yet it is an elementary zero formation in the standard odd `L`-group, since the two Lagrangians have a common Lagrangian complement.

**Repair:** implement both standard elementary-zero relations, or an equivalent presentation with a proved comparison. Simply changing the choice of one fixed stabilizer is not enough. Require both `(H(ℤ); L, L*)` and `(H(ℤ); L, L)` to represent zero as explicit tests before using odd-dimensional vanishing.

**Source:** [Ranicki, *Foundations of algebraic surgery*, §2, printed pp. 6–7](https://arxiv.org/pdf/math/0111315), which distinguishes complementary Lagrangians from Lagrangians with a common Lagrangian complement in the allowed stabilization relations.

## 9. The James tensor-algebra homology calculation still needs coefficient hypotheses

**Location:** README Stage 3B.3.

The unqualified formula `H_*(JX; R) ≅ T(Ĥ_*(X; R))` is not an unrestricted integral homology theorem. The usual Bott–Samelson formulation requires suitable freeness hypotheses, or field coefficients. Connectedness alone does not imply them.

For `X = M(ℤ/2, 2)`, the tensor algebra on reduced integral homology has no degree-five part. But the smash square has

```
Ĥ₅(X ∧ X; ℤ) ≅ Tor₁^ℤ(ℤ/2, ℤ/2) ≅ ℤ/2,
```

and the James stable splitting gives a corresponding summand in the homology of `JX`.

**Repair:** qualify the tensor-algebra theorem. For the general integral James proof retain the chain-level or smash-power calculation, including its Tor terms. This does not require weakening the desired connected James equivalence. Include the Moore-space test alongside `X = S¹`.

A precise coefficient-qualified statement is [Cutler, *The Bott–Samelson Theorem*, Theorem 6.2](https://www.math.uni-bielefeld.de/~tcutler/pdf/Week%2011%20-%20The%20Bott-Samelson%20Theorem.pdf); the full counterexample was also recorded in the linked September 8 review.

## Repairs that remain accepted

The compact-generation adjunction now has the correct direction; the Thom-space factorization is corrected; and the James proof no longer wrongly treats every connected input as simply connected. The expanded Bott proof plan, the separation of absolute parallelizability from prescribed relative framing, the restored geometric quotient declarations, and the literal AlgebraicTopology import/`#check` gate are improvements. Do not undo them while repairing the remaining issues.

## Recommended next revision and validation

First make the small geometric examples fit: empty/proper-open atlas codes, genuine closed framings, disk exteriors, the proper product defect tube, and empty/disconnected/product bordisms with the correct normal and orientation data. Then rebuild the quotient, historical-comparison and exact-sequence constructions on those definitions. Settle the elementary Wall formations and coefficient-sensitive James calculation before using their group calculations downstream.

This is a mathematical and source-level review, **not a claim to have run a local Lean build**. The successful head-associated CI run previously inspected establishes elaboration of roadmap declarations with `sorry`, not the truth of these proposed constructors. No approval or merge is being issued.

CBirkbeck (changes requested):

**Head reviewed and rechecked:** `34167306e5a1726e2574b039dc1a27f913985f0f`.

## Disposition: substantial changes still required

This is still the head covered by the September 8, September 10 and September 17 reviews. This review consolidates the outstanding repairs into an implementation checklist; it does not present the unchanged findings as new author regressions.

The objections are to the meanings and hypotheses of proposed constructors, not to the ambition of the programme or to the legitimate use of `sorry` in roadmap signatures. The [September 8 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/284#pullrequestreview-5139196960) and [September 10 review](https://github.com/TauCetiProject/TauCetiRoadmap/pull/284#pullrequestreview-5171898005) contain the earlier detailed counterexamples.

## 1. Atlas realization: replace literal transition equality by overlap agreement

**Locations:** `Suggested.lean`, `SmoothAtlasCode.realization_atlas` and `SmoothAtlasCode.realized_transition`.

The current statement

```lean
(C.realizedChart i).symm ≫ₕ C.realizedChart j = C.transition i j
```

equates partial homeomorphisms literally, including their total-function values outside the source/target. Geometric gluing only determines the transition on its overlap.

The one-chart example with chart domain `(-1,1)` and self-transition `OpenPartialHomeomorph.ofSet` already contradicts the two realization theorems. The singleton atlas forces the realized chart's source to be the entire realization. Literal self-transition equality then forces its target to be `(-1,1)` but also forces `e (e.symm t) = t` for every real `t`, including `t = 2`. The latter cannot lie in that target.

**Repair:** use the appropriate `OpenPartialHomeomorph.EqOnSource` relation (`≈`) with explicit overlap/source identifications. Do not prescribe irrelevant extensions outside the overlaps.

**Acceptance test:** realize a single proper nonempty open chart, not just a global Euclidean chart. The relevant pinned conventions are in [Defs.lean](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/OpenPartialHomeomorph/Defs.lean) and [Composition.lean](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/OpenPartialHomeomorph/Composition.lean).

## 2. Atlas realization: handle empty charts without excluding the empty manifold

**Locations:** `SmoothAtlasCode`, `Realization`, and `realizedChart`.

One listed chart with empty domain is still permitted. Its realization is empty, but `realizedChart` must supply a partial homeomorphism whose inverse is a total function from the inhabited Euclidean model into that empty realization.

**Repair:** require every listed chart domain to be nonempty while allowing zero charts, or prune empty charts and construct realized charts only for the remaining indices. A global nonemptiness assumption on all manifolds is not a repair: framed bordism needs the empty cycle.

**Acceptance tests:** a zero-chart empty realization; a rejected/pruned empty chart; one global Euclidean chart; and the proper-open chart from item 1.

## 3. Framings: smooth compatibility is not designated-atlas membership

**Locations:** `TangentFraming.memAtlas`, `StableTangentFraming.memAtlas`, and their consumers.

Both structures still combine a global trivialization with `MemTrivializationAtlas`. At the pin, the canonical tangent bundle's designated trivializations come from manifold charts, and their base sets are the chart sources. Requiring a global member of that atlas therefore requires one chart to cover the entire manifold. For a nonempty compact boundaryless positive-dimensional manifold, that would identify it with a nonempty compact open subset of Euclidean space.

The product-bundle atlas has intersections of the designated base sets, so stabilization does not remove this restriction.

**Repair:** retain the canonical tangent bundle and topology, but require a global linear trivialization with actual smooth compatibility, or an actual smooth vector-bundle equivalence to the trivial bundle. Do not replace the canonical instances to make atlas membership hold.

**Acceptance tests:** the Lie-group framing of `S³`, its stabilization, and a stable framing of a standard sphere.

Pinned sources: [tangent-bundle construction](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Geometry/Manifold/VectorBundle/Tangent.lean), [bundle core/designated atlas](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/FiberBundle/Basic.lean), and [product bundles](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/Topology/FiberBundle/Constructions.lean).

## 4. Disk exteriors and proper defect tubes need different boundary data

**Locations:** `SmoothEmbeddedClosedDisk.closedExteriorCharted`, `closedExteriorManifold`, `exteriorCollar`, `standardClosedDiskInSphere`, and `AlmostFramedBordismWitness.defectTrack`.

Moving the exterior collar into the closed complement was a correct earlier repair. However, that closed exterior still uses the ambient chart model. The closed exterior of a standard disk in a boundaryless sphere is a disk with boundary, not a boundaryless Euclidean-model manifold.

There is also a separate defect-track issue: the reflexive tube `Dⁿ × I ⊂ M × I` has endcaps in the ambient trace boundary. An interior point of an endcap has a neighbourhood in the trace contained in the tube, so the proposed exterior collar cannot approach it from outside the tube. A collar along the whole intrinsic tube boundary is the wrong contract.

**Repair:** give the closed exterior its own manifold-with-boundary model and boundary identification. Model a proper defect tube separately, with named endcaps, endpoint disk identifications, and an exterior collar along the lateral boundary only.

**Acceptance tests:** the actual standard disk/exterior in a sphere and the actual product defect tube. Construct these before asserting almost-framed bordism reflexivity.

## 5. Framing restriction must consume a framed normal bundle

**Location:** `restrictStableFramingAlongCodimOne`.

The current function accepts an arbitrary smooth embedding and returns a stable framing after adding one trivial line, without a dimension equation, codimension-one hypothesis, normal-line trivialization, coorientation, or collar datum.

An embedding supplies a decomposition with the actual normal bundle, not automatically a trivial line. Even codimension one alone is insufficient: the interior of an embedded Möbius band in framed `ℝ³` is nonorientable, whereas a framing of its tangent bundle plus a trivial line would orient it.

**Repair:** consume a chosen normal-line trivialization and the actual bundle decomposition, or specialize to genuine boundary/face inclusions with their collars and outward normals. For higher codimension, use the chosen normal framing. Encode the incoming normal reversal and summand-order conventions in the construction/equations, not just the docstring.

This is a prerequisite for the endpoint framing comparisons in both the defect and filling bordisms.

## 6. Separate general bordisms from connected fillings; restore dimension ranges

**Locations:** `CollaredOrientedManifold.connected`, `boundary_nonempty`, general framed-bordism witnesses, and the h-cobordism cylinder/group declarations.

The common trace is required to be connected with nonempty boundary. An empty-to-empty boundary identification forces the boundary carrier empty, contradicting that field; disconnected cylinders are also excluded. Those restrictions may be appropriate for a particular filling, not for every framed bordism.

The unrestricted h-cobordism cylinder statement also fails at `n = 0`, where an endpoint of homotopy type `S⁰` cannot be homotopy equivalent to the required connected trace. The group declaration needs its intended dimensional scope, rather than silently quantifying over every natural number.

**Repair:** introduce/use a general collared-bordism carrier permitting empty and disconnected traces and boundaries. Put connectedness/nonempty-boundary assumptions on the specific filling and sphere objects that need them, and state the sphere-group results in the roadmap's intended range.

**Acceptance tests:** empty, disconnected, and connected high-dimensional product cylinders with the actual boundary identifications.

## 7. Tie every filling-face orientation and the trace framing to the trace orientation

**Locations:** `StableFillingBoundaryFaces` and `StableFillingBordismWitness`.

The vertical face and its oriented endpoint comparisons are now present; keep that repair. What is missing is comparison of all horizontal/vertical face orientations with the orientation induced from the trace. The trace framing is also not required to induce the trace orientation. Reversing the trace and boundary orientations together leaves the current face maps and framing equations unchanged.

**Repair:** add the derivative/normal-line pullback equations for all faces, with incoming-minus/outgoing-plus and corner conventions, and agreement of the trace framing with its orientation. Coordinate these with the named orientation/boundary suppliers rather than introducing another public duplicate.

**Acceptance test:** the oriented/framed product filling bordism, including its vertical face and corner intersections.

## 8. Use the full elementary-zero relations for odd Wall formations

**Locations:** `OddWallFormation.ElementaryModification`, `hyperbolicWallFormation`, and the odd cases of `WallSurgeryDatum.StablyEquivalent`.

The displayed transvections, isometries and complementary-hyperbolic stabilization preserve the rank of the intersection of the two Lagrangians. Consequently `(H(ℤ); L, L)` cannot become zero under those moves, although it is an elementary zero formation because the two Lagrangians have a common Lagrangian complement.

**Repair:** implement both standard elementary-zero relations, or an equivalent presentation with a proved comparison. Replacing one fixed stabilizer by another does not supply the missing relation.

**Acceptance tests:** both `(H(ℤ); L, L*)` and `(H(ℤ); L, L)` represent zero before the odd group-vanishing theorem is invoked. The preceding detailed review gives the source comparison with Ranicki's definition.

## 9. Qualify the James tensor-algebra calculation by its coefficients

**Location:** `README.md`, Stage 3B.3.

The tensor-algebra description of `H_*(JX; R)` needs the appropriate coefficient/freeness hypotheses; connectedness alone does not give an unrestricted integral formula. The Moore-space test `X = M(ℤ/2, 2)` detects the missing Tor term: the tensor algebra on reduced integral homology has no degree-five term, whereas the smash-square/James-splitting calculation contributes `ℤ/2` there.

**Repair:** state the coefficient-qualified Bott–Samelson calculation. For the general integral connected James theorem, retain the chain-level or smash-power argument with its Tor terms. Do not weaken the desired James equivalence merely to accommodate a false intermediate formula.

**Acceptance tests:** both `X = S¹` and the Moore-space example.

## What should remain accepted

Keep the corrected compact-generation adjunction, Thom-space factorization through `Σ^k M_+`, connected James scope, expanded Bott proof plan, distinction between absolute parallelizability and prescribed relative framings, geometric quotient declarations, and the literal AlgebraicTopology supplier import/checks.

## Recommended revision order

Fix atlas realization and framing definitions first. Then make standard disks/exteriors, proper defect tubes, and empty/disconnected/product bordisms fit with the correct normal and orientation data. Rebuild the quotient relations and comparison maps only on those corrected carriers. Settle the Wall relations and coefficient-sensitive James calculation before using their group computations downstream.

This is a source-level review and consolidation of outstanding findings, **not a claim to have run a local Lean build**. Successful elaboration with `sorry` does not establish these constructors. No approval, merge, branch edit, or draft-state change is being issued.

PR #271, opened 2026-08-20
Add roadmap: planar and surface topology

mccorvie:
## Summary

- Add a planar topology roadmap covering Jordan separation, surface triangulation, the two-dimensional Hauptvermutung, Schoenflies, and PL/smooth structure uniqueness.
- Add a surface topology roadmap covering generalized maps, finite presentations, surface classification, mapping class groups, and planarity.
- Connect the new roadmaps to the existing conformal mapping and geometric topology roadmaps.

## Validation

- lake build
- git diff --check

## AI assistance

OpenAI Codex (GPT-5) was used for repository setup, build diagnosis, mechanical formatting, and pull-request preparation. Claude Fable 5.1 (via Claude Code) drafted the review-response edits to both roadmaps and their `Suggested.lean` files, the edits to the neighbouring READMEs, and the replies to the review threads, under my direction; I read and approved every edit and reply before it went in.

PR #289, opened 2026-08-27
feat: Roadmap for the Four color theorem

metakunt:
This is meant to be a port from the four color theorem repository in Coq. It is something that I always wanted to do but now with AI it seems to be autoformalisable. Coq has some horrendously ugly proofs, so I think the port will be more idiomatic.

I have generated the roadmap with OpenCode and I have read and iterated upon it, but I haven't completely crossed all t's.

I'm not sure if this is in scope for Tau Ceti, but it would be cool if we could port some major projects from other proof assistants.

The PR title and descriptions are all human-generated.

### DIFF
--- a/TauCetiRoadmap/FourColorTheorem/README.md
+++ b/TauCetiRoadmap/FourColorTheorem/README.md
@@ -1,0 +1,527 @@
+# Roadmap: the four colour theorem
+
+The goal of **FourColorTheorem** is a Lean 4 formalisation of the four colour theorem for
+planar maps, mirroring the structure of the successful
+[Coq proof of G. Gonthier](https://github.com/rocq-community/fourcolor) (the `fourcolor`
+library) and following its division of labour. The centrepiece is the *combinatorial* theorem
+for finite hypermaps,
+
+```lean
+theorem four_color_hypermap
+    (red : Reducibility) (fork : DischargeForkConsistent) (parts : AllReducedParts)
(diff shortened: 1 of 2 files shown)


mccorvie:
The proposed roadmap #271 for surface topology includes combinatorial representations of graphs on surfaces which is intended to serve as the foundation for planar graph theory.  In particular that roadmap includes the five color theorem, and mentions the four color theorem in the roadmap for a roadmap.  I would recommend not accepting both of these pull requests unless they are harmonized.

metakunt:
Ok, so the goal is to then get your PR merged and I'll reinstruct it to generate a roadmap that incorporates your ideas

kim-em:
I don't think this can fly as is. We're not going to use `native_decide` in TauCeti. Also, see the roadmap requirements about "definiteness" (you can't say something is "optional" or "deferred").

kim-em:
That isn't to say that parts of 4CT, or requirements of 4CT, can't be done in Tau Ceti. But I think it needs to be started over with a more focused roadmap that follows Tau Ceti requirements.

metakunt:
Hm, maybe I'm misunderstood but I want to have everything modulo native_decide. The three "carve-outs" are specifically such that neither native_decide nor the axioms enter the codebase. 4CT in Coq has a bunch of jobs that do proof by reflection, but apart from some heavy calculations the rest could be ported I think.

I want to use this 4CT as a base to do streaming proofs for lean for experimentations, specifically because proofs by reflections are a contentious topic.

PR #225, opened 2026-08-14
docs(SpinRepresentations): split the Layer 7 dependencies

utensil:
## Summary

Layer 7 has two dependency branches:

- the real Clifford recurrence and mod-8 classification consume Layer 0, the pure-axis recurrences, and the real base entries;
- the real Pin and Spin group branch consumes the Layer 2 double cover.

Chevalley supplies the split-matrix, orthogonal-sum, and signature-convention inputs. Lawson–Michelsohn supplies the mixed recurrence and periodic classification. This edit records that split in the Layer 7 description, Ordering paragraph, and matching `cliff_bott` documentation without changing a target or signature.

## Lean impact

The algebraic branch is represented by merged [TauCeti#2550](https://github.com/TauCetiProject/TauCeti/pull/2550), [#2524](https://github.com/TauCetiProject/TauCeti/pull/2524), and [#2804](https://github.com/TauCetiProject/TauCeti/pull/2804). Under literal scope enforcement, the blanket ordering currently blocks review of these next Lean slices:

- the quaternion recurrence `Cliff(p,q+2) ≃ Cliff(q,p) ⊗ ℍ`;
- its eight-step consumer `Cliff(p+8,q) ≃ Cliff(p,q) ⊗ M₁₆(ℝ)`;
- the full matrix/complex/quaternionic classification indexed by `(q-p) mod 8`.

All three compile locally against the merged recurrence chain. This PR states their dependency boundary; it does not waive their code review.

## Audits and validation

A source-dependency audit checked the quaternion recurrence and its eight-step consumer against their exact imports and declarations; their dependency closure ends at Layer 0, the real base entries, and the algebraic recurrence chain. A roadmap audit identified the blanket Ordering sentence as the scope blocker.

The literature audit used locally available literature and the periodicity convention audit. It verified TauCeti's signature swap, generator equations, and recurrence direction, rejected a locally unavailable Atiyah–Bott–Shapiro citation, and corrected the exact roles of Chevalley and Lawson–Michelsohn.

The branch is rebased on current `main`; README/`Suggested.lean` consistency, whitespace, identity, and metadata checks pass. An earlier independent review approved the source roles, convention bridge, and dependency split; this revision removes only redundant negative explanations.

## AI assistance

Prepared with GPT-5.6 Sol high; the author checked the sources, convention bridge, dependency split, and wording.

PR #290, opened 2026-08-27
docs(SpinRepresentations): drop roadmap Layer references from Suggested.lean

kim-em:
This PR removes the roadmap's "Layer" vocabulary from `SpinRepresentations/Suggested.lean` and replaces a chapter-wide Lawson-Michelsohn citation with a precise one.

`Suggested.lean` is a Lean code document, so it should use terms that fit in the final Lean source rather than roadmap planning concepts, per https://github.com/TauCetiProject/TauCetiRoadmap/pull/225#discussion_r3793323400. All 32 `Layer` references are gone: section headings lose the `Layer N:` prefix, and each in-prose reference names the object it meant (`Layer 3's bivectorLieRing` becomes `bivectorLieRing`, `the Layer-1 structure theorem` becomes `the structure theorem`, `Independent of Layers 6-8` becomes `Independent of the exceptional isomorphisms, the real forms, and triality`, and so on). Nothing outside the file referenced these headings.

The Lawson-Michelsohn entry now cites sections rather than the whole of Chapter I: §I.2 for Pin and Spin, §I.3 for the algebras, §I.4 for the classification, §I.5 for the representations, §I.10 for the `(1,1)`-periodicity theorem. `cliff_bott` gains the specific pointer: it is Theorem I.4.1 (4.3), the two recurrences it needs are (4.1) and (4.2), eight-periodicity is Theorem I.4.3 (4.11)-(4.12) with `Cl(8,0) = M₁₆(ℝ)` at (4.14), and the tables are Tables I and II.

It also records a trap: Lawson-Michelsohn's `Cl(r,s)` is this roadmap's `Cliff(s, r)`. Their §I.4 (4.0) sets `Cl(1,0) = ℂ` and `Cl(0,1) = ℝ ⊕ ℝ`, while the base entries here are `Cliff(1,0) ≅ ℝ × ℝ` and `Cliff(0,1) ≅ ℂ`. Anyone following the citation without that note gets the mirror image of the table.

Every claim above was checked against the book rather than recalled: §I.4 pp. 25-29, including the section list on the contents page.

Only comments and docstrings change; no target, signature or statement is touched.

🤖 Prepared with Claude Code

### DIFF
--- a/TauCetiRoadmap/RepresentationTheory/SpinRepresentations/README.md
+++ b/TauCetiRoadmap/RepresentationTheory/SpinRepresentations/README.md
@@ -692,7 +692,11 @@ the CAR instance is a self-contained claimable unit needing only the abstract re
   module, the highest-weight theory, and the fundamental representations of `Bₗ` and `Dₗ`.
 - H. B. Lawson, M.-L. Michelsohn, *Spin Geometry*, Princeton (1989), Chapter I - the definitive account of
   Clifford algebras `Cliff(p, q)`, their Bott-periodic classification (the eightfold table), the Pin and Spin
-  groups as double covers, and the real and complex spinor representations.
+  groups as double covers, and the real and complex spinor representations. By section: §I.2 for Pin and
+  Spin, §I.3 for the algebras themselves, §I.4 for the classification (Theorem 4.1 for the three
+  recurrences, Theorem 4.3 for eight-periodicity, Tables I and II for the tables), §I.5 for the
+  representations, and §I.10 for the `(1,1)`-periodicity theorem. Their `Cl(r,s)` is this roadmap's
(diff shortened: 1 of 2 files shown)


CBirkbeck (dismissed):
## Approval by ChatGPT Pro

**Head reviewed:** `0e8969c12b956c0ea8af66ba61a8dadecd6c8e29`.

Approving the documentation changes reviewed on 8 September 2026. The head is unchanged from that review.

The replacements of roadmap-layer references by the relevant mathematical objects preserve the intended meaning. The more precise Lawson–Michelsohn references and the explicit reversal between their Clifford-signature convention and this roadmap's convention are useful and correct in the reviewed changes.

No theorem statement or target signature is changed by this PR. I found no blocking issue in its diff. This approval is limited to these documentation changes, not a fresh certification of every existing SpinRepresentations target. No local Lean build was run.

PR #277, opened 2026-08-22
progress: AdicSpaces report for 2026-08-21

mccorvie:
This PR records progress on the AdicSpaces roadmap for the window `c9165b9..01f38d7` of TauCeti `main`, covering 102 merged pull requests.

`STATUS.md` is rewritten as a snapshot at `01f38d7`; `PROGRESS.md` gains one appended section for the window. Both files are machine-owned and their prose is not security-validated.

Pull requests in the window: #2968, #2972, #2975, #2987, #2989, #3009, #3035, #3037, #3039, #3040, #3043, #3044, #3046, #3048, #3052, #3060, #3062, #3066, #3093, #3099, #3101, #3107, #3113, #3115, #3126, #3130, #3137, #3141, #3142, #3144, #3158, #3174, #3181, #3186, #3187, #3205, #3207, #3209, #3217, #3240, #3244, #3251, #3255, #3258, #3269, #3272, #3274, #3287, #3309, #3342, #3352, #3353, #3357, #3365, #3611, #3612, #3624, #3625, #3627, #3629, #3632, #3668, #3669, #3671, #3672, #3675, #3694, #3733, #3735, #3736, #3737, #3742, #3743, #3749, #3751, #3797, #3804, #3832, #3865, #3877, #3882, #3891, #3907, #3918, #3932, #3947, #3963, #3965, #3966, #3972, #3988, #3994, #3998, #4000, #4011, #4015, #4021, #4031, #4047, #4052, #4059, #4069

Generated by [TauCetiProgress](https://github.com/TauCetiProject/TauCetiProgress) at `880e8b9737973bfbd8f1f214f4ac2ded67f5b856`.

🤖 Prepared with Claude Code

PR #294, opened 2026-08-30
roadmap(AdicSpaces): analytic-locus inputs and a milestone for adic morphisms

CBirkbeck:
Two places where the roadmap named a consumer without making its input a target. Both are in the
analytic-locus chain, and the second also adds the milestone that chain terminates in.

## §2.3 — the inputs Proposition 7.49(2) consumes

The section asked for Proposition 7.49(2) but not for what its proof runs on. Now stated as
milestones:

- **Microbiality of analytic valuations** (Remark 7.40(5)): some convex subgroup of the value
  group has height-one quotient. Two things are pinned about how this is said. Height stays a
  condition on convex subgroups — "height at most one" is that the only convex subgroups are `⊥`
  and `⊤`, height exactly one adds nontriviality — so no rank or height theory is introduced. And
  the condition is on the value group the valuation *attains*, not on the ambient group of units:
  the ambient reading is satisfied by trivial valuations, whose attained value group is not of
  height one.
- **The height-one vertical generization** a microbial valuation admits, by coarsening along the
  convex subgroup the definition supplies, with the predicate pinned from both sides — rank-one
  valuations are microbial, trivial-value-group valuations are not.
- **Lemma 7.45**: for a complete affinoid ring and a prime ideal that is not open, an analytic
  point of height one whose support contains that prime, with the noetherian refinement where the
  ideal of definition is principal and the valuation may be taken discrete.

The height-one vertical generization is what the proof of Proposition 7.49(2) runs on, and Lemma
7.46(2) consumes Lemma 7.45 in turn — so the chain is now gap-free end to end.

## Layer 5 §5.2 — adic morphisms

A dedicated milestone ahead of locally finite type, and the layer's references now run through
Wedhorn §8.4 rather than stopping at §8.3.

- **Definition 8.38, the adic predicate itself.** Neither Mathlib nor Tau Ceti has a predicate for
  an adic homomorphism — Mathlib's `IsAdic`, `IsAdicComplete` and `IsAdicCauchy` are conditions on
  a topology, not on a map — so the predicate is a target here, stated independently of the chosen
  pairs of definition, with the API its consumers need: identity and composition are adic, and
  adicness is equivalent to openness of the extended ideal of definition. The section also records
  that a containment between chosen ideals of definition is *strictly stronger* than adicness,
  since ideals of definition are comparable only up to powers.
- **Lemma 7.46, parts stated separately at their own hypotheses.** Part (1): the non-analytic
  direction needs only continuity, the analytic direction needs adicness; both directions and the
  locus equality `f⁻¹(Spa(A,A⁺)^a) = Spa(B,B⁺)^a`. Part (2) consumes Lemma 7.45 and keeps the
  completeness hypothesis carried there — it is not stated under continuity alone.
- **Proposition 8.39** — adic iff the analytic locus is preserved, while every morphism preserves
  the non-analytic locus — proved from Lemma 7.46 directly, and **Corollary 8.40**, the induced
  affinoid ring maps.

Sections 5.3, 5.4 and 5.5 are the former 5.2, 5.3 and 5.4, renumbered. No section is referred to
by number anywhere else in the file, so nothing else moves.

## Scope

`TauCetiRoadmap/AdicSpaces/README.md` only — 50 insertions, 5 deletions, no other file. The
generated `STATUS.md` and `PROGRESS.md` are untouched, so this does not collide with the open
progress PR #277, which changes exactly those two generated files for this roadmap.

`Suggested.lean` is unchanged. The adic predicate does not exist in Mathlib or Tau Ceti, so its
signature cannot be prototyped against a pinned API, and the README asks for an honest `sorry` or
an omission rather than a `Prop`-typed placeholder that asserts nothing.

## Code branches this scope covers

Both stay open; neither is a dependency of this PR.

- TauCetiProject/TauCeti#5017 — `Valuation.IsMicrobial`, its height-one quotient, and the
  vertical generization by coarsening, together with the two pinning results. This is the
  microbial/vertical-generization substrate §2.3 now names.
- TauCetiProject/TauCeti#5025 — Lemma 7.46(1), both directions plus the locus equality. These are
  the two directions Proposition 8.39 is proved from.

## AI assistance

This amendment was written by Claude (Opus 5).


CBirkbeck:
Closing: this roadmap amendment was opened from an incorrect interpretation of the owner's scope ruling. The project policy is to rely on the existing roadmap and contest an incorrect scope review; new roadmap PRs require explicit owner approval. The linked TauCeti code PRs remain open. Branch preserved.

PR #297, opened 2026-08-31
feat: roadmap for algebraic vector bundles

0xSimex:
Adds a roadmap entry, **`TauCetiRoadmap/AlgebraicVectorBundles/`**, for the scheme-theoretic theory of algebraic vector bundles.

## Scope

- finite locally free sheaves, locally constant rank, pullback, tensor, dual, symmetric/exterior powers, and determinant;
- the relative-spectrum anti-equivalence
  ```
  QCAlg(X)ᵒᵖ ≃ AffSch/X;
  ```
- the Stacks/EGA correspondence
  ```
  QCoh(X)ᵒᵖ ≃ LinearSch(X),
  ```
  and, after restricting and dualizing,
  ```
  FinLocFree(X) ≃ GeomVB(X);
  ```
- the section-valued universal property of the geometric total space and its compatibility with pullback and the standard algebra of bundles.

The conventions distinguish `linearSpec(F) = Spec_X Sym(F)`, which represents linear functionals and is contravariant, from `totalSpace(E) = Spec_X Sym(Eᵛ)`, which represents sections and is covariant.

## Existing boundary

The roadmap was audited against the pinned Mathlib `05ae010` and Tau Ceti `e8af08d`. It consumes the existing `IsQuasicoherent`, `IsFinitePresentation`, `IsLocallyFree`, `tildeEquiv`, `InvertibleSheaf`, and `FinitelyPresentedSheaf` APIs, and specifies the sheaf-monoidal, relative-`Spec`, and geometric-total-space layers developed by the roadmap.

It also coordinates the intended shapes with [mathlib4#27098](https://github.com/leanprover-community/mathlib4/pull/27098), [mathlib4#39553](https://github.com/leanprover-community/mathlib4/pull/39553), [mathlib4#39989](https://github.com/leanprover-community/mathlib4/pull/39989), and [mathlib4#40194](https://github.com/leanprover-community/mathlib4/pull/40194), following compatible theorem names and interfaces.

## Successors

Projective, Grassmann, and flag bundles, together with their quotient-classifying universal properties, are identified as a separate successor roadmap. Chow groups and characteristic classes then form a further successor, followed by a cohomological-realization roadmap adding cycle classes and Betti/de Rham comparison. Together with the merged Hodge-structures roadmap #49, this gives the interface needed to formulate the Hodge Conjecture.

`Suggested.lean` records representative target signatures for the three layers, including the relative-Spec and vector-bundle equivalences and the section-valued universal property of the geometric total space.

Roadmap: AlgebraicVectorBundles

*(Disclosure: this roadmap and PR description were drafted and revised with OpenAI Codex under my direction and editing.)*


CBirkbeck (changes requested):
> **🤖 AI-authored review.** This review was written jointly by **Claude** (Anthropic, via Claude Code) and **ChatGPT 5.6 Pro** (OpenAI), and is being posted from @CBirkbeck's account on their behalf. Please read it as a Claude + ChatGPT 5.6 Pro review, not a human one.

# Review of TauCetiRoadmap PR #297

**Assessment: request changes.**

I reviewed the current head, `a4ef07f…`. The roadmap has a good overall direction, particularly the distinction between

$$
\operatorname{linearSpec}(\mathcal F)=\operatorname{Spec}_X\operatorname{Sym}(\mathcal F)
$$

and

$$
\operatorname{Tot}(\mathcal E)
 =\operatorname{Spec}_X\operatorname{Sym}(\mathcal E^\vee),
$$

together with the resulting contravariant and covariant functorialities. The relative-Spec anti-equivalence, base-change goals, and section-valued universal property are also the right major pieces. However, the two proposed geometric target categories currently have the wrong morphisms, the determinant functor is not defined on the stated category, and several of the most important constructions are represented only by unrelated opaque `sorry` definitions. The current head also does not build.

## 1. `GradedVectorBundle` and `GeometricVectorBundle` have the wrong morphisms

This is the main blocking issue.

The roadmap defines

```lean
abbrev GradedVectorBundle (X : Scheme) :=
  (linearSpec X).EssImageSubcategory

abbrev GeometricVectorBundle (X : Scheme) :=
  (totalSpace X).EssImageSubcategory
```

and then proposes

```lean
noncomputable def linearSpecEquiv :
    (QuasicoherentSheaf X)ᵒᵖ ≌ GradedVectorBundle X

noncomputable def totalSpaceEquiv :
    FiniteLocallyFreeSheaf X ≌ GeometricVectorBundle X
```

Mathlib’s `EssImageSubcategory` is a **full** subcategory of the target category. Thus these categories contain every morphism of affine schemes over $X$ between objects in the essential image, not merely grading-preserving or fibrewise-linear morphisms.

Already over $X=\operatorname{Spec} k$, the rank-one object has total space $\mathbb A^1_k$. The scheme endomorphism

$$
t\longmapsto t^2
$$

is a morphism over $k$, and hence appears in both proposed full essential-image categories. It does not arise from a linear map $k\to k$. Consequently the induced `linearSpec` and `totalSpace` functors are not full, and the stated equivalences do not describe the intended categories.

The README actually gives the correct idea: morphisms in the target must preserve the grading, or equivalently preserve the zero, addition, and scalar-multiplication structure. But that structure never appears in `Suggested.lean`.

I suggest choosing one of the following two implementations.

### Option A: graded quasicoherent algebras

Define a category of nonnegatively graded quasicoherent $\mathcal O_X$-algebras, with **graded algebra morphisms**, and then define the full subcategory of algebras freely generated in degree one. For example, schematically:

```lean
GradedQuasicoherentAlgebra X
gradedSymmetricAlgebra :
  QuasicoherentSheaf X ⥤ GradedQuasicoherentAlgebra X

FreeGradedQuasicoherentAlgebra X :=
  (gradedSymmetricAlgebra X).EssImageSubcategory
```

Using an essential image is appropriate here, because the ambient morphisms are already required to preserve the grading. Mathlib has generic graded-object infrastructure that should be used rather than representing a grading by an unrelated family of modules.

The target condition should preferably be an isomorphism of graded algebras

$$
\operatorname{Sym}^{\mathrm{gr}}(\mathcal A_1)\cong\mathcal A,
$$

not merely separate module isomorphisms

$$
\operatorname{Sym}^n(\mathcal A_1)\cong\mathcal A_n.
$$

The latter do not by themselves record compatibility with multiplication and the unit.

### Option B: affine module schemes

A more geometric API would define an affine $\mathcal O_X$-module scheme: an affine $V\to X$ with

```text
zero     : X ⟶ V
add      : V ×ₓ V ⟶ V
smul     : 𝔸¹ₓ ×ₓ V ⟶ V
```

satisfying the module laws. Morphisms would preserve all three operations. This directly excludes polynomial maps such as $t\mapsto t^2$.

There is already generic category-theoretic infrastructure for module objects in a cartesian monoidal category in `CategoryTheory.Monoidal.Cartesian.Mod`, so this should be investigated before introducing a one-off category of geometric bundles.

Under either implementation, keep separate functors

```lean
linearSpecScheme :
  (QuasicoherentSheaf X)ᵒᵖ ⥤ AffineSchemeOver X

linearSpec :
  (QuasicoherentSheaf X)ᵒᵖ ⥤ LinearSchemeOver X
```

with a theorem saying that forgetting the module or grading structure recovers `linearSpecScheme`. Then define the equivalence using the **structured** functor, not as an arbitrary opaque equivalence.

Also, the name `GradedVectorBundle` is misleading for an arbitrary quasicoherent sheaf. Unless finite local freeness is imposed, the resulting object is a linear scheme, not a vector bundle. I would use names such as `LinearScheme` or `QuasicoherentLinearScheme`, reserving `GeometricVectorBundle` for the finite locally free case.

## 2. The proposed determinant functor does not exist on all morphisms

The target

```lean
noncomputable def determinant :
    FiniteLocallyFreeSheaf X ⥤ InvertibleSheaf X
```

is not correct when `FiniteLocallyFreeSheaf X` contains all finite locally free sheaves and all module morphisms.

For a map $\mathcal E\to\mathcal F$, the usual exterior-power map

$$
\bigwedge^r\mathcal E\longrightarrow\bigwedge^r\mathcal F
$$

gives a map between determinant lines only when the ranks agree. The standard determinant construction on a morphism is stated under equality of the source and target ranks.

There is no functorial repair obtained by declaring the determinant map to be zero when the ranks differ. For example, take a split inclusion and projection

$$
\mathcal O_X\longrightarrow \mathcal O_X^2
   \longrightarrow \mathcal O_X
$$

whose composite is the identity. Sending both unequal-rank maps to zero would send their composite to zero, whereas the determinant of the identity is the identity.

There are several correct APIs:

```lean
determinantObj :
  FiniteLocallyFreeSheaf X → InvertibleSheaf X
```

with functoriality only for isomorphisms;

```lean
determinant (r : ℕ) :
  FiniteLocallyFreeSheafOfRank X r ⥤ InvertibleSheaf X
```

for fixed rank;

or a determinant functor on the core/groupoid of finite locally free sheaves. A Knudsen–Mumford determinant functor on the exact category of vector bundles is another possibility, but that is a substantially larger piece of infrastructure.

For this roadmap, the fixed-rank functor is probably the cleanest first target.

## 3. The closed monoidal structure does not restrict to arbitrary quasicoherent sheaves as stated

The README asks for a symmetric monoidal closed structure on `X.Modules` and says that it should be restricted to quasicoherent sheaves.

The tensor product does restrict to quasicoherent sheaves. The ordinary sheaf internal Hom does **not** preserve quasicoherence for arbitrary quasicoherent source and target. A standard sufficient hypothesis is that the source be of finite presentation; in particular, for finite locally free $\mathcal E$,

$$
\mathcal H\!om(\mathcal E,\mathcal F)
 \cong \mathcal E^\vee\otimes\mathcal F
$$

is quasicoherent.

The roadmap should say:

* `X.Modules` is symmetric monoidal closed;
* `QuasicoherentSheaf X` is symmetric monoidal;
* internal Hom from a finitely presented, and especially a finite locally free, source lands in quasicoherent sheaves;
* finite locally free sheaves form a rigid symmetric monoidal category.

The closed structure should also be pinned by an adjunction, not merely by an opaque bifunctor:

```lean
tensorLeftAdjunction (E : X.Modules) :
  tensoringLeft E ⊣ internalHom.obj (Opposite.op E)
```

or an equivalent natural Hom equivalence. Otherwise the target does not specify that the proposed `internalHom` is the internal Hom for the proposed tensor product.

## 4. The intrinsic geometric categories promised in the README are absent from `Suggested.lean`

The README says that a graded vector bundle should intrinsically consist of:

* an affine morphism $p:V\to X$;
* a grading on $p_*\mathcal O_V$;
* an identification of degree zero with $\mathcal O_X$;
* isomorphisms from the symmetric powers of degree one;
* grading-preserving morphisms.

None of these structures or recognition theorems appears in `Suggested.lean`. Instead, the target category is only the full essential image discussed above.

This leaves out one of the main advertised achievements of the roadmap: recognizing vector bundles intrinsically rather than simply naming the image of a functor. At minimum there should be target declarations for:

```lean
GradedLinearScheme X
GradedLinearScheme.Hom
gradedLinearSchemeOfSymmetricAlgebra
degreeOne :
  GradedLinearScheme X ⥤ QuasicoherentSheaf X
linearSpecDegreeOneIso
degreeOneLinearSpecIso
```

and the analogous finite locally free subcategory.

A useful factorization would be:

$$
\mathrm{QCoh}(X)^{\mathrm{op}}
 \simeq
 \{\text{free graded quasicoherent algebras}\}^{\mathrm{op}}
 \simeq
 \{\text{graded affine linear schemes over }X\}.
$$

The first equivalence is algebraic and the second is relative Spec. This separates two conceptually different arguments and provides much better intermediate APIs.

## 5. The current target file does not build

At the current head, dependency resolution succeeds but the Lean build fails with two errors.

The first is in `finiteLocallyFreeToFinitelyPresented`: the use of the generic full-subcategory inclusion does not provide the required finite-presentation instance for the arbitrary object introduced internally. This can be avoided by defining the object map directly from `E.property.2`, or by using the appropriate object-property lift with an explicit preservation proof.

The second is in `pullbackObjIso`: the selected `Scheme.Modules.pullback.mapIso` is not an isomorphism of the expected objects. This is also evidence that `pullback` should be defined by lifting the existing module pullback functor through the finite-locally-free object property. Then the forgetful comparison should be supplied by the standard lifted-subcategory API, and will often be definitional rather than requiring a custom `mapIso`.

I would require a green build before merging the roadmap, since these are not warnings in proof bodies but failures in the proposed declaration shapes.

## 6. The relative-Spec equivalence needs to be decomposed into named functors and universal properties

The bare target

```lean
noncomputable def relativeSpecEquiv :
    (QuasicoherentAlgebra X)ᵒᵖ ≌ AffineSchemeOver X
```

conceals almost all of the mathematical work. Relative Spec and its affine anti-equivalence are correct goals, including the pushforward-algebra and base-change statements.

The roadmap should expose at least:

```lean
relativeSpec :
  (QuasicoherentAlgebra X)ᵒᵖ ⥤ AffineSchemeOver X

affineFunctions :
  AffineSchemeOver X ⥤ (QuasicoherentAlgebra X)ᵒᵖ

relativeSpecHomEquiv :
  (T ⟶ relativeSpec.obj (Opposite.op A)) ≃
    (pullbackAlgebra T.hom).obj A ⟶ 𝒪_T

relativeSpecPushforwardIso :
  affineFunctions.obj (relativeSpec.obj (Opposite.op A)) ≅
    Opposite.op A

relativeSpecCounitIso :
  relativeSpec.obj (affineFunctions.obj V) ≅ V

relativeSpecBaseChangeIso :
  baseChange f (relativeSpec.obj (Opposite.op A)) ≅
    relativeSpec.obj (Opposite.op ((pullbackAlgebra f).obj A))
```

together with naturality and the triangle identities needed to construct the equivalence.

There is substantial hidden infrastructure here:

* pushforward of the structure sheaf along an affine morphism as a quasicoherent algebra;
* pullback of quasicoherent algebras;
* the strong symmetric monoidal structure on module pullback;
* comparison with ordinary affine `Spec` when $X$ is affine;
* compatibility with composition and identity base changes.

These should appear as explicit dependency layers rather than being hidden inside `relativeSpecEquiv`.

## 7. The universal property of total space needs naturality

The proposed

```lean
totalSpaceHomEquiv (E) (T) :
  (T ⟶ totalSpace E) ≃
    (𝒪_T ⟶ pullback T.hom E.obj)
```

is the right pointwise statement, but a collection of unrelated equivalences is not enough. The universal property should be natural in both $T$ and $E$. In particular, a morphism $\mathcal E\to\mathcal F$ must correspond under the equivalence to postcomposition on sections, and a morphism $T'\to T$ must correspond to pullback of sections.

I suggest either a Yoneda-level natural isomorphism or explicit naturality lemmas:

```lean
totalSpaceHomEquiv_natural_left
totalSpaceHomEquiv_natural_right
totalSpaceHomEquiv_id
totalSpaceHomEquiv_comp
```

The construction should also be visibly obtained from the relative-Spec universal property and finite-locally-free duality:

$$
\operatorname{Hom}_X
  \bigl(T,\operatorname{Spec}_X\operatorname{Sym}(\mathcal E^\vee)\bigr)
 \cong
\operatorname{Hom}_{\mathcal O_T}
  \bigl(f^*\mathcal E^\vee,\mathcal O_T\bigr)
 \cong
\operatorname{Hom}_{\mathcal O_T}
  \bigl(\mathcal O_T,f^*\mathcal E\bigr).
$$

Add a characteristic theorem such as

```lean
totalSpace_eq_linearSpec_dual
```

or define `totalSpace` through `linearSpec` and `dual` so that this relation is definitional.

## 8. Too many opaque definitions lack characteristic equations

The target file currently introduces opaque definitions for:

* finite-locally-free pullback;
* tensor product;
* internal Hom;
* dual;
* symmetric and exterior powers;
* determinant;
* symmetric algebra;
* relative Spec;
* linear Spec;
* total space;
* the two summit equivalences.

Many have no theorem identifying their underlying object with the intended existing construction. With `sorry`, these declarations could all be satisfied by unrelated choices, so the target file does not currently enforce the architecture described in the README.

For each construction, add a projection or comparison theorem. For example:

```lean
@[simp]
theorem pullback_obj_val :
    ((pullback f).obj E).obj =
      (Scheme.Modules.pullback f).obj E.obj

@[simp]
theorem tensor_obj_val :
    (tensorFiniteLocallyFree.obj (E, F)).obj =
      E.obj ⊗ F.obj

theorem dual_obj_iso :
    (dual.obj (Opposite.op E)).obj ≅
      internalHomObj E.obj (𝒪 X)

theorem symmetricPower_obj_iso :
    ((symmetricPower n).obj E).obj ≅
      SymmetricPower n E.obj

theorem exteriorPower_obj_iso :
    ((exteriorPower n).obj E).obj ≅
      ExteriorPower n E.obj

theorem linearSpec_obj_iso :
    affineSchemeOverForget.obj
      ((linearSpec X).obj (Opposite.op F)) ≅
      relativeSpecScheme (symmetricAlgebra.obj F)

theorem totalSpace_obj_iso :
    affineSchemeOverForget.obj ((totalSpace X).obj E) ≅
      relativeSpecScheme
        (symmetricAlgebra.obj ((dual X).obj (Opposite.op E)))
```

For the proposed equivalences, state that their forward functors are the named functors:

```lean
linearSpecEquiv_functor_iso :
  linearSpecEquiv X |>.functor ≅ linearSpecStructured X

totalSpaceEquiv_functor_iso :
  totalSpaceEquiv X |>.functor ≅ totalSpaceStructured X
```

Without such a statement, an opaque equivalence does not establish that `linearSpec` or `totalSpace` is the equivalence.

## 9. Rank should be a bundled locally constant function

The roadmap currently proposes a function

```lean
rank (E : FiniteLocallyFreeSheaf X) (x : X) : ℕ
```

followed by a theorem that it is locally constant. Since local constancy is fundamental to almost every later operation, it would be cleaner to define

```lean
rank :
  FiniteLocallyFreeSheaf X → LocallyConstant X ℕ
```

and derive pointwise notation from evaluation. Rank is locally constant for finite locally free modules.

Then add:

```lean
rankLocus (E) (r : ℕ) : Opens X
rank_pullback
rank_iso
rank_tensor
rank_dual
rank_directSum
rank_symmetricPower
rank_exteriorPower
```

and define the fixed-rank property and subcategory:

```lean
isFiniteLocallyFreeOfRank (X) (r : ℕ) :
  ObjectProperty X.Modules

FiniteLocallyFreeSheafOfRank X r
```

This resolves the determinant problem and gives the local-triviality theorems a precise statement.

## 10. The promised rank-one comparison is only one-way

The README promises to identify rank-one finite locally free sheaves with the existing `InvertibleSheaf` API, but `Suggested.lean` only defines a functor

```lean
invertibleToFiniteLocallyFree
```

There should be an equivalence

```lean
invertibleSheafEquivFiniteLocallyFreeRankOne :
  InvertibleSheaf X ≌ FiniteLocallyFreeSheafOfRank X 1
```

with a theorem identifying the underlying module-sheaf functors.

This is important because TauCeti already has an `InvertibleSheaf` and tensor-product API; the new vector-bundle development should extend that API rather than create an adjacent rank-one theory.

## 11. The fibre entry in the dictionary is imprecise

The README identifies

```text
module fibre E ⊗ k(x)  |  scheme fibre over x
```

These are not literally the same kind of object. The correct statement is

$$
\operatorname{Tot}(\mathcal E)\times_X\operatorname{Spec}\kappa(x)
 \cong
\operatorname{Spec}_{\kappa(x)}
 \operatorname{Sym}\bigl((\mathcal E_x\otimes\kappa(x))^\vee\bigr).
$$

Its $\kappa(x)$-rational points identify with the vectors of the module fibre. The roadmap should distinguish the affine-space scheme associated to the fibre from its set of rational points.

A useful target is:

```lean
totalSpaceFiberIso (E) (x : X) :
  fiber (totalSpace E) x ≅
    affineSpaceOfModule (E.fiber x)
```

together with the induced equivalence on $\kappa(x)$-points.

## 12. The monoidal structure should use the generic subcategory machinery

Mathlib already has generic infrastructure for giving a full subcategory a monoidal structure from an object property closed under the tensor product and containing the unit.

Rather than adding unrelated instances

```lean
modulesMonoidalCategory X
quasicoherentMonoidalCategory X
```

I suggest:

1. build the monoidal and closed structure generically for sheaves of modules over a sheaf of rings;
2. prove that quasicoherence is an `ObjectProperty.IsMonoidal`;
3. prove that finite local freeness is an `ObjectProperty.IsMonoidal`;
4. obtain the full-subcategory structures from the generic API;
5. prove that finite locally free sheaves form a rigid category.

The existing presheaf-of-modules monoidal infrastructure should be the starting point, with sheafification supplying the sheaf tensor product.

This would make the following statements consequences of coherent categorical structure rather than separate constructions:

```lean
tensorFiniteLocallyFree
dual
evaluation
coevaluation
doubleDualIso
internalHomIsoDualTensor
```

Pullback should similarly be a strong symmetric monoidal functor, since that single API supplies the tensor, dual, symmetric-power, exterior-power, and determinant base-change comparisons.

## 13. Base change needs one coherent pseudofunctorial API

The roadmap includes pullback and a few identity/composition examples, but the main constructions all require compatible base-change isomorphisms.

I would introduce named functors and comparisons:

```lean
pullbackQCoh
pullbackFiniteLocallyFree
pullbackQuasicoherentAlgebra
baseChangeAffineSchemeOver

pullbackIdIso
pullbackCompIso
pullbackTensorIso
pullbackDualIso
pullbackSymmetricPowerIso
pullbackExteriorPowerIso

relativeSpecBaseChangeIso
linearSpecBaseChangeIso
totalSpaceBaseChangeIso
```

and record the identity and composition coherence conditions. Otherwise each later layer will construct its own slightly different change-of-base isomorphism.

## 14. Direct sums, flatness, and exact sequences are promised but not targeted

The README includes direct sums, rank formulae, flatness, and locally split short exact sequences, but `Suggested.lean` has no corresponding declarations.

For direct sums and additive structure, add at least:

```lean
finiteLocallyFree_zero
finiteLocallyFree_biproduct
rank_biproduct
totalSpace_biproduct_iso_product
```

For exact sequences, the useful theorem is that a short exact sequence of vector bundles is locally split when the quotient is locally free. The category of vector bundles is naturally treated as an exact category with these sequences.

The assertion that finite locally free is equivalent to finitely presented and flat also requires a definition of flatness for module sheaves, locality of flatness, compatibility with stalks and affine restriction, and the affine finite-presentation-plus-flat theorem. I did not find an existing sheaf-level flatness API at the pinned Mathlib revision. Either add a genuine flatness sublayer or remove this from the core deliverables rather than leaving it implicit.

## 15. The affine comparison should be more than an opaque equivalence

The proposed

```lean
tildeFiniteProjectiveEquiv :
  FiniteProjectiveModule R ≌
    FiniteLocallyFreeSheaf (Spec R)
```

is important enough to expose its compatibility with all subsequent operations:

```lean
tildeFiniteProjectiveEquiv_obj_iso
globalSections_tilde_iso
tilde_tensor_iso
tilde_dual_iso
tilde_symmetricPower_iso
tilde_exteriorPower_iso
tilde_totalSpace_iso
rank_tilde
```

I would also avoid inventing a new bundled `FiniteProjectiveModule` unless necessary. Prefer a full subcategory of the existing `ModuleCat R` cut out by Mathlib’s finite and projective predicates. This will make the existing module tensor, dual, and symmetric/exterior-power APIs available without adapters.

## 16. Add concrete normalization tests

The roadmap would benefit from a few literal endpoint examples:

```lean
totalSpace_free :
  totalSpace (𝒪_X ^⊕ r) ≅ affineSpace X r

totalSpace_unit :
  totalSpace (𝒪_X) ≅ affineLine X

linearSpec_affine :
  linearSpec (ModuleCatSheaf.associated M) ≅
    Spec (SymmetricAlgebra R M)

totalSpace_affineFiniteProjective :
  totalSpace (tilde P) ≅
    Spec (SymmetricAlgebra R (Module.Dual R P))
```

These test the dual convention, the variance, and the relative-Spec normalization. They would catch exactly the kind of accidental reversal that is otherwise easy to introduce in this development.

## Suggested restructuring

The current layers conceal several independent projects. I would split them approximately as follows:

### Layer 0A: sheaves of modules

* tensor product and unit;
* internal Hom and tensor–Hom adjunction;
* symmetric monoidal and closed coherence;
* pullback as a strong symmetric monoidal functor.

### Layer 0B: quasicoherent and finite locally free sheaves

* quasicoherence of tensor;
* finite local freeness, rank and rank loci;
* affine finite-projective comparison;
* direct sums and tensor products;
* rigid duality and dualizable-object characterization.

### Layer 0C: polynomial operations

* symmetric and exterior powers;
* pullback and affine comparison;
* fixed-rank determinant;
* rank-one comparison with `InvertibleSheaf`.

### Layer 1A: quasicoherent algebras and affine morphisms

* quasicoherent algebra category;
* pushforward algebra of an affine morphism;
* relative Spec functor and universal property.

### Layer 1B: relative-Spec anti-equivalence

* unit and counit;
* affine-base comparison;
* base change;
* the actual anti-equivalence.

### Layer 2A: structured linear schemes

* graded-algebra or affine-module-scheme structure;
* linear Spec with its zero, addition and scalar maps;
* full faithfulness and recognition via degree one.

### Layer 2B: geometric vector bundles

* finite locally free degree-one piece;
* local triviality as affine space;
* covariant total-space functor;
* section universal property;
* total-space equivalence and fibre theorem.

This would make the roadmap much more implementable and would give later connections, Higgs-bundle, and Hodge-theoretic developments stable interfaces rather than one very large summit.

# What is already good

Several choices in the roadmap are worth preserving:

* The distinction between `linearSpec F` and `totalSpace E := linearSpec (Eᵛ)` is correct and important.
* The variance of the two functors is handled correctly in the prose.
* Relative Spec and the anti-equivalence between quasicoherent algebras and affine schemes over $X$ is the right foundational theorem.
* The section-valued universal property for total spaces is the right geometric endpoint.
* Keeping finite locally free sheaves as a full subcategory of module sheaves is sensible.
* Exact dependency pins and explicit tracking of upstream Mathlib work are good.
* The roadmap correctly emphasizes base change rather than treating it as an afterthought.
* Reusing TauCeti’s existing invertible-sheaf and finitely-presented-sheaf APIs is the right direction.

# Changes I would require before merge

1. Replace the full essential-image categories in `AffineSchemeOver X` by structured categories whose morphisms are grading-preserving or fibrewise linear.
2. Replace the determinant functor on all finite locally free sheaves by a fixed-rank, isomorphism-only, or exact-category version.
3. Correct the claim that the closed structure restricts to arbitrary quasicoherent sheaves.
4. Add the intrinsic graded/module-scheme categories and recognition theorems promised by the README.
5. Decompose `relativeSpecEquiv` into named functors, universal properties, unit/counit, and base-change theorems.
6. Add naturality to the total-space universal property.
7. Add characteristic equations for every opaque functor and for the forward functors of the summit equivalences.
8. Bundle rank as a locally constant function and add fixed-rank subcategories.
9. Add the rank-one equivalence with `InvertibleSheaf`.
10. Correct the fibre dictionary.
11. Add coherent strong-monoidal pullback and graded-algebra or affine-module-scheme APIs.
12. Make the target file build on the pinned dependency graph.

With those changes, this could become a strong roadmap. At present, the object-level picture is largely right, but the categorical morphisms and the target signatures do not yet formalize the mathematical theorem the README claims.

---

*— Claude (Anthropic) + ChatGPT 5.6 Pro (OpenAI), posted via @CBirkbeck*

0xSimex:
Thanks @CBirkbeck — this was very helpful. I reworked the roadmap so that the geometric side carries the graded/linear structure and therefore has the correct morphisms; I also corrected the fixed-rank determinant and made the relative-Spec and total-space constructions explicit.

🤖 AI disclosure: This revision and comment were prepared with OpenAI Codex under my direction and review.

lsqrl:
Hi @CBirkbeck: just a gentle ping on this one.
Whenever you have a chance, a second pass would be very appreciated, or some further indication so that we can keep the wheel spinning. Thanks again for the detailed review you provided last time!

0xSimex:
Hi @CBirkbeck: just a gentle ping on this one.
Whenever you have a chance, a second pass would be very appreciated, or some further indication so that we can keep the wheel spinning. Thanks again for the detailed review you provided last time!

CBirkbeck:
Sorry yes I'm on it!

CBirkbeck (changes requested):
## 🤖 Review by ChatGPT Pro

**Head reviewed:** `81ebd901a5e49e1c66387612dc51c197b5c8f5c4`.

I reviewed both current files, the previous review and the author's response, relevant definitions at the actual Mathlib/Tau Ceti pins, and newer sheaf-tensor infrastructure. The head is unchanged at submission. The head-associated CI run `33504954123` and its build job succeeded; I did not run a local Lean build.

**Assessment: a focused revision is still needed before approval.** The main mathematical programme is sound, and several previous objections are genuinely resolved. The remaining concerns are chiefly whether the representative declarations express that programme, plus two precise qualifications in the prose.

### Repairs accepted

The essential image is now taken inside graded commutative algebras, rather than the full category of unstructured affine schemes. The new `QuasicoherentLinearScheme.Hom` record includes a graded coordinate map and its compatibility with the scheme map. This is the right repair to the old polynomial-endomorphism problem, subject to the category-instance issue below.

The determinant functor is correctly restricted to a fixed-rank subcategory, with a separate determinant object for locally varying rank. Internal Hom from a finitely presented source is distinguished from tensor closure of arbitrary quasicoherent sheaves. The `linearSpec`/dualized `totalSpace` variance distinction remains correct. The relative-Spec/graded-algebra factorization and separation of the later projective-bundle/Chow/realization projects are also appropriate.

The old build failure should not be re-requested: current CI is green.

## 1. Make the actual category use the new structured morphism record

**Location:** `QuasicoherentLinearScheme`, immediately after `structure Hom`.

The file declares a `Quiver` with `Hom := Hom`, but then independently supplies

```lean
noncomputable instance (X : Scheme.{u}) :
    Category.{u} (QuasicoherentLinearScheme X) := by
  sorry
```

`Category` is not a proposition asserting laws about an already selected `Quiver`. At the pin it extends `CategoryStruct`, which itself contains a `Quiver`. Its type does not require the inherited morphism family to agree with the preceding standalone quiver instance. Therefore this declaration does not yet bind the category used by the later functors/equivalences to `QuasicoherentLinearScheme.Hom`.

Please construct the category with the morphism family explicitly fixed to that record. Pin identity and composition too: for `f : V → W` and `g : W → Z`, the scheme component is `f.schemeHom ≫ g.schemeHom`, while the coordinate component is `g.coordinateHom ≫ f.coordinateHom`. The compatibility and category-law proofs can remain roadmap proof targets.

Define `linearSchemeForget` by the actual object and `schemeHom` projections, and `linearSchemeCoordinates` by the actual coordinate algebra and the opposite of `coordinateHom`. Add the corresponding map equations and a morphism extensionality lemma. An elementary type-checking test should project `.schemeHom` from a categorical morphism and identify it with `(linearSchemeForget X).map f`.

This is not a request to return to the old full-subcategory model. The new record is good; it needs to be the morphism carrier of the public category, not a separate unused record.

Pinned source: [Mathlib Category/Basic.lean](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/CategoryTheory/Category/Basic.lean).

## 2. The new natural universal property must use the actual point and section functors

**Locations:** `totalSpacePoints`, `pullbackSections`, and `totalSpaceUniversalProperty`.

Both endpoint functors are currently independent opaque definitions of the same type. The declaration

```lean
totalSpaceUniversalProperty (X) :
    totalSpacePoints X ≅ pullbackSections X
```

therefore does not express naturality of `totalSpaceHomEquiv`. For example, the two endpoint definitions and this isomorphism could all be implemented using the constant singleton functor, without establishing anything about total-space morphisms or sections. The correctly typed pointwise `totalSpaceHomEquiv` is a separate declaration and does not exclude that implementation of these three targets.

Please define the point functor from `totalSpaceScheme`, `affineSchemeOverForget`, and Yoneda, with an explicit universe lift if needed. Define the section functor to have value

`Hom_{O_T}(O_T, f* E)`

at `E` and `f : T → X`, with maps given by the actual module pullbacks and their identity/composition comparisons. Then derive `totalSpaceHomEquiv` from the natural isomorphism, or add an equation identifying it with the corresponding component.

Likewise, express the promised naturality of `relativeSpecHomEquiv` using its actual Hom/algebra-map functors, rather than leaving it solely in the README.

**Acceptance test:** for the actual free rank-one sheaf, recover `Γ(T,O_T)` and verify that a map `T' → T` induces restriction/pullback of those sections. For a map of free bundles given by a matrix, the induced total-space map should give that matrix on sections. This checks both variances and the maps, not only the cardinalities of the pointwise sets.

## 3. Finish the characteristic comparisons at the key construction boundaries

**Locations:** the relative-Spec, graded-degree-one, dualization, and normalization declarations.

Some new comparisons still end at another opaque helper without reaching the construction they purport to describe. In particular:

* `relativeSpecPushforwardIso` uses `affineFunctions`, but nothing identifies the underlying sheaf algebra of `affineFunctions(V).unop` with the actual `p_* O_V`, or its map with the map induced by a scheme morphism. Its name alone does not impose that condition.
* `gradedAlgebraDegreeOne` has no equation identifying its underlying sheaf with the actual degree-one component `A.obj.X 1`. The later `degreeOne` should be defined through that extraction and `linearSchemeCoordinates`, rather than chosen independently. The standard essential-image inclusion should similarly supply `freeGradedAlgebraForget`.
* `dualAsQuasicoherentOp` is not connected to the previously declared `dual`. Define it through `dual` and the actual inclusion into quasicoherent sheaves, or give a natural comparison identifying precisely that functor. The underlying dual itself should be compared to the L0 internal Hom into `O_X`.
* The worked target `totalSpaceFreeIso` compares two new opaque objects. `freeFiniteLocallyFree X r` has no equation with `SheafOfModules.free` on a finite rank-`r` index, and `affineSpaceOver X r` has no equation with the **already imported** `AlgebraicGeometry.AffineSpace`. As written, this does not test the advertised free-bundle normalization. The analogous `affineSpaceOfModuleFiber` needs the displayed residue-field `Spec(Sym(dual fibre))` construction as its characteristic target.

These are a small number of important boundaries, not a demand to implement every future proof now. A comparison to another new name is useful only when that name is itself tied to the intended object and maps. Proof fields may remain `sorry`; the carrier definitions and comparison statements need to carry the content.

There are concrete suppliers to use. The pinned [AffineSpace.lean](https://github.com/leanprover-community/mathlib4/blob/05ae0103f49b1ad1248f6039bbbad43d8aeb52a9/Mathlib/AlgebraicGeometry/AffineSpace.lean) already provides `AffineSpace`, `AffineSpace.homOfVector`, `AffineSpace.homOverEquiv`, and `AffineSpace.SpecIso`. Use these for the free-object and coordinate tests, rather than introduce an unrelated affine-space placeholder.

Also, newer Tau Ceti now has [the sheafified tensor product and its defining comparison](https://github.com/TauCetiProject/TauCeti/blob/77140d1fa19c0e8b74a4b681f590f69691a30e3d/TauCeti/Algebra/Category/ModuleCat/Sheaf/TensorProduct/Basic.lean), together with [the scheme specialization `Scheme.Modules.tensorProduct`](https://github.com/TauCetiProject/TauCeti/blob/77140d1fa19c0e8b74a4b681f590f69691a30e3d/TauCeti/AlgebraicGeometry/Modules/TensorProduct.lean). This is beyond this PR's pin, not a claim that its old inventory was false. Coordinate with that construction and pin the tensor underlying `modulesMonoidalCategory` to it/the same sheafification construction. The full closed symmetric monoidal coherence remains genuine work; these files should not be described as already supplying all of L0A.

## 4. Qualify the affine-module-scheme reformulation

**Locations:** the introductory “Equivalently” sentence, the definitions section, and L2A's equivalence with the module-scheme presentation.

Every linear scheme defined here has the asserted `A¹_X`-module-scheme structure. The converse is not true for **arbitrary** affine module schemes. Please retain the free-degree-one condition, or an equivalent explicit recognition condition, when describing the module-scheme side of the equivalence.

For example, over `k = F_p`, the affine group scheme

`alpha_p = Spec(k[t]/(t^p))`

is an `A¹_k`-module scheme. On every `k`-algebra `R`, its points are `{x in R | x^p = 0}`, an `R`-module under ordinary addition and scalar multiplication: `(x+y)^p=0` and `(a x)^p=0`. These operations are induced by scheme morphisms and satisfy the module laws.

But its coordinate algebra is not `Sym_k(M)` for any vector space `M`: it has a nonzero nilpotent, whereas a symmetric algebra over a field is a polynomial algebra and is reduced. With the evident grading it is generated in degree one, but not **freely** generated there.

**Repair:** state the module-scheme structure and the correspondence of structure-preserving morphisms for the existing free-graded objects; for an intrinsic module-scheme recognition theorem, carry the condition that the canonical symmetric-algebra map is an isomorphism. In the finite-bundle version, the appropriate local linear trivializations also suffice. Do not identify the entire category of affine module schemes with `QCoh(X)^op`.

The Stacks [vector-bundle definition](https://stacks.math.columbia.edu/tag/01M1) retains precisely the symmetric-power isomorphism condition. The point is to keep that condition in the alternate presentation, not to change the current graded definition.

## 5. Make the internal-Hom exception explicit in the blanket base-change criterion

**Locations:** L0A/L0B and the first cross-cutting acceptance criterion, which asks for arbitrary-base-change theorems for every construction.

The restriction to a finitely presented source correctly gives quasicoherence of internal Hom, but it does not give an internal-Hom **base-change isomorphism under an arbitrary nonflat map**.

Take `R = k[t]`, `S = R/(t)`, `M = R/(t)`, and `N = R`. Then

`S tensor_R Hom_R(M,N) = 0`,

while

`Hom_S(S tensor_R M, S tensor_R N) = Hom_S(k,k) = k`.

Both input modules are finitely presented. This is an affine example within the proposed scope.

Please distinguish the general comparison morphism from its isomorphism hypotheses: finite-presentation source with **flat** base change is a standard sufficient case; finite locally free source allows arbitrary base change. Similarly, an unrestricted-source stalk formula should be a comparison map unless its isomorphism hypotheses are supplied. [Stacks tag 0C6I](https://stacks.math.columbia.edu/tag/0C6I) states the finite-presentation/flat version explicitly.

This qualification does not weaken the correct arbitrary-base-change statements for relative Spec, symmetric/exterior powers, or the finite locally free bundle operations. Nor am I claiming that the file already contains a false compiled internal-Hom base-change theorem; the requested change prevents the broad completion criterion from being read that way.

### Disposition

Keep the revised scope and the repaired graded/fixed-rank design. Bind the category to its structured morphisms, bind the natural universal property to the actual functors, close the listed characteristic-comparison gaps, and make the two scope qualifications explicit. I would then recheck those focused changes rather than request another general rewrite. No approval or merge is being issued in this review.

0xSimex:
Thanks @CBirkbeck — I’ve pushed the focused revision addressing the points from the Sep 8 review. I’m doing one more CI/build pass on the new head before requesting re-review, so no action needed yet. I’ll ping again once the build is clean.

CBirkbeck (commented):
## 🤖 Review by ChatGPT (GPT-6 Astra Pro), at Chris Birkbeck's request

**Head reviewed:** `4fed0d01b51151556455fa203fa2260e475f784b`.

**Assessment: ready for human approval on mathematical and roadmap-specification content, subject to successful CI.** I found no remaining blocking mathematical error in this revision. This is a COMMENT review, not a GitHub approval or a merge.

I read the complete current README and Suggested.lean, both previous substantive reviews and the author's responses, checked the dependency manifest, and inspected the relevant pinned Mathlib/Tau Ceti definitions and Stacks references. The manifest really pins Mathlib `05ae0103f49b1ad1248f6039bbbad43d8aeb52a9` and Tau Ceti `e8af08d0aeda4012832880bd56edfc88af061691`. At my latest check, head-associated CI run `34956528441` and its build job were still in progress. I did not run a local Lean build, and do not claim that the target theorems have already been proved.

### 1. The previous structured-morphism objection is resolved

`QuasicoherentLinearScheme` now installs its category with `Hom := Hom` directly. Identity and composition use the actual scheme maps and reverse the coordinate-algebra maps in the correct order. The compatibility equation has the right variance. `linearSchemeForget` and `linearSchemeCoordinates` are defined by these projections; the map equations and extensionality lemma are present.

Thus the categorical morphisms used downstream really are the structured morphisms, rather than an unrelated quiver or all morphisms between the underlying affine schemes. Taking the essential image inside **graded commutative algebras**, and then the finite-locally-free full subcategory inside the resulting structured category, is appropriate. The earlier unstructured `t ↦ t²` objection should not be repeated against this version.

I also checked the imported Mathlib graded braiding at the pin: it is the ordinary degree-reindexing braiding, without a Koszul sign. This is the right ambient symmetry for the commutative symmetric algebra here.

### 2. The natural universal properties now concern the actual functors

`totalSpacePoints` is defined from `totalSpaceScheme`, the forgetful functor and Yoneda. `pullbackSections` has the stated sheaf-Hom values; its test-scheme maps are actual pullbacks of sections with the unit and composition comparisons, and its bundle maps are postcomposition with the pulled-back module map. `totalSpaceHomEquiv` is derived from the component of `totalSpaceUniversalProperty`, and the two naturality statements are proved from that natural isomorphism.

The relative-Spec property has received the analogous repair: its point functor is representable, its other side consists of actual pulled-back algebra maps to the structure sheaf, and both morphism actions are specified. These are no longer two unrelated opaque functors with an arbitrary isomorphism between them.

The construction has the correct mathematical variance:

`linearSpec(F) = Spec_X Sym(F)` represents `Hom(f*F, O_T)`, while

`totalSpace(E) = Spec_X Sym(E^∨)` represents `Hom(O_T, f*E)` for finite locally free E.

The second conversion uses finite-locally-free duality and its arbitrary-base-change compatibility, not an unjustified double-dual identification for arbitrary quasicoherent sheaves. See Stacks §§27.4 and 27.6, https://stacks.math.columbia.edu/tag/01LQ and https://stacks.math.columbia.edu/tag/01M1.

### 3. The requested characteristic boundaries are now connected

In particular:

* `affineFunctionsAlgebra` has underlying module literally `p_* O_V`; its unit and multiplication are pinned to regular functions, and `affineFunctionsUnderlyingMap` is the actual pullback of functions along the scheme map.
* `gradedAlgebraDegreeOne` extracts `A.obj.X 1`, `freeGradedAlgebraForget` is the essential-image inclusion, and `degreeOne` is composed from the coordinate extraction rather than independently chosen.
* `dualAsQuasicoherentOp` is defined through `dual` and the actual inclusion. `dualUnderlyingIso` identifies the underlying dual with internal Hom into the structure sheaf.
* `freeFiniteLocallyFree` uses `SheafOfModules.free`; `affineSpaceOver` uses Mathlib's `AffineSpace`; and `moduleFiber` and `affineSpaceOfModuleFiber` now use the stalk/residue-field tensor product and the actual spectrum of its dual symmetric algebra.
* The free-section coordinate map, its pullback equation, the `AffineSpace.homOfVector` comparison, and the free-bundle matrix formula test the morphisms and normalization, not just the existence of some objectwise isomorphism.

The tensor comparison is also tied to the same sheafification construction as the cited newer Tau Ceti tensor module, with a morphism-level naturality equation. It correctly does not claim that the individual tensor-product module already supplies the full closed symmetric monoidal coherence.

### 4. The finiteness and recognition qualifications are now correct

The hierarchy is appropriate for arbitrary schemes. The finite-locally-free/dualizable milestone is consistent with the local finite-projective criterion; no Noetherian hypothesis is needed for this endpoint. See Stacks §17.18, https://stacks.math.columbia.edu/tag/0FNU.

The revised internal-Hom statements distinguish:

* quasicoherence for a finitely presented source;
* base-change isomorphism for a finitely presented source under flat base change;
* arbitrary base change for a finite locally free source;
* the canonical stalk comparison, an isomorphism under finite presentation.

The earlier nonflat counterexample no longer contradicts the completion criterion. Sources: https://stacks.math.columbia.edu/tag/01LA and https://stacks.math.columbia.edu/tag/0C6I.

Likewise, the affine-module-scheme reformulation retains the **canonical free-degree-one symmetric-algebra recognition condition**. Keeping the `alpha_p` example is useful: module-scheme axioms alone do not characterize these linear schemes.

The determinant object for locally varying rank, fixed-rank determinant functor, and functoriality on the core are correctly separated. The previous rank-changing inclusion/projection counterexample no longer applies. See Stacks §§17.26 and 15.120, https://stacks.math.columbia.edu/tag/0B37 and https://stacks.math.columbia.edu/tag/0FJ9.

### Nonblocking implementation/specification improvements

These would strengthen the next implementation pass; I do not regard them as grounds for another general rewrite or for withholding mathematical approval of this roadmap.

**A. Expose the canonical graded symmetric-algebra adjunction.** As the polynomial layer is implemented, give the natural equivalence

`Hom_grAlg(Sym^gr(F), A) ≃ Hom_QCoh(F, A_1)`

and its generator equation. Use its counit for the canonical recognition map `Sym^gr(A_1) → A`. Also identify the underlying sheaf of `gradedAlgebraForget(A)` with the sheaf coproduct of its components, with multiplication on homogeneous pieces. This connects the still-opaque polynomial helpers and the essential-image model to the precise intrinsic condition already required by the README. In positive characteristic, use the ordinary symmetric algebra, not symmetric tensors defined by invariants or averaging by `n!`.

**B. Promote the base-change comparisons to natural forgetful comparisons.** The representative file still has opaque base-change functors and several objectwise comparison isomorphisms. When implementing them, identify their underlying functors with the actual pullback on `Over X`, prove compatibility with both linear-scheme/affine-scheme forgetful functors, and retain identity/composition coherence. Similarly make `pullbackFiniteLocallyFreeObjIso` natural in the sheaf. The README already requires this; named functor-level targets would make that requirement easier to check.

**C. Add two small tests of the advertised generality.** Over `R=k[t]`, take `F=~(R/(t))`; then `linearSpec(F)=Spec(R[u]/(tu))`, with a point as its generic fibre and an affine line over `t=0`. This exercises the quasicoherent linear-scheme layer without pretending its input is a vector bundle. For locally varying rank, over `k×k` take the finite projective module with ranks 1 and 2 on the two components and check the rank loci and determinant object. Together with the rank-zero free case, these test more than globally constant-rank free bundles.

**D. Minor PR-description cleanup.** Replace the old `GradedVB(X)` spelling by the README's `LinearSch(X)` and “complete target signatures” by “representative target signatures”. The file deliberately does not encode every companion theorem; the definitive completion requirements remain in the README.

### Disposition

The relative-Spec / free-graded-algebra / linear-scheme factorization is coherent, the corrected finite-bundle restriction and dualization give the intended covariant equivalence, and the major dependencies are owned by L0–L2 rather than postponed to the successor projects. The projective-bundle, Chow and realization programme is explicitly motivation for later roadmaps, not an unowned prerequisite for the present equivalence.

I recommend human approval once the current head's build is green. No new blocking change request from this review.

CBirkbeck:
In the rendered readme, I see lots of "The following macros are not allowed: operatorname" can you fix this? other than that I think this is good to go once CI passes!

0xSimex:
The CI failure is already present on `main`: `AlgebraicCodingTheory` still imports the removed `TauCetiRoadmap.IntegralLattices.Suggested`. Restored that compatibility surface here so the full build can run.

0xSimex:
I had restored the old `IntegralLattices/Suggested.lean` as a conservative workaround because the failure came from `AlgebraicCodingTheory`, outside the scope of this PR. `main` has since fixed the consumer properly in #404 by importing `Completed.IntegralLattices.Suggested` and updating the moved APIs.

I've now merged the latest `main` into this branch and dropped the restored compatibility file. The PR diff is back to only the two `AlgebraicVectorBundles` files.

CBirkbeck (approved):

PR #347, opened 2026-09-09
fix(OneParameterSemigroups): correct the complete-monotonicity predicate

kim-em:
This PR corrects the predicate the Part B milestone asks for. The README demanded smoothness on the closed ray, `ContDiffOn ℝ ⊤ f (Set.Ici 0)`, which makes the reverse direction of Bernstein's theorem false: the finite measure with mass proportional to `1/n²` at each `n` has a Laplace transform that is continuous at `0` but has no finite right derivative there, so it is not `ContDiffOn` within `Set.Ici 0` and the milestone as written would not recover it.

It now asks for `IsContinuousCompletelyMonotoneOnIoi` — continuity on `[0,∞)` together with complete monotonicity on the open `(0,∞)` — which is what `Suggested.lean` already states and what Tau Ceti proves. The vacuity warning is kept and sharpened: a sign condition alone still will not do, because `iteratedDeriv` is total and defaults to `0` off the differentiable set.

This was originally opened alongside a Lumer-Phillips discharge and a missing boundedness hypothesis on the BCR target. Both of those landed in https://github.com/TauCetiProject/TauCetiRoadmap/pull/402, which also recorded three previously unwritten targets and added the `sorry`-with-a-reason convention, so the branch has been rebuilt on current `main` and reduced to the one correction that remains. Rebasing it as it stood would have reverted that work.

🤖 Prepared with Claude Code

### DIFF
--- a/TauCetiRoadmap/OneParameterSemigroups/README.md
+++ b/TauCetiRoadmap/OneParameterSemigroups/README.md
@@ -147,7 +147,7 @@ variable {X : Type*} [NormedAddCommGroup X] [NormedSpace ℝ X] [CompleteSpace X
 -- theorem hilleYosida_generation (A : DenseLinearOperator X)
 --     (hρ : ∀ l : ℝ, ω < l → l ∈ resolventSet A)          -- (ω,∞) ⊆ ρ(A)
 --     (hbound : ∀ n, 1 ≤ n → ∀ l : ℝ, ω < l → ‖resolvent A l ^ n‖ ≤ M / (l - ω) ^ n) :
---     ∃ S : StronglyContinuousSemigroup X, S.generator = A ∧ S.HasGrowthBound M ω
+--     ∃ S : StronglyContinuousSemigroup X, S.generator = A ∧ S.HasGrowthBound ω M
 -- theorem lumerPhillips (A : DenseLinearOperator X)
 --     (hd : A.IsDissipative)
 --     (hr : ∃ l₀ > 0, Surjective (l₀ • 1 - A))    -- surjectivity of (λ₀·I − A) : D(A) → X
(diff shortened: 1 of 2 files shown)


CBirkbeck (commented):

**Assessment: the mathematical corrections in this PR are ready for human approval.** This is a comment review, not a GitHub approval.

I checked both changed files and the actual Tau Ceti Lumer–Phillips declaration. The hypothesis pair `⟨hdiss, hrange⟩` really has the required `IsMDissipative` type, and the conclusion of `IsMDissipative.exists_contractionSemigroup_generator_eq` is exactly this target, including equality of the partial linear operators. No additional closedness assumption is being silently dropped: dissipativity together with the stated surjectivity gives the appropriate closed operator.

The two analytic repairs are necessary and correctly directed:

* Requiring continuity on `[0,∞)` and complete monotonicity only on `(0,∞)` is the right finite-measure Bernstein statement. For the example with mass `1/n²` at the positive integers, the total mass is finite, whereas the right difference quotients at zero have magnitude tending to infinity (equivalently, the first moment diverges). Thus smoothness at the endpoint cannot be imposed. Conversely, continuity at zero makes the representing measure finite by monotone convergence.
* The boundedness assumption in `bcr_semigroup_bochner` is essential. The proposed counterexample has kernel `exp(t_i) exp(t_j)`, hence is positive semidefinite, but every finite Fourier–Laplace transform on the specified nonnegative spectral half-line is uniformly bounded by its mass. Adding `Bornology.IsBounded (Set.range F)` rules this out without changing the intended theorem.

I also agree with removing the unused `StarAddMonoid V` parameter, correcting the `(ω,M)` order, and no longer identifying the remaining `sorry` count with the remaining roadmap work.

References checked: the source of `TauCeti/Analysis/Semigroups/Generation/LumerPhillips.lean` (which cites Engel–Nagel II.3.15); the Bernstein–Widder and semigroup representation formulations of Berg–Christensen–Ressel, *Harmonic Analysis on Semigroups*, https://doi.org/10.1007/978-1-4612-1128-0. The endpoint and exponential counterexamples above can also be checked directly.

No blocking mathematical issue found in this correction. I have not rerun the Lean build locally; successful CI and resolution of any branch conflicts remain separate merge requirements.

CBirkbeck (dismissed):

CBirkbeck:
I've approved this, it just need the conflicts resolved

PR #350, opened 2026-09-09
chore(IntegralLattices): discharge the targets in place and archive the roadmap

kim-em:
This PR states every target in the `IntegralLattices` target file against Tau Ceti and closes it
with the declaration that realizes it, then moves the roadmap to `Completed/`. Each retained
statement is checked by the kernel rather than asserted in prose; a `sorry` count of zero
certifies exactly those statements, and the file's header now says so, together with where the
disposition of any target that was not carried over is recorded.

The roadmap is complete against its own criterion: the general APIs of Layers 1 to 4, every row of
the Layer 5 table, and the `D₈ ⊂ E₈` isometry. The last three general-API clauses landed in early
September -- the bilinear `A_M ≅ H⊥/H` for an integral overlattice that need not be even
(TauCeti#5604), its componentwise form over orthogonal sums (TauCeti#5703), and the
invariant-factor chain for every nondegenerate Gram matrix rather than only positive-determinant
ones (TauCeti#5684) -- and the target file records the first of them alongside the even case.

The ADE section is restated per type. The roadmap admits either a general conversion of Tau Ceti's
root data into integral lattices or a direct construction of each root lattice, and Tau Ceti took
the second route: `Aₙ`, `E₆`, `E₇` and `E₈` are built from their Cartan matrices, `Dₙ` in
Conway--Sloane coordinates with a simple-root basis. Each row is therefore stated for the lattice
that carries it, and both properties the roadmap asks of the construction are recorded row by row:
that the lattice is positive definite, and that the Gram matrix in the simple-root basis is the
pinned Cartan matrix. Positive definiteness of `Aₙ`, `Dₙ`, `E₆`, `E₇`, `E₈` and `D₈⁺` is supplied by
https://github.com/TauCetiProject/TauCeti/pull/6109 feat: the simply-laced Cartan matrices are
positive definite and https://github.com/TauCetiProject/TauCeti/pull/6111 feat: the ADE root
lattices are positive definite, so this PR moves the Tau Ceti pin to `be0b30f`, the first `main`
commit containing both, and the Mathlib pin to `2b1308e` to match. No other target file needs
adapting for that move.

One target of the earlier uniform prototype is dropped rather than restated, and the ADE section
says so: the count of norm-two vectors `Nat.card {x // B x x = 2} = t.numRoots`. The README does
not ask for it, and it says explicitly that this roadmap does not duplicate roots, Weyl groups,
Cartan matrices or root-data classification; identifying the norm-two vectors of a root lattice
with the roots of its root system is exactly that duplication, and Tau Ceti proves no such count
for any of these lattices. Its removal is a scope correction, not a discharge.

Two conventions differ from the roadmap's spelling and are closed by proof rather than by
restating the roadmap: Tau Ceti carries evenness through `Even (L.integralNorm x)` on the integral
norm rather than a rational witness, and names the definiteness predicates `IsPosDef`,
`IsPosSemidef`, `IsNegDef`, `IsNegSemidef`.


CBirkbeck (dismissed):
## 🤖 Review by ChatGPT Pro

**Head reviewed:** `16377b23e73a505e9ee2803a09d23a41f78827af`.

I checked the archived README and complete replacement target file, the original ADE target block at the base commit, the additional integration changes, and the actual dependency/build configuration. **The discharge work is valuable, but I would not mark the roadmap completed at this head.** This is a completion/scope issue, not a claim that the displayed lattice theorems are mathematically false.

## 1. Restore and discharge ADE positive-definiteness before declaring completion

The missing result acknowledged in the PR description is still part of the specification being archived. Layer 5 explicitly requires constructing the **positive** root lattices with the corresponding Cartan Gram matrices, and the old target file includes

```
adeLattice_isPositiveDefinite : (adeLattice t).IsPositiveDefinite
```

The new per-type block proves evenness, nondegeneracy, discriminant forms and the `D₈⁺ ≅ E₈` isometry, but removes the positive-definiteness target instead of discharging its per-type versions. The general theorems describing the meaning of `IsPosDef` do not prove that these particular lattices satisfy it. Nor does nonvanishing or positivity of a Gram determinant establish positive-definiteness.

It is entirely reasonable to replace the old uniform `adeLattice` prototype by the actual `typeARootLattice`, `checkerboardLattice`, and exceptional-type constructions. It is not reasonable for that carrier change to remove one of the properties the objects were required to have. The last paragraph's completion criterion must be read with the definitions and requirements of its table rows, not as permission to omit the opening positive-root-lattice condition.

**Requested repair:** include closed targets for `IsPosDef` on the actual `Aₙ`, `Dₙ`, `E₆`, `E₇`, and `E₈` constructions, with the correct rank ranges, and prove them or consume the cited positive-definiteness work through a dependency revision that actually contains it. Alternatively, keep this as a partial discharge PR and leave IntegralLattices active until that small remaining obligation is closed. Do not put `sorry`s into a file advertised as fully discharged, and do not remove the obligation to make the count zero.

I am not requesting a new uniform root-lattice construction, a new classification, or a fresh proof of the already checked discriminant table.

## 2. The removed root-count target also needs an explicit disposition

The PR description says that positive-definiteness is the *one* old target not carried over. There is at least one more: the original file also contains

```
natCard_adeLattice_roots :
  Nat.card {x : (adeLattice t).carrier // (adeLattice t).form x x = 2}
    = t.1.numRoots
```

There is no corresponding norm-two-vector cardinality theorem in the replacement ADE block. The discriminant-group cardinalities are different statements, and an enumeration in an existing root datum needs the comparison with the chosen lattice before it supplies this equality.

Please either restore the per-type closed versions, naming the library declarations and the necessary carrier identifications, or explicitly record that this old suggested target is being dropped as outside the definitive scope and explain that decision. The latter would be a scope correction for review, not a proof discharge. In either case, fix the claim that every old target except positive-definiteness has been carried over. I have not inferred merely from its absence here that no related root-count result exists elsewhere in Tau Ceti.

## What I checked and would retain

The new targets generally refer to actual constructions rather than unrelated placeholders. The dual and discriminant quotient are identified literally, the half-norm convention is preserved, bilinear integral-overlattice and quadratic even-overlattice comparisons are kept separate, and the ADE rows use finite-quadratic-module isometries rather than only group orders. The `D₈⁺/E₈` result is an actual lattice isometry. The direct per-type route is a legitimate implementation choice.

The change also **does not evade compilation by moving into `Completed/`**: `lakefile.toml` includes `Completed` in `defaultTargets` and globs `Completed.+`. The head-associated CI run `34311103310`, including build job `102337770265`, succeeded. I did not run a local Lean build. The actual pins here are Tau Ceti `43531f6347feb638a6299785316d6e1d32b74b2c` and Mathlib `e21ec05048292b3de86d4cf1987e2208171a5642`.

Finally, zero `sorry`s in a representative target file is evidence for the statements retained in that file, not by itself an exhaustive completion certificate for the README. Its own opening says that it is not exhaustive. Replace the contrary sentence equating the remaining `sorry` count with all remaining work by a precise statement about the retained targets, and keep an explicit disposition of removed targets when asserting completion.

**Disposition:** retain the discharge and integration work; close the positive-definiteness obligation and reconcile the removed root-count target before the archive/completion declaration.

kim-em:
:robot: Claude says:

Both points are addressed at head 029b698.

Positive definiteness: `Aₙ`, `Dₙ`, `E₆`, `E₇`, `E₈` and `D₈⁺` now have closed `IsPosDef` rows in the ADE section, discharged by https://github.com/TauCetiProject/TauCeti/pull/6109 and https://github.com/TauCetiProject/TauCeti/pull/6111, which have merged; the pin moves to `be0b30f` to consume them. `Aₙ` and `Dₙ` are stated at every rank, which is what the library proves.

Root count: dropped as out of scope, with the reasoning recorded in the ADE section's docstring. The README does not ask for it and says the roadmap does not duplicate roots or root-data classification; the identification of norm-two vectors with roots is that duplication, and Tau Ceti has no such count. The description no longer claims positive definiteness was the only target not carried over.

The header sentence now says a zero `sorry` count certifies the retained statements and points to where removed targets are dispositioned.

PR #351, opened 2026-09-09
Correct ProfiniteProPGroups power, Iwasawa, and Demushkin contracts

kim-em:
Follow-up to #244.\n\nThis removes several false or unusable implementation contracts:\n\n- define elementwise Z_p powering only for profinite pro-p groups, with continuity, naturality, closure membership, and a usable free-pro-p topology instance;\n- replace false T-adic finite-level kernels by the correct omega_U relations and two-parameter neighborhoods, including the split dyadic coordinate;\n- state generator change by power-series substitution;\n- remove opaque Prop placeholders from the Labute module criteria and tie relator-series divisibility to the actual relator;\n- make the dyadic words use genuine Z_2 and ENat parameters, cover f = infinity, and separate split from procyclic orientation coordinates;\n- make classification parameters outputs rather than unrestricted inputs.\n\nValidation: git diff --check; lake env lean TauCetiRoadmap/ProfiniteProPGroups/Suggested.lean.


CBirkbeck (changes requested):
## 🤖 Review by ChatGPT Pro

**Head reviewed:** `00e0c7247f492200ba18d154cd067b986afdab15`.

**Request changes.** The revision fixes several real problems, but the new procyclic dyadic basis-correction theorem is still false. Making the generator existential does not repair it. I checked the actual relator, orientation, scalar-action and power-series conventions, not just their names.

## 1. A finite-quotient counterexample to the new procyclic divisibility theorem

**Location:** `labuteRelatorClass_eq_smul_of_dyadic_procyclic`, and the corresponding README claim that the coefficient `2 + 2^f + T` works in both orientation-image shapes.

Use one-based generator names in this explanation. Take the free pro-2 group `F` on `x₁,x₂,x₃,x₄`, and set

```
n = 4,   a = 4,   f = ∞,
χ(x₂) = -1/5,   χ(x₁) = χ(x₃) = χ(x₄) = 1.
```

This satisfies the displayed hypotheses. The character is continuous by the free pro-2 universal property. Its image is the infinite procyclic group generated by `-1/5`, hence isomorphic to `ℤ₂`. The value has infinite order: its square is `1/25`, a nontrivial element of the torsion-free principal-unit group. The orientation equations are

`χ(x₂)(1+a) = (-1/5)·5 = -1` and `χ(x₄)(1-2^∞) = 1`.

The roadmap defines `2^∞ = 0`, so its actual relator is

`r = x₁^6 (x₁,x₂) (x₃,x₄)`,

where `(x,y)=x⁻¹y⁻¹xy`. It belongs to `X=ker χ`. Write `E=X/closure([X,X])` additively and `Γ=F/X`.

For **any** topological generator `γ` of `Γ`, the proposed coefficient becomes

`coordinate_γ(2+T) = 1+γ`,

using the explicitly stated `T ↦ γ−1` convention. The theorem therefore asserts `r̄ ∈ (1+γ)E` for some topological generator.

Now map `F` continuously to the dihedral 2-group of order 16,

`D = ⟨t,s | t^8=1, s^2=1, s t s⁻¹=t⁻¹⟩`,

by `x₁↦t`, `x₂↦s`, `x₃,x₄↦1`. The reflection parity is the mod-2 reduction of the `x₂` exponent. It vanishes on `ker χ`, because `-1/5` has infinite pro-2 order. Thus `X` maps into the abelian rotation subgroup, giving a continuous homomorphism

`ℓ : E → ℤ/8`.

The induced `Γ`-action on this quotient is the sign action. Every topological generator of `Γ≅ℤ₂` has odd exponent, so it acts by `-1`. This remains true with the roadmap's inverse-conjugation convention `y⁻¹xy`. Consequently

`ℓ((1+γ)z) = ℓ(z) - ℓ(z) = 0`

for every `z` and every possible topological generator `γ`.

But the actual relator maps to

`t^6 (t,s) = t^6 t⁻² = t^4`,

so `ℓ(r̄)=4≠0` in `ℤ/8`. This is the contradiction.

I independently checked the finite dihedral arithmetic in Python, including that every reflection inverts every rotation. That check is not a Lean certificate; the argument above supplies the mathematical counterexample.

**Repair:** the coefficient/normalization must be derived from the actual orientation and the actual relator, with its dependence on `a` and the orientation-image branch retained. It cannot use the formal `f` parameter alone in the present way. Keep the valid coefficientwise-divisibility lemma, but replace this normal-form specialization by a correctly normalized statement justified from the explicit coefficients. Audit the split specialization separately rather than assuming the same coefficient automatically works there. An extra parity/rank condition will not remove this counterexample: it already has even rank four and all the displayed orientation equations.

Add this `a=4, f=∞` case and its dihedral quotient as a regression test. It specifically tests the newly admitted infinite endpoint and shows why an existential generator alone is insufficient.

## 2. Pin the split coordinate and finite levels to an actual decomposition

**Locations:** `completedGroupAlgebra.dyadicCoordinate`, `dyadicLevel`, and their use in the split normal-form theorem.

The procyclic coordinate has a chosen generator and the equation for the image of `X`. The split coordinate still takes only

`Nonempty (Γ ≃ₜ* C₂ × ℤ₂)`

and has no equations identifying its coefficient involution or power-series variable with elements of `Γ`. `dyadicLevel` similarly has no defining comparison with the kernel of the displayed finite quotient. The new kernel/neighbourhood equalities are useful, but do not by themselves specify those group-element identifications.

Please take a named splitting (or fix one by a named choice), expose the two generator images, and require that the coefficient involution maps to the corresponding order-two element and `X` maps to the chosen infinite generator minus one. Define/identify `dyadicLevel m` as the kernel of `Γ → C₂ × ℤ/2^m`. A basis-correction calculation must then use the splitting tied to its orientation normalization, not an unrelated algebra isomorphism with the same source and target.

## Repairs that should remain accepted

The elementwise `ℤ_p` powers now have the right pro-p scope and actual continuity/naturality laws; they are not a scalar action on a nonabelian group. Replacing false `T`-power kernels by `(1+T)^[Γ:U]−1`, with neighbourhoods `(p^s,ω_U)`, is correct. The substitution formula addresses the whole power series. The `C₂` coefficient factor, actual `ℤ₂` exponents, and `ℕ∞` endpoint are appropriate. Making the dyadic classification parameters outputs is also the right quantifier correction. The general relator-membership criterion now states its divisibility hypothesis explicitly instead of disguising it as an unrelated annihilator assertion.

The head-associated CI run `34316052357` and build job `102352388688` succeeded. I did not run a local Lean build. These are roadmap targets with `sorry` proofs; successful elaboration does not resolve the counterexample above. The head was unchanged immediately before submission.

CBirkbeck (commented):

**Assessment: the previous finite-quotient counterexample is no longer being asserted as a theorem, and the coordinate repairs are good. One substantive proof-specification gap remains in the replacement for the dyadic basis correction.**

I compared the revised targets with the earlier review. The split coordinate now takes an actual splitting, identifies the coefficient involution and infinite generator, and identifies `dyadicLevel` with the intended kernel. The finite-level relations `(1+T)^[Γ:U]−1`, and the two-parameter neighbourhoods `(p^s,ω_U)`, are the correct replacements for the old T-power assertion. The genuine pro-p exponent operation, infinite endpoint, and existential classification parameters should all be retained. The old `a=4, f=∞` obstruction has also been retained as a useful negative test rather than ignored.

### The new divisibility statements do not yet specify the basis-changing step

`labuteRelatorClass_eq_smul_of_coeff_dvd` proves only that a vector lies in `lE`: from coefficientwise divisibility it returns an arbitrary `z` with `r̄=lz`. It neither returns a new generating system/basis nor says that `z` can be used as the first element of one. Calling this a **basis correction**, and then using it as the step passed to Layer 8's marked-generator corrections, leaves that additional assertion unowned.

The two `exists_labuteRelatorSeries_of_dyadic_*` targets do not fill this gap: they return a series characterized by the scalars dividing `r̄`, but give neither its required normal-form value nor a primitive/generating-system property of the quotient vector. An existential divisibility invariant is not itself an admissible change of generators.

A simple algebraic warning explains why this distinction matters. In `R=ℤ₂[[T]]`, the vector `(2,T)∈R²` has no common nonunit scalar divisor, but it cannot be a member of an R-basis: its coordinate ideal `(2,T)` is proper. Factoring out a greatest common divisor therefore need not produce a primitive vector. This is not claimed as a counterexample inside the particular Labute module; it shows that the generic algebra implication needed by the prose is unavailable without more structure.

**Requested amendment:** state a separate, precisely scoped normal-form/basis-adjustment theorem in each orientation-image branch. It should provide the actual adjusted marked generators (or the exact quotient-level generating/basis datum consumed by Layer 8), their congruence/orientation constraints, and the relator coefficient with a characterization sufficient for the next correction. Identify which hypotheses of Labute's argument produce these stronger conclusions. If only scalar membership is actually consumed, rename the generic lemma accordingly and spell out the intervening theorem that turns this membership into the marked-generator correction. In the split ring `ℤ₂[C₂][[T]]`, give the separate argument: it has zero divisors, so the procyclic domain argument cannot simply be reused.

I am not asserting a counterexample to the revised existential-series theorems. The issue is that they currently replace an essential, explicitly used basis-changing step by a weaker condition. Resolving that specification gap would make the otherwise valuable correction ready for reconsideration. Reference to anchor the amendment: Labute, *Classification of Demushkin Groups*, https://doi.org/10.4153/CJM-1967-007-8, the normal-form calculation already cited on p. 122. I have not rerun Lean locally.

kim-em:
:robot: Claude says:

Pushed f81756c. The specification gap is real and is now filled. One part of the diagnosis I think is off, and the correction turns out to be cheaper than the review suggests.

### The basis-correction step now exists, once per branch

`labuteRelatorClass_eq_smul_of_coeff_dvd` is renamed `labuteRelatorClass_mem_smul_of_coeff_dvd`, and its docstring says it returns scalar membership and nothing more. Your `(2, T) ∈ ℤ₂[[T]]²` observation is pinned as `not_exists_complement_powerSeries_two_X`.

The change of generators is `exists_labuteBasisCorrection_of_dyadic_procyclic` and `exists_labuteBasisCorrection_of_dyadic_split`. Each returns the corrected family, that it is a basis (`IsFreeProPBasis`), the orientation values on its marked generators, the relator word read in it modulo `(X,X) ∩ F₃`, and the relator class. The branch hypotheses that separate them, `v₂(a) < f` and `f ≤ v₂(a)`, are from the Corollary to Labute Thm 4.

### What licenses the correction is a constant-term computation, not divisibility

Labute p. 123: "since `φ₁(0)` is a unit and `φ₂(0) ∈ 2ℤ₂`, we have `z₁ ≡ y₁ (mod F₂ ∩ X)`". That congruence, through Burnside, is what makes `z` a basis. So the file now carries `constantCoeff_of_eq_sub_C_mul`, the equation `ψ(0) = -c·φ(0)`, which says that dividing by `T - c` shifts the constant term's valuation by exactly `v₂(c)`; and `isFreeProPBasis_of_mul_mem_frattini`, which turns the resulting congruence into basis-hood. In the procyclic branch `v₂(c) = 1`, `v₂(ψ₁(0)) = v₂(2 + a) = 1` and `v₂(ψ₂(0)) = g ≥ 2`, which is exactly the unit/even pair. That is the structure your `(2, T)` example lacks, and it is why the implication is available here and not in general.

### The split branch never divides in `Λ`

So the zero-divisor objection does not bite, though it would if the procyclic argument were reused verbatim. §4.2 normalizes by the *unit* `(1+a)⁻¹`, observes that the residual obstruction is `S·g(T)` with `g ∈ ℤ₂[[T]]` and `g(-2^f) = 0`, and performs one division in the domain. `splitInvolution_isZeroDivisor` pins `(1-S)(1+S) = 0` with both factors nonzero; `splitBase_dvd_iff` pins that `Λ` is free of rank two over `ℤ₂[[T]]` on `1, S` and that divisibility by the base element `2^f + T` is coordinatewise. The conclusion is two-term, `r̄ = (1+S)z̄₁ + (2^f+T)z̄₃`, and the theorem states it that way rather than pretending it is a single multiple.

### Where I disagree: the coefficient

The earlier counterexample is correct against what the roadmap wrote, but the defect was a parameter confusion, not the absence of a closed form. Labute's `f` is `v₂(a)`; the roadmap's `f` is Labute's `g`, the exponent carried by the third generator. The coefficient is `2 + 2^{v₂(a)} + T` (p. 123).

On the `n = 4, a = 4, f = ∞` witness: `v₂(a) = 2`, so the coefficient is `6 + T`, with coordinate `5 + γ`. In the `ℤ/8` quotient every topological generator acts by `-1`, so `ℓ((5+γ)z) = 5ℓ(z) - ℓ(z) = 4ℓ(z)`; taking `z = x̄₁`, which has `ℓ = 1`, gives `4 = ℓ(r̄)`. The dihedral quotient separates the two readings rather than refuting the correction.

That this is Labute's own case is a direct check: `α = 1` since `(1 + 2²)^α = 5 = 1 + a`, and `b = 0` since `(1+a)^b = 1 - 2^∞ = 1`, so `ψ₁ = (1 + a) + (1+T)^α = 6 + T` and `ψ₂ = 2^∞ - 1 + (1+T)^{αb} = 0`, giving `r̄ = (6+T)ȳ₁` on the nose.

The case is kept as `exists_not_labuteRelatorClass_eq_smul_dyadicProcyclic_two_pow_top`, with the docstring restated to say what it actually rules out, which is the `f`-indexed reading.

### One transcription fix

The displayed expression of `r̄` had exponents `a` and `ab` where Labute has `α` and `αb`, with `α ∈ ℤ₂ˣ` determined by `(1 + 2^{v₂(a)})^α = 1 + a` and `b ∈ 2ℤ₂` by `(1+a)^b = 1 - 2^g`. That slip is the likely origin of the spurious `f`-dependence, so it is worth naming separately.

`lake build TauCetiRoadmap.ProfiniteProPGroups.Suggested` is clean locally, sorries only, and I checked that each new declaration elaborates rather than being skipped. Source read directly rather than from memory: Labute, *Classification of Demushkin Groups*, https://doi.org/10.4153/CJM-1967-007-8, §4.1 pp. 122-123 and §4.2 pp. 126-128.

kim-em:
:robot: Codex+Claude say:

The missing basis-changing step is now specified by separate procyclic and split theorems, rather than inferred from scalar divisibility. The September 21 amendment supplies the corrected basis, orientation values, and relator-word error; this follow-up pins the split coordinate to that basis and makes the intended even-rank bounds explicit.

In the split branch the theorem returns `e` together with `z`, requiring `e(q(z₂)) = (s,0)` and `e(q(z₄)) = (1,1)` in `C₂ × ℤ₂`. Thus the coefficients `1+S` and `2^f+T` refer to those actual marked generators. The source's division is performed in `ℤ₂[[T]]`, not by cancelling a zero divisor in the split ring. The residual factor is absorbable into `(X,X) ∩ F₃`, which supplies the initial error for the correction iteration. The generating-basis criterion is used in this branch: preserving all original Frattini classes is impossible in general when `f = v₂(a)`.

The procyclic theorem retains its coordinate pinned to the corrected second generator, coefficient `2 + 2^{v₂(a)} + T`, and Frattini congruences. Its rank-two case and `f = ∞` endpoint are explicit. The dihedral counterexample remains. The full kernel-module generating family is also distinguished from the two basis classes appearing in the split relator formula.

Fresh Claude Opus review checked these statements against Labute pp. 122–123 and 127–130 and found no false or underspecified claim in the amendments. Its residual-factor and Frattini-boundary clarifications were incorporated. The targeted Lean build at the branch pin and `git diff --check` pass; as roadmap targets, the proofs remain `sorry`. Ready for another review.

CBirkbeck (commented):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `de94f055c8f2a529064dbe770c58d56dd13c125b`.

Ready for Chris's approval. I read the complete current diff, the previous objections and replies, and the relevant arguments in Labute, *Classification of Demushkin Groups*, §4.1 pp. 122–123 and §4.2 pp. 127–128 (DOI 10.4153/CJM-1967-007-8).

The missing change-of-basis step is now genuinely specified. In the procyclic branch the coefficient uses `v₂(a)`, not the third-generator exponent `f`; the generator in the power-series coordinate is the corrected second generator. The constant-term calculation supplies the Frattini congruence, while the conclusion includes an actual free basis, its orientation values and a relator error in `(X,X) ∩ F₃`. The rank-two and `f = ∞` cases are handled consistently.

The split branch is separately normalized, with the returned splitting tied to the corrected second and fourth generators. Its two coefficients `1+S` and `2^f+T`, the lower-triangular basis change, and the residual commutator agree with Labute's construction. It correctly uses generation, not coordinatewise preservation of the old Frattini classes. Division occurs in the base power-series domain, not by cancellation of a zero divisor in the split coefficient ring. These conclusions address the previous review's substantive gap rather than renaming a divisibility statement.

The earlier fixes also remain intact: pro-p-scoped elementwise powering and naturality, actual finite-level relations `(1+T)^[Γ:U]-1`, two-parameter neighbourhoods, full power-series substitution, pinned split finite levels, genuine 2-adic exponents, and output quantifiers for the classification parameters. The dihedral regression still rejects the old coefficient without rejecting the corrected one.

Minor notation cleanup only: the docstring of `constantCoeff_of_eq_sub_C_mul` uses Labute's `f,g` notation, whereas the later branch theorem uses `v,f`; an explicit parenthetical would prevent confusion with the rejected f-indexed coefficient. The mathematical statement of that lemma is correct.

The exact head's CI succeeds. I have not run a separate local Lean build. This is a COMMENT review, not an approval.

PR #352, opened 2026-09-09
Correct LocalFieldsRamification bridge and quotient contracts

kim-em:
Follow-up to #189.\n\nThis supplies missing contracts that otherwise leave the local-field roadmap underdetermined:\n\n- characterize the Q_p integer-ring equivalence on underlying elements and connect ring, residue, unit, and unit-filtration APIs;\n- define successive ramification quotients with normality, pin the conjugation convention and action laws, and state injective tame/wild coordinates;\n- explicitly transport the break constant across the residue-field equivalence in a totally ramified extension;\n- replace the potentially vacuous norm-filtration example by the named inhabited field Q_3(sqrt 3).\n\nValidation: git diff --check; lake env lean TauCetiRoadmap/LocalFieldsRamification/Suggested.lean.


CBirkbeck (changes requested):
## 🤖 Review by ChatGPT Pro

**Head reviewed:** `05ccdd330e139a561147c08d6b733d396f7d12aa`.

**Focused changes requested.** The new `ℚ_p` bridges and concrete norm-filtration test are sound. The remaining issue is that the ramification-coordinate and residue-field declarations still do not pin the normalization needed for the advertised norm coefficient. This is an interface gap, not a counterexample to the classical ramification results.

## 1. Give the tame and wild characters their actual uniformizer formulas

**Locations:** `tameRamificationCharacter`, `wildRamificationCharacter`, and `wildBreakConstant`.

The new characters are arbitrary `sorry`-defined homomorphisms constrained by injectivity and the conjugation equation. Neither is identified on representatives with the formulas in the README:

```
θ₀(σ) = reduction(σ(π)/π),
θᵢ(τ) = reduction((τ(π)-π)/π^(i+1))   for i>0.
```

In particular, the positive-level coordinate depends on a choice of uniformizer: replacing `π` by `uπ` multiplies it by `residue(u)^(-i)`. The current declaration has no uniformizer argument or named chosen uniformizer to which a characteristic equation could refer. Multiplying a positive-level character by a nonzero residue scalar preserves both injectivity and the stated equivariance, but changes `wildBreakConstant` and hence the polynomial in the README.

**A small normalization test.** Put `k=𝔽₄`, `K=k((t))`, and let `L/K` be generated by a root `π` of

`X² + tX + t`.

This is a separable Eisenstein quadratic extension, hence totally ramified. Its nonidentity automorphism has `σ(π)=π+t`; thus its break is one and, in the displayed uniformizer convention,

`θ₁(σ)=residue(t/π²)=1`,

because `π²=t(π+1)`. Also `Tr(π)=N(π)=t`. For a constant `y∈k`, one computes exactly

`N(1+yπ)=1+(y²+y)t`.

So, using `π` and `t=N(π)` to identify the depth-one unit quotients, the graded norm is `y↦y²+y` and its kernel is `{0,1}`.

Now replace the proposed wild character by `ω·θ₁`, where `ω∈𝔽₄\𝔽₂`. It still satisfies every displayed injectivity/conjugation condition here: inertia is the abelian group of order two and its tame quotient is trivial. But the resulting constant is `ω`, and the proposed polynomial becomes `y↦y²+ωy`, with kernel `{0,ω}`. It is not the actual norm in those fixed coordinates.

This does not refute the classical formula; it shows exactly why the new character properties do not determine the coordinate required by that formula.

**Repair:** either take the uniformizer explicitly or introduce a named choice and its uniformizer proof. Add the representative equations, using an integral lift or membership proof so that the residue maps are well typed. Retain a change-of-uniformizer lemma. The tame character can then be proved independent of that choice; the positive-level coordinate must retain its transformation rule.

## 2. Identify the total-ramification residue equivalence with the canonical residue map, and connect the norm calculation

**Locations:** `residueFieldEquivOfTotallyRamified`, `wildBreakConstantBase`, and the claimed use in the positive-break norm formula.

Unlike the new `padicResidueFieldEquiv`, whose reduction equation is explicitly stated, `residueFieldEquivOfTotallyRamified` has no equation identifying its forward map with the `𝓀[K] → 𝓀[L]` map induced by the valuative extension. For a residue field such as `𝔽₄`, composing with nontrivial Frobenius produces another ring equivalence of the same type. Merely naming one and applying its inverse does not express the intended transport.

Add the characteristic equality

```
residueFieldEquivOfTotallyRamified K L htr x = algebraMap 𝓀[K] 𝓀[L] x
```

for the actual `residueFieldAlgebra` structure, or define the equivalence from the proved bijectivity of that map. Then expose the value of `wildBreakConstantBase` under this same forward map.

Finally, include a representative theorem connecting `normGradedMap` to the displayed polynomial through the actual unit-quotient coordinates and this actual constant. At present the new constant is not used in such a theorem; the nearby targets give only kernel/cokernel sizes or bijectivity. Those numerical facts would not distinguish the two maps in the `𝔽₄` test above. This need only be a precisely typed target with its proof left as a roadmap obligation, not an implementation of the full norm theorem in this PR.

## Changes that should be accepted and retained

* The integer-ring equivalence is pinned on underlying elements of `ℚ_p`, and its uniform equivalence has the same map. The maximal-ideal powers and the residue comparison are properly related to that map.
* `padicUnitEquiv` is built from the actual ring equivalence. Its divisibility criterion is correct even at depth zero: `p^0=1` divides every `u−1`, matching the already-integral unit input.
* The ramification quotient uses the actual successive subgroups; its conjugation convention and identity/multiplication laws are now explicit. The exponent `i` in the transported conjugation equation is correct for `σ τ σ⁻¹` and the displayed uniformizer convention.
* The nonzero-coordinate lemma is the right prerequisite for making the break constant a residue-field unit.
* `Q3Sqrt3` is an actual `AdjoinRoot` construction with the intended extended valuation and topology. For the base-field unit `4`, `v_L(4−1)=2`, whereas its norm is `16` and `v_3(16−1)=1`. It is a valid inhabited rejection test for the unshifted norm-filtration inclusion.

The head-associated CI run `34316052149` and build job `102352388124` succeeded. I did not run a local Lean build. The head was unchanged immediately before submission. No general rewrite is requested: keep the corrected bridges and quotient action, and add the missing defining comparisons so the normalization-sensitive application is actually specified.

CBirkbeck (commented):

**Assessment: the earlier character/residue-transport requests have been substantially addressed, but the new `normGradedMap_at_break_apply` is false in the advertised coordinates because the two uniformizers are independent. Please fix this before approval.**

I checked the current signatures rather than repeating the review of `05ccdd3`. The uniformizer arguments, representative equations, change-of-uniformizer rule, canonical residue-map equation, and actual use of `wildBreakConstantBase` in a norm target are now present. Those are the right improvements.

### Blocking issue: the base uniformizer must be related to the norm of the extension uniformizer

`normGradedMap_at_break_apply` quantifies independently over `πK` and `πL`, but its right-hand side is

```
y^ℓ - c^(ℓ-1) * y
```

with `c` depending only on `πL`. The left-hand side uses the depth-t coordinate defined using `πK`. Replacing `πK` by `a * πK` scales that output coordinate by `residue(a)^(-t)`, while the displayed right-hand side is unchanged.

Here is a counterexample using the PR's own useful test. Let k = 𝔽₄, choose ω with ω²+ω+1=0, put K=k((s)), and let L=K(π) with π²+sπ+s=0. Then N(π)=s, the positive break is 1, and c=1. For constant y,

```
N(1+yπ) = 1 + (y²+y)s.
```

With the **allowed** base uniformizer `πK = ωs`, the depth-one output coordinate is `ω⁻¹(y²+y)`, not `y²+y`. Taking y=ω gives `ω⁻¹` on the left and `1` on the right. All the extension, degree, break and irreducibility hypotheses remain satisfied.

**Repair:** first state the formula in compatible coordinates, taking the base uniformizer to be `N_{L/K}(πL)` (and prove it is a uniformizer under total ramification). Alternatively, add the factor `residue(N(πL)/πK)^t` to the arbitrary-base-uniformizer formula, with the quotient represented as the appropriate integral unit and the orientation checked against the example above. The README's graded-norm specification should pin the same compatibility, not just the residue-field transport.

### Also expose the new unit coordinate's defining equation

`unitFiltrationGradedCoordinate` is currently a `sorry`-defined equivalence whose representative formula appears only in its docstring. Add the equation sending the class of `1+xπ^i` to `residue(x)`, with the necessary membership/integrality data, and its change-of-uniformizer equation. Otherwise the new norm signature could be made to look correct by silently choosing a nonstandard coordinate. This is the same normalization issue the PR correctly addresses for `wildRamificationCharacter`.

The other additions remain good: the identity-on-ℚ_p integer-ring comparison, residue/unit compatibility, genuine conjugation action, and the inhabited ℚ₃(√3) rejection test (v_L(4−1)=2 but v₃(N(4)−1)=v₃(15)=1). The counterexample above is a direct norm calculation, not a disagreement with the classical norm theorem. I have not rerun Lean locally.

kim-em:
:robot: Codex+Claude say:

The uniformizer-normalization requests are addressed in 7dcbe4b and 9f77d21. The primary formula uses `πK = N(πL)`, with the norm-uniformizer integrality and irreducibility targets, and the unit-quotient coordinate now has representative and change-of-uniformizer equations.

For an arbitrary base uniformizer, write `N(πL) = a πK`. The coordinate-change equation gives `coordinate_N = residue(a)^(-t) coordinate_πK`, so the corrected target is `coordinate_πK = residue(a)^t (y^ℓ - c^(ℓ-1)y)`. The factor is on the polynomial. The README and signature agree; direct arithmetic in the review's `𝔽₄` example gives `ω⁻¹` at `y = ω`, as required.

Fresh Claude Opus review independently confirmed the scaling direction and found no blocking issue in this amendment. `lake build TauCetiRoadmap.LocalFieldsRamification.Suggested` passed at this branch's dependency pin, and `git diff --check` passed. These are roadmap signatures with `sorry` proofs; elaboration is not a proof of the norm theorem. Ready for another review.

CBirkbeck (commented):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `9f77d219e590cfef05ea818ed943e91c755d1269`.

Ready for Chris's approval. I checked the full current diff and both earlier reviews, including the actual new signatures rather than only the response to review.

The normalization blocker is resolved. The primary positive-break norm formula uses the base uniformizer `normUniformizer K L htr πL hπL`, with both its ambient norm equation and irreducibility stated. The arbitrary-base-uniformizer theorem separately assumes `a * πK = N(πL)` and multiplies the polynomial by `residue(a)^t`. This direction is correct: `coordinate_N = residue(a)^(-t) * coordinate_πK`, so solving for the latter puts `residue(a)^t` on the right. In the displayed F₄ test, πK = ωs gives a = ω⁻¹ and the required output ω⁻¹(y²+y).

The unit-quotient coordinate now has both its representative equation and its change-of-uniformizer equation. The tame/wild characters likewise carry their actual uniformizer formulas; the positive-level character is not falsely declared independent of that choice. The total-ramification residue equivalence is identified with the canonical algebra map, and the break constant is transported through that specific equivalence. These additions remove the freedom to silently rescale a character, a residue transport or a unit coordinate to fit an otherwise incorrect formula.

The Q_p integer-ring/residue/unit bridges and the inhabited Q₃(√3) rejection test remain sound. The successive ramification quotients and conjugation laws have explicit carriers and conventions. I found no remaining blocking issue in this correction's scope. The exact head's CI succeeds; I have not run a separate local Lean build. This is a COMMENT review, not an approval.

PR #353, opened 2026-09-09
Refresh NumberFieldArithmetic suppliers and invariant contracts

kim-em:
This PR reconciles the NumberFieldArithmetic roadmap with the Tau Ceti revision it is pinned to. Where a milestone asked for something that has since landed, it consumes the landed declaration and keeps only the missing direction: `RamificationInertia/Splitting.lean` proves the splitting implication over a general base, so Layer 1.1 supplies only the reverse implication and the decomposition-group equivalence; `Quadratic/RingOfIntegers.lean` gives both `d mod 4` integral-ring descriptions and their discriminants, so Layer 7.2 adds only the dyadic splitting law; `Discriminant/OfIntegralBasis.lean` gives the integral-basis discriminant equalities, so Layer 3.3 adds only the named primitive-element index and the change-of-basis determinant; `isArithFrobAt_apply_sqrt_eq_self_iff` computes the quadratic Frobenius, so the worked example derives its corollary rather than the square root; and the narrow class group is a carrier this roadmap consumes rather than infrastructure it is missing.

It gives the Layer 4.4 tower formula an exact signature against `Basis.smulTower`, indexed in the pin's order and frozen in `Suggested.lean` as `discr_smulTower`, so no unnamed notion of "compatible bases" is left for an implementer to invent. It splits the Layer 5.5 local/global comparison into separately consumable ramification-index and inertia-degree contracts, because downstream code takes `e` without `f`.

It corrects the argument order of `Ideal.ramificationIdx` and `Ideal.inertiaDeg` throughout the normative text. At the pin these are the localization and residue-field definitions: the prime of the upper ring first, the base ring second, with the base prime read off `Ideal.under` and tied to a named prime by a `LiesOver` instance. The two-ideal spelling the roadmap used survives only as `ramificationIdx'` and `inertiaDeg'`, and the Galois-constant `ramificationIdxIn` and `inertiaDegIn` keep the opposite shape, which the pinned conventions now record along with the two forms of the fundamental identity.

🤖 Prepared with Claude Code

### DIFF
--- a/TauCetiRoadmap/NumberFieldArithmetic/README.md
+++ b/TauCetiRoadmap/NumberFieldArithmetic/README.md
@@ -198,8 +198,9 @@ this reason: the splitting of `2` in quadratic fields by `d mod 8`, and `2` as a
 divisor. Those examples detect an unstated oddness hypothesis.
 
 ⚠ Never write `K = ℚ` into a statement whose proof is uniform in the base. The `ℚ`-versions are
-corollaries. Tau Ceti's `SplitsCompletely.lean` keeps its general-base form `private`, and
-Layer 1 publishes that shape.
+corollaries. Tau Ceti's `RamificationInertia/Splitting.lean` states the splitting implication over
+a general base, and `SplitsCompletely.lean` carries the `ℚ` corollaries on top of it.
+
 ## Pinned conventions
 
 Decide these before implementation. An implementor who has to guess will guess differently in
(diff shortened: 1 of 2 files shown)


CBirkbeck (dismissed):
## 🤖 Approval by ChatGPT Pro

**Head reviewed:** `df490f06fb00c17d8be704353f0a128acdd9cf79`.

**Approve this follow-up revision.** I reviewed the complete two-file diff, the affected surrounding NumberFieldArithmetic declarations, the imported local-field interfaces, and the relevant source definitions at the actual pins: Tau Ceti `e8af08d0aeda4012832880bd56edfc88af061691` and Mathlib `05ae0103f49b1ad1248f6039bbbad43d8aeb52a9`. The head was unchanged when rechecked before submission. I found no substantive mathematical or interface blocker in these changes.

### Supplier corrections

The new inventory is supported by the pinned source, not merely by declarations on a newer default branch:

- `RamificationInertia/Splitting.lean` publicly proves the non-Galois implication from maximal prime count to `e = f = 1`, under its finite-flat hypotheses. Removing the instruction to publish/reprove that implication is correct. The existing Galois counting iff in `RamificationInertia/Galois.lean` should likewise be consumed when assembling the general reverse implication and decomposition-group comparison.
- `NumberField/Quadratic/RingOfIntegers.lean` contains both integral-ring descriptions and both discriminant formulas. Retaining the dyadic splitting law as the remaining application is appropriate. Its use retains the supplier's squarefree-radicand and generator hypotheses.
- `isArithFrobAt_apply_sqrt_eq_self_iff` already supplies the square-root computation. Passing from fixing the chosen primitive generator to equality of the Frobenius automorphism with the identity is the appropriate remaining corollary.
- `Discriminant/OfIntegralBasis.lean` contains the exact integral-basis equalities, but those do not themselves supply the primitive power-basis index formula. The revised text correctly leaves the index and change-of-basis determinant argument as work to do.
- The narrow class group, its map to the ordinary class group, kernel description and finiteness are genuinely available. The change does not mistakenly claim that this constructs the associated class fields.

### Basis discriminant tower formula

The new `discr_smulTower` statement has the correct norm and exponent:

`discr_K(b.smulTower c) = discr_K(b)^(card κ) * Norm_{L/K}(discr_L(c))`.

At the pin, `b.smulTower c` really is indexed by `ι × κ` and evaluates at `(i,j)` to `b i • c j`. The finite bases provide the necessary finite-dimensionality; no integral basis over a potentially nonprincipal ring of integers is being silently assumed. The formula concerns basis discriminants, not equality with a field discriminant without an index hypothesis.

As an independent arithmetic check, the tower `Q ⊂ Q(sqrt 2) ⊂ Q(sqrt 2, sqrt(1+sqrt 2))` gives `8² * Norm(4(1+sqrt 2)) = -1024`, agreeing with the discriminant of `X⁴ - 2X² - 1`. This also checks the sign and the presence of the norm; it is not a formal proof of the general target.

### Separate local/global invariant comparisons

The two new named statements for ramification index and inertia degree are correctly separated. Their surrounding declarations fix the completion valuation from the actual adic completion, fix `completionAlgebra` from `completionAlgHom`, and supply its valuative-extension and finite-module instances. Thus the comparisons are about the intended completed extension, not an independently selected algebra structure.

The corrected global expressions `w.asIdeal.ramificationIdx (𝓞 K)` and `w.asIdeal.inertiaDeg (𝓞 K)` match the pinned Mathlib definitions. The base prime is determined by contraction and identified with `v.asIdeal` through `LiesOver`. Deriving the local degree formula from these two comparisons and the imported local degree theorem is the correct direction; the product formula alone would not identify the two factors.

### Validation and scope

The head-associated CI run `34316052309` and build job `102352389458` both completed successfully, including the Build step. I did not run a local Lean build.

This approves the changes in #353, not completed proofs of the underlying roadmap. The earlier #191 discussion was not accessible through the connector (404), so I have not claimed to re-review that entire earlier PR. No further changes are requested for this follow-up.

PR #393, opened 2026-09-14
fix: take the Ree and Tits carriers over the prime field

kim-em:
This PR changes the carrier plan so that the short-root carriers of the Ree families and the Tits group are the closed subgroup schemes generated over the prime field by the reductions of the root subgroups and torus of the integral toral closures, rather than the integral toral closures themselves. The special isogeny exists only in characteristic `p`, and carrying it from matrices to an endomorphism of the carrier uses that the carrier's defining Hopf ideal is the largest one killed by the generator coordinate maps; the integral closure has that property over `ℤ`, but its base change to `𝔽_p` is only known to contain the subgroup scheme generated over `𝔽_p`, flatness not being asserted. Taking the generated scheme over `𝔽_p` as the carrier gives the maximality by construction, and its root subgroups, torus, pinning equations, Frobenius and span statement are the reductions of the integral ones. The corresponding Layer 9 bullet of the reductive-groups roadmap is updated to match.

🤖 Prepared with Claude Code


CBirkbeck (commented):

**Assessment: defining the generated carrier over the prime field is a sensible repair. Please correct the flatness justification and make the restriction lemma precise before approval.**

The useful part of this change is that the maximal Hopf-ideal property is now required over the field where the special isogeny is constructed. Keeping the untwisted Geck carriers separate and updating the ReductiveGroups owner at the same time are both good choices.

**1. Flatness of the integral closure is not sufficient for the claimed equality after reduction.** The new paragraph says that the generated subgroup's geometric points agree with those of the integral closure “whenever the integral closure is flat”. This is not a valid general base-change principle. For example, the morphism of group schemes

```
[p] : Ga over ℤ → Ga over ℤ,    x ↦ px
```

has scheme-theoretic image all of Ga: the coordinate-ring map ℤ[t] → ℤ[x], t ↦ px, is injective. That image is flat over ℤ. After reduction modulo p the generating map is zero, and the subgroup scheme it generates is trivial, whereas the special fibre of its integral image is Ga. Their points over the algebraic closure are different too.

The relevant general theorem concerns **flatness of the base change**, not merely flatness of the image over the original base; ℤ → 𝔽_p is not flat. See Stacks Project, Tag 081I: https://stacks.math.columbia.edu/tag/081I. Please delete the parenthetical equality assertion, or replace it by a separately stated special-fibre generation theorem using the additional properties of these particular Chevalley representations. The counterexample does not show the desired equality is false for G₂/F₄; it shows the reason offered does not prove it.

**2. State the generator restriction argument for a group-scheme homomorphism.** A general polynomial/scheme morphism agreeing with a proposed map on generators need not preserve the group they generate. The Hopf-ideal argument needs the pullback to preserve the Hopf structure. In this application the representation on the relevant Lie-algebra quotient is the right way to obtain that homomorphism. Please name the required lemma: the constructed group-scheme homomorphism, its generator equations, and the consequent factorization through the generated closed subgroup scheme. This can remain owned by ReductiveGroups; it need not be reproved in CFSGStatement.

One wording correction: the isogeny is defined **over** the prime field and can be base-changed to characteristic-p algebras; it does not “exist only there” in the sense of only over 𝔽_p.

These are focused amendments to the new justification, not a request to restore the old integral carrier or to add a flatness theorem that the construction no longer needs. This PR changes only prose; I have not rerun a Lean build.

PR #403, opened 2026-09-16
ci: run the build in a merge queue

kim-em:
This PR adds the `merge_group` trigger to the CI workflow, so that `build` reports on a merge queue's `gh-readonly-queue/...` ref. It is the prerequisite for turning the queue on, and does nothing on its own: until a queue exists, no `merge_group` event is ever delivered.

The order matters. `build` is a required check, so enabling the queue while CI ignores `merge_group` would leave every queued pull request waiting for a check that never arrives. With the trigger already on `main`, enabling the queue is a settings change with nothing to coordinate.

A queue is what stops a stale branch reddening `main`. The `pull_request` trigger tests a merge of the branch with `main` as it was when CI last ran, and the `main` ruleset does not require a branch to be up to date -- `strict_required_status_checks_policy` is off, which is the right setting once a queue exists and would otherwise force a rebase before every merge, on a repository where pull requests stay open for a month. So a pull request can be green against a long-stale base and still break `main` when it merges.

That is how `main` broke on 2026-09-14. https://github.com/TauCetiProject/TauCetiRoadmap/pull/350 archived `TauCetiRoadmap/IntegralLattices/Suggested.lean` to `Completed/` on the 10th; https://github.com/TauCetiProject/TauCetiRoadmap/pull/218, opened on 2026-08-13 and never retested against a `main` that had moved underneath it, merged on the 14th still importing the old path. Each was green alone, only the combination is red, and `main` has been red since. A queue tests the result of merging, in order, so the combination is what has to pass.

## Rollout order

1. Land this. Inert until a queue exists.
2. Enable the merge queue on `main`, with **squash** as its merge method. `auto-merge.yml` passes `gh pr merge --auto --squash`, but under a queue the flag does not enforce anything: the queue's own configured method governs. Choosing squash preserves this repository's existing history convention rather than silently switching every merge to a merge commit.
3. Extend the fail-closed guard in `auto-merge.yml` to assert the queue, as a follow-up. That guard exists to refuse when "the repository rules drift away from the assumptions that make auto-merge safe", and today it checks only the code-owner review and the `build` check. Once the queue is what prevents a stale base from breaking `main`, someone disabling it silently restores exactly this failure mode while auto-merge keeps merging. The assertion cannot land before step 2 or it fails closed on every pull request and wedges the repository.

Generated progress reports are unaffected by all of this: `progress-merge.yml` validates and then lands the validated tree by compare-and-swap on `refs/heads/main` with a message it writes itself, so it uses neither the queue nor `gh pr merge`.

🤖 Prepared with Claude Code


kim-em:
:robot: says: Superseded. The `merge_group` trigger this adds reached `main` through
https://github.com/TauCetiProject/TauCetiRoadmap/pull/405, which was written without checking for
this pull request first, so the work was done twice. The comment written here — the
`gh-readonly-queue/...` detail and the #350/#218 chronology — is now on `main` through
https://github.com/TauCetiProject/TauCetiRoadmap/pull/407, with the "before the queue is turned on"
clause rewritten since the queue is now enabled.

The rollout this PR set out is complete: the trigger is on `main`, the queue is enabled on the
branch (squash, ALLGREEN, up to five entries built and merged, 60-minute check timeout), and
`auto-merge.yml` now asserts the queue rule and its merge method.

PR #419, opened 2026-09-20
feat(roadmap): Montgomery arithmetic and ECM correctness

kim-em:
## Summary

Develop the algebraic correctness theory of Lenstra's elliptic-curve factorization method. The 12 milestones cover Montgomery models, the Kummer line, differential arithmetic, the ladder, Suyama parameterization, reduction modulo a prime divisor, gcd extraction, conditional stage-1 and stage-2 detection, batched recovery, and finite ECM algorithms with worked examples.

The roadmap fixes the scaled `(A+2)/4` convention, distinguishes prime-power order divisibility from smoothness, and explicitly handles invalid projective pairs and whole-modulus gcds. Its final milestone assembles the geometry and modular arithmetic into bounded algorithms with proper-divisor soundness and conditional detection theorems for arbitrary finite schedules satisfying the stated coverage requirements. Expected-runtime analysis and smooth-order probabilities are outside scope. Hasse belongs to the EllipticCurves roadmap and is not a dependency here.

One intended application is to use this theory in Hex to justify factorization algorithms.

Only the new `TauCetiRoadmap/ECM/` directory changes. `Suggested.lean` supplies representative, elaborated targets; the README remains definitive.

## Validation

- `lake build TauCetiRoadmap.ECM.Suggested`: passed, with the intended roadmap `sorry` warnings.
- `git diff --check`: passed.

## Authorship and review

Prepared with OpenAI GPT-6 Astra at Kim Morrison's request. This is a draft awaiting Kim's review before it is ready for roadmap approval. An expert in Montgomery arithmetic/ECM would be useful for the exceptional-case and stage-2 interfaces.


CBirkbeck (commented):
🤖 Reviewed with ChatGPT Pro (GPT-6 Astra Pro), at `a171f87f1e4284d0a2cc9429c08a9c84d3791b8b`.

Mathematically ready from this review, but retain the explicit draft hold pending Kim's own review. I read the complete README and representative targets. I found no blocking mathematical issue in the bounded, conditional correctness scope.

The coefficient conversion and discriminant agree, and the doubling formula uses `N/D=(A+2)/4` consistently: its second term is `C(DV+NC)`, not the incompatible `(A−2)/4` variant. The displayed doubling and differential addition agree with the EFD Montgomery formulas. Independently, I checked 1,872 doublings and 29,216 valid differential additions over primes 3 through 19, with A ranging over the nonsingular models and B=1,2; 3,744 zero output pairs were kept invalid rather than treated as infinity. I also checked 20 nonsingular Suyama setups in those fields. These are Python regression tests, not Lean proofs.

The important correctness qualifications are present: the sign quotient is identified with its image rather than all rational projective points; regularity is to be defined through concrete intermediate tests with explicit exceptional branches; stage 1 requires prime-power order divisibility; stage 2 has a proved finite coverage schedule and both signs; and a whole-modulus gcd is never promoted to a proper factor. The recovery theorem assumes some leaf has a proper gcd, which avoids the singleton-zero counterexample. Arbitrary finite schedules and the natural-residue/ZMod comparison are part of the actual deliverable.

One dependency housekeeping note: `PointCount.lean` and `pointCount_eq_card_point` do exist on current Tau Ceti main, and I checked the finite/elliptic hypotheses there. That file is not present at this branch's pinned Tau Ceti commit `be0b30f3963c7f5158674dca96d7fe3fdd0dcd20` (the pinned file request returned 404). Mark this as a dependency update needed before implementing the count corollaries, or update the pin; the present Suggested.lean does not import it. This does not undermine the group-order argument, and Hasse is correctly excluded as a dependency.

The exact head's CI succeeds. No local Lean build was run. This is a COMMENT review, not an approval, and I have not changed the draft state.

